PR #68, opened 2026-07-28
refactor(tools): server-agnostic tool registration seam (mcp-use port, PR1/4)

tonychang04:
First of four PRs porting the HTTP server onto mcp-use. This one is a pure refactor — **no behaviour change, no dependency change, nothing about mcp-use lands here.** It exists so the later PRs don't have to touch 22 tool registrations.

## What

The tool modules were already decoupled: they take a `RegisterContext` and never reference `McpServer`. The one remaining coupling was in `registerInsforgeTools`, which called `server.tool(...)` directly.

- New `src/shared/tools/host.ts` defines `ToolHost` — the single call shape all 22 tools already use: `(name, description, zodRawShape, handler)`.
- `sdkToolHost(server)` adapts an MCP SDK server; `asToolHost()` still accepts a bare server, so existing callers and the integration-bridge test are untouched.
- `registerInsforgeTools` now takes `McpServer | ToolHost`. Version gating and local-only filtering are unchanged and happen before the host is called.

## Verification

- `tsc --noEmit` clean, `eslint` clean, `vitest run` 24/24 passing (6 new).
- End-to-end smoke against the **built** `dist/index.js` stdio bin (the published `insforge-mcp` entrypoint) with a stub backend: `initialize` + `tools/list` output is **byte-identical to `master`** — same 17 tools, same order.
- New tests cover the seam itself, that backend-version gating still filters (`fetch-sdk-docs` absent at backend 1.4.6), and that `bulk-upsert` is still skipped in remote mode.

## Blast radius

`src/shared/tools/index.ts`, `src/stdio/index.ts`, `src/http/session-manager.ts`. Both entrypoints import the changed function, so a mistake here would affect the published npm package and the hosted server — which is why the smoke test diffs the real bin against `master` rather than relying on unit tests. `mcp.json`, `server.json`, `package.json` bin entries and dependencies are untouched.

## Not in this PR

The mcp-use HTTP server (PR2, behind an `MCP_ENGINE` flag), the cutover (PR3), and removing the legacy engine (PR4). The stdio entrypoint stays on `@modelcontextprotocol/sdk` — mcp-use's `MCPServer` is HTTP-only and has no server-side stdio transport.

---
## Summary by cubic
Introduces a server-agnostic tool registration seam so the tool layer no longer depends on `@modelcontextprotocol/sdk`. Dropped `asToolHost`; `registerInsforgeTools` now requires a `ToolHost`. No behavior or dependency changes.

- **Refactors**
  - Added `ToolHost` and `sdkToolHost(server)` adapter.
  - `registerInsforgeTools(host, ...)` now takes a `ToolHost`; `stdio` and HTTP session manager pass `sdkToolHost(server)`.
  - Removed `asToolHost` to avoid mis-detecting `McpServer.registerTool` and misplacing arguments.
  - Added `host.test.ts` using a real `McpServer` over `InMemoryTransport`; tests cover the seam, backend-version gating, and local-only filtering.

<sup>Written for commit 98463fa643c9a30573b23d521a2fa1b1b35db237. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/68?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a standardized tool-hosting layer for registering and exposing MCP tools.
  * Tool registration now works consistently across session, streaming, and standard server connections.
  * Added support for validating tool availability based on backend version and connection mode.

* **Bug Fixes**
  * Improved reliability of tool discovery and invocation across supported integrations.

* **Tests**
  * Added comprehensive coverage for tool registration, invocation, filtering, and duplicate prevention.

### DIFF
--- a/src/http/session-manager.ts
+++ b/src/http/session-manager.ts
@@ -3,6 +3,7 @@ import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/
 import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
 import { getRedisClient } from './redis.js';
 import { registerInsforgeTools } from '../shared/tools/index.js';
+import { sdkToolHost } from '../shared/tools/host.js';
 import { PACKAGE_VERSION } from '../shared/version.js';
 
 /**
@@ -76,7 +77,7 @@ export class SessionManager {
       version: PACKAGE_VERSION,
     });
 
-    const toolsConfig = await registerInsforgeTools(server, {
+    const toolsConfig = await registerInsforgeTools(sdkToolHost(server), {
       apiKey: sessionData.apiKey,
       apiBaseUrl: sessionData.apiBaseUrl,
       mode: 'remote',
@@ -195,7 +196,7 @@ export class SessionManager {
       version: PACKAGE_VERSION,
     });
 
-    await registerInsforgeTools(server, {
+    await registerInsforgeTools(sdkToolHost(server), {
       apiKey: sessionData.apiKey,
       apiBaseUrl: sessionData.apiBaseUrl,
       mode: 'remote',
@@ -235,7 +236,7 @@ export class SessionManager {
       version: PACKAGE_VERSION,
     });
 
-    const toolsConfig = await registerInsforgeTools(server, {
+    const toolsConfig = await registerInsforgeTools(sdkToolHost(server), {
       apiKey: sessionData.apiKey,
       apiBaseUrl: sessionData.apiBaseUrl,
       mode: 'remote',
--- a/src/integration/real-project.test.ts
+++ b/src/integration/real-project.test.ts
@@ -18,6 +18,7 @@
 import 'dotenv/config';
 import { describe, it, expect, beforeAll } from 'vitest';
 import { registerInsforgeTools } from '../shared/tools/index.js';
+import { sdkToolHost } from '../shared/tools/host.js';
 import {
   createMockServer,
   getRequiredEnv,
@@ -58,7 +59,7 @@ describe.skipIf(!integrationEnabled)('Real Project Integration Tests (Phase 1)',
     const { server, tools: registeredTools } = createMockServer();
 
     // Register tools — NO try/catch: a failure here must crash the suite
-    await registerInsforgeTools(server, {
+    await registerInsforgeTools(sdkToolHost(server), {
       apiKey: API_KEY,
       apiBaseUrl: API_BASE_URL,
       mode: 'local',
--- a/src/shared/tools/host.test.ts
+++ b/src/shared/tools/host.test.ts
@@ -1,5 +1,9 @@
 import { describe, it, expect, vi } from 'vitest';
-import { asToolHost, sdkToolHost, type ToolHost } from './host.js';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
+import { z } from 'zod';
+import { sdkToolHost, type ToolHost } from './host.js';
 import { registerInsforgeTools } from './index.js';
 
 // registerInsforgeTools refuses to start unless it can read the backend version,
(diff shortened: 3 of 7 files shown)


jwfing (commented):
## Review: server-agnostic tool registration seam (PR1/4)

**Summary:** A clean, well-tested, behavior-preserving refactor that extracts a single `ToolHost` seam so the 22 tool registrations no longer depend on `McpServer` — I verified it is genuinely a no-op on behavior and found nothing blocking.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and there is no matching spec/plan doc for this change — assessing against the PR description alone (an explicitly-scoped pure refactor: introduce a `ToolHost` seam, no behavior/dependency change). The PR's own "no behaviour change" claim is the acceptance bar, so I focused verification there.

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering / functionality — the seam silently narrows the call contract with no guard** (`src/shared/tools/index.ts:187-194`). The old code forwarded `(server.tool as any)(toolName, ...args)`, resilient to whatever arity a tool passed. The new code hard-destructures `const [description, inputSchema, handler] = args`, and `RegisterToolFn` is still typed `(...args: any[]) => boolean` (`src/shared/tools/types.ts:6`), so there is no compile-time protection. I confirmed this is safe **today**: all 22 registrations pass exactly `(name, description, {schema}, withUsageTracking(...))` and none use the SDK's 5-arg annotations overload (grep for `readOnlyHint`/`annotations` etc. is empty). But the day someone adds a tool with an `annotations` object, `handler` will silently bind to the annotations and the real callback is dropped — and the registration tests below wouldn't catch it. Consider asserting `args.length === 3` in `registerTool`, or tightening `RegisterToolFn`/`ToolHost.registerTool` to the exact shape so the SDK's variadic overloads can't be passed unnoticed.

- **Software engineering — full-surface tests don't exercise handlers** (`src/shared/tools/host.test.ts:113-146`). The `registerInsforgeTools through a ToolHost` cases assert name/description/schema presence and that `toolCount === calls.length`, and the SDK round-trip case checks `listTools`, but only the standalone `echo-tool` test (`:44-72`) actually `callTool`s through the seam. Since handler forwarding is the whole point of the seam, a single `callTool` against one real registered insforge tool (even a stubbed-backend one) would close the gap that the manual byte-diff smoke test currently covers out-of-CI. Non-blocking given the positional forwarding is trivial.

#### Information

- **Stale PR description** — the "## What" section (and the cubic summary) says "`asToolHost()` still accepts a bare server, so existing callers and the integration-bridge test are untouched." The code exports only `sdkToolHost` (`src/shared/tools/host.ts:32`); there is no `asToolHost`, and `integration-bridge.test.ts:63` **was** changed to wrap with `sdkToolHost(mockServer as any)`. The test comment at `host.test.ts:39-42` also refers to "the earlier duck-typed `asToolHost()`". Code is correct — just update the description so reviewers of PR2-4 aren't looking for a symbol that doesn't exist.

- **Security:** no security-relevant changes — no new user input paths, no SQL/shell/HTTP construction, no secrets logged or returned, no auth/authorization changes, no dependency changes (`package.json`/lockfile deps untouched).

- **Performance:** no performance-relevant changes — `sdkToolHost` allocates one closure per server construction (once per stdio boot / HTTP session), off any hot path. Version gating and local-only filtering still run before the host is invoked, unchanged.

### Verification I ran
- All 22 `registerTool` call sites across `database/deployment/docs/functions/storage.ts` use the identical 4-arg shape; no annotations/5-arg overload anywhere → the destructuring in `index.ts:193` is behavior-preserving.
- `sdkToolHost.registerTool` → `(server.tool as any)(name, description, inputSchema, handler)` is positionally identical to the old `(server.tool as any)(toolName, ...args)`.
- The `createMockServer` mock (`src/integration/helpers.ts:46-73`) handles the `args.length >= 3` branch, so the two rewired tests (`integration-bridge`, `real-project`) still resolve description/schema/cb correctly through the wrapper.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). No Critical findings; the two Suggestions are low-blast-radius hardening and the Information items are doc/test-coverage notes.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Summary

Clean, correctly-scoped pure refactor introducing a `ToolHost` seam so the tool layer no longer depends on `@modelcontextprotocol/sdk`; the change is behaviorally identical to the prior `(server.tool as any)(toolName, ...args)` path and is well covered by new tests.

## Requirements context

No `docs/superpowers/` (or any `docs/` spec/plan) directory exists in this repo — assessing against the PR description alone. The stated intent (decouple the 22 tool registrations from `McpServer` ahead of the mcp-use port, with **no behaviour/dependency change**) is the source of truth used below.

## Findings

### Critical
(none)

- **Functionality:** The stated goal is met. `src/shared/tools/index.ts` drops the `McpServer` import; the core tool layer (`index.ts` + domain modules) now depends only on the `ToolHost` interface, with `sdkToolHost` (`src/shared/tools/host.ts:32`) as the only SDK-aware adapter. All five call sites were updated to wrap with `sdkToolHost(...)` (`src/stdio/index.ts:28`, `src/http/session-manager.ts:80,199,239`, plus both integration tests) — no caller passes a bare server, so nothing is left half-migrated.
- **Behavior equivalence:** The new `host.registerTool(toolName, description, inputSchema, handler)` (`src/shared/tools/index.ts:193-194`) is equivalent to the old spread call because every one of the 22 `registerTool(...)` sites passes exactly `(name, description, zodRawShape, handler)` (spot-checked `database.ts:87`, `functions.ts:22`), and `sdkToolHost` forwards positionally into the same `server.tool(name, description, paramsSchema, cb)` overload. Version gating and local-only filtering still run before the host is invoked.
- **Security:** No security-relevant changes — no new user input paths, SQL/shell/HTTP surface, auth checks, dependencies, or logging of secrets. The diff only moves an existing call one indirection deeper.
- **Performance:** No concern — `sdkToolHost` allocates one small closure per `registerInsforgeTools` call (≈3×/session), negligible; no new queries, loops, or blocking work.

### Suggestion

- **SE — untyped seam leaks past the new interface:** The internal helper stays `registerTool = (toolName: string, ...args: any[])` (`src/shared/tools/index.ts:187`) / `RegisterToolFn` (`src/shared/tools/types.ts:6`), and `index.ts:193` destructures `[description, inputSchema, handler]` out of that `any[]`. A caller that ever omitted or added an argument would now silently pass `undefined` / drop the extra rather than fail to compile. Zero blast radius today (all callers conform), but since you've just defined an explicit `ToolHost.registerTool(name, description, inputSchema, handler)` contract, tightening `RegisterToolFn`/the internal helper to the same 4-arg signature would make that guarantee compiler-enforced instead of convention-enforced.
- **SE — version-gating test comment vs assertion:** In `src/shared/tools/host.test.ts:150-155`, the setup comment says `1.4.6` is "one below create-deployment's 1.4.7 minimum" but the assertions only check `fetch-sdk-docs` (requires 1.5.1) absent and `fetch-docs` present. Since `create-deployment` is also gated out at 1.4.6, asserting `names).not.toContain('create-deployment')` would make the test match its own comment and cover the boundary case it set up.

### Information

- **Stale PR description:** The body states "`asToolHost()` still accepts a bare server, so existing callers and the integration-bridge test are untouched." In the actual diff there is no `asToolHost` (the cubic summary correctly notes it was dropped in favor of `sdkToolHost`), and `src/shared/tools/integration-bridge.test.ts` **was** modified to use `sdkToolHost(...)`. Harmless to the code, but worth correcting so reviewers aren't misled about blast radius.
- **Wrapping a deprecated SDK method:** `sdkToolHost.registerTool` forwards to the SDK's `server.tool()` (`host.ts:36`), which `@modelcontextprotocol/sdk@^1.27.1` marks deprecated in favor of `server.registerTool()`. Correct to keep as-is here (no-behaviour-change PR), but a natural follow-up now that the seam is named `registerTool`.
- Minor: trailing extra blank line at the end of `src/shared/tools/host.ts` (eslint reported clean, so not blocking).

## Verdict

**approved** (informational — human still approves via the approve flow). Zero Critical findings: the refactor is behavior-preserving, fully migrated across all call sites, and backed by meaningful tests that use a real `McpServer` over `InMemoryTransport` rather than a stub. The Suggestion/Information items are non-blocking.

PR #69, opened 2026-07-28
fix(oauth): make OAuth discoverable — WWW-Authenticate + per-endpoint resource metadata

tonychang04:
Three files. One thing: a spec-correct MCP client cannot currently discover how to log in.

Independent of the mcp-use port. The `verify:deploy` tooling that used to live here has moved to #73 — it drew 8 of this PR's 10 review comments and its pushes dismissed this PR's approval, which is why a fix with users waiting sat for a day.

## Two defects, and the second is why the first isn't enough

**1. No `WWW-Authenticate` on 401.** Unauthenticated requests to `/mcp` and `/sse` return a bare 401. The MCP authorization spec requires the challenge to carry `resource_metadata` (RFC 9728 §5.1) — that header is how a client discovers where the metadata lives and starts OAuth. Confirmed against production:

```
$ curl -si -X POST https://mcp.insforge.dev/mcp -d '{...initialize...}' | grep -i www-authenticate
(nothing)
```

The string appears nowhere in the source; all seven 401 sites omitted it. Clients that connect today are guessing the well-known path.

**2. The document we'd point them at was unusable.** RFC 9728 §3.1 derives the metadata URL by inserting the well-known path *between the host and the resource's path*, and §3.3 requires:

> The `resource` value returned MUST be identical to the protected resource's resource identifier value into which the well-known URI path suffix was inserted to create the URL used to retrieve the metadata. If these values are not identical, the data contained in the response MUST NOT be used.

`master` served **one** document, at the bare well-known path, declaring `resource: <base>/mcp`. A client retrieving it from that URL derived `<base>` — so a conforming client must discard it. Fixing (1) alone would have pointed strict clients at a document they're required to throw away.

## The fix

- `sendUnauthorized()` sets the challenge before the body, on the five 401s at the MCP resource endpoints. The two on `/api/projects` keep their plain 401 — our own API, not the MCP resource.
- Each protected endpoint serves its own document at its RFC-derived URL (`/mcp`, `/sse`); the challenge names the one for the endpoint actually hit.
- The bare path is kept, describing the origin, for clients that probe it instead of following the header.
- **Trailing slashes normalised** (review finding): a `MCP_SERVER_URL` ending in `/` produced `https://host//.well-known/…`, a URL no route matches. Every field — the derived URL, `resource`, and `authorization_servers` — normalises through one helper so they cannot diverge.

Helpers live in `src/http/auth-challenge.ts` because `server.ts` calls `startServer()` at import time and can't be imported by a test.

## Verification

32/32 vitest, `tsc` and `eslint` clean. On the wire against the built server:

```
POST /mcp  → 401, WWW-Authenticate: Bearer resource_metadata=".../oauth-protected-resource/mcp"
GET  .../oauth-protected-resource/mcp → {"resource":"https://mcp.insforge.dev/mcp", ...}
GET  /sse  → 401, ... resource_metadata=".../oauth-protected-resource/sse"
GET  .../oauth-protected-resource     → {"resource":"https://mcp.insforge.dev", ...}
```

Every document's `resource` matches its own URL. The trailing-slash test now actually feeds it `/` and `//` — the previous one asserted the guarantee while passing a base that had none, so it could never have failed.

## ⚠️ Lands its benefit on deploy, not on merge

`mcp.insforge.dev` is not running `master` — it reports `version: 1.0.0` and serves the old shape, and there's no deploy workflow in the repo. Merging this changes nothing for real clients until the deployment is current. A green merge should not be read as "clients can discover OAuth now."

## Blast radius

The public OAuth path. Status codes and bodies are unchanged, so clients ignoring the header behave exactly as before. The bare-path document's `resource` changes from `<base>/mcp` to `<base>` — that value was unusable at that URL under §3.3, so no conforming client could have relied on it.


jwfing (commented):
## Review — fix(oauth): send `WWW-Authenticate` on 401

**Summary:** A tight, additive fix that attaches an RFC 9728 §5.1 `WWW-Authenticate: Bearer resource_metadata=…` challenge to the five MCP-resource 401s so spec-compliant clients can discover the authorization server; no critical issues found.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`, `docs/plans/`) directory exists in this repo — assessing against the PR description alone. The stated intent (RFC 9728 / MCP authorization discovery on the protected-resource 401s) maps cleanly onto the diff.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Robustness — trailing-slash normalization** · `src/http/auth-challenge.ts:12-14`
  `buildResourceChallenge` concatenates `${publicUrl}${OAUTH_ENDPOINTS.protectedResource}` with no normalization. Today `SERVER_CONFIG.publicUrl` defaults to a slash-free value (`http://localhost:3000`) and prod is configured without a trailing slash, so this is correct now. But `publicUrl` comes straight from `MCP_SERVER_URL` (`src/http/config.ts:31`) with no trimming, so a deploy that sets `MCP_SERVER_URL=https://mcp.insforge.dev/` would emit `…//.well-known/oauth-protected-resource`, which strict clients reject. Your own test (`auth-challenge.test.ts:22-24`) documents this assumption but doesn't guard the code path. Consider `publicUrl.replace(/\/$/, '')` at the source or in the helper. Low blast radius — non-blocking.

#### Information

- **Software engineering — call-site wiring is not integration-tested** · `src/http/auth-challenge.test.ts`
  The 4 new unit tests exercise `buildResourceChallenge`/`sendUnauthorized` in isolation (header value, status/body, header-before-body ordering) — good coverage of the helper. They do not assert that each of the 5 server.ts call sites actually invokes it. That's a reasonable tradeoff given `server.ts` runs `startServer()` at import time and can't be imported by a test (as the PR notes), and the on-the-wire verification in the description covers the gap. Flagging only so it's a conscious choice.
- **Functionality — `resource` mismatch on the legacy `/sse` path** · `src/http/server.ts:174-178` vs `940-953`
  The challenge is now served from both `POST /mcp` and `GET /sse`, but the protected-resource metadata document it points at declares `resource: ${baseUrl}/mcp`. A strict client that discovers via a `/sse` 401 will read a resource identifier of `/mcp`, not `/sse`. This is pre-existing (the metadata doc predates this PR) and `/sse` is the deprecated 2024-11-05 transport, so it's not introduced here — noting for awareness, not action.
- **Functionality — invalid/expired tokens also receive the challenge** · `src/http/server.ts:768-773`
  Because `resolveProjectFromToken` returns null for a bad token, an invalid/expired OAuth token falls into the `project_binding_required` branch, which now carries the discovery header — so discovery is covered for that case too. The semantic label (`project_binding_required` rather than RFC 6750 `invalid_token`) is pre-existing and out of scope here.

---

### Dimension coverage

- **Security:** Clean. The header value is built from `SERVER_CONFIG.publicUrl` (config/env), never from the request `Host` — consistent with the existing anti-spoofing comment at `server.ts:169`, so no reflected-host injection into the challenge. No secrets logged, no auth check weakened (status and body are byte-for-byte unchanged), no new dependencies.
- **Performance:** Negligible — a single `setHeader` on the 401 error path only.
- **Functionality:** Solves the stated bug; the two `/api/projects` 401s (`server.ts:663`, `687`) are correctly left plain since they aren't the MCP protected resource.
- **Software engineering:** Helper extracted for testability with a clear rationale; ESM `.js` import style and error-body shape match surrounding code.

---

### Verdict

**approved** (informational — human approval via the separate approve flow). No blocking issues; the two items above are non-blocking. Nice baseline-before-you-change discipline, and thanks for flagging the deployed-server version drift separately rather than folding it in.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: fix(oauth): send `WWW-Authenticate` on 401 so MCP clients can discover the AS

**Summary:** A tight, correct, test-backed fix that attaches an RFC 9728 §5.1 `WWW-Authenticate: Bearer resource_metadata="..."` challenge to exactly the five MCP-resource 401s, leaving the two `/api/projects` 401s as plain 401s — no Critical findings.

### Requirements context
No `/docs/superpowers/` (or any `docs/` spec/plan) directory exists in this repo, so there's no spec to assess against — reviewed against the PR description, RFC 9728 §5.1, and the MCP authorization spec as the source of intent.

### Verification of scope claims
Confirmed against the checked-out head (`fd3b3a2`):
- `grep` for `status(401)` / `WWW-Authenticate` across `src/` finds **exactly seven** 401 sites. Five are now routed through `sendUnauthorized()` (three in `POST /mcp` at `src/http/server.ts:763,771,794`, two in `GET /sse` at `src/http/server.ts:943,950`); the two on `/api/projects` (`src/http/server.ts:663,687`) correctly keep bare 401s. This matches the PR body precisely.
- The challenge value `Bearer resource_metadata="${publicUrl}/.well-known/oauth-protected-resource"` matches the RFC 9728 §5.1 example format exactly (`src/http/auth-challenge.ts:12-14`, `OAUTH_ENDPOINTS.protectedResource` = `/.well-known/oauth-protected-resource`).
- Leaving `/api/projects` as plain 401 is the right call — those are the server's own API, not the MCP protected resource, so RFC 9728 discovery doesn't apply.

---

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / functionality — the trailing-slash guard test is vacuous** (`src/http/auth-challenge.test.ts:22-26`). The test `does not leave a trailing slash from the public URL` asserts `.not.toContain('//.well-known')` but feeds an input (`'https://mcp.insforge.dev'`) that already has no trailing slash, so it passes trivially. `buildResourceChallenge` (`src/http/auth-challenge.ts:12-14`) does **no** normalization — if `MCP_SERVER_URL` were ever configured with a trailing slash (`https://mcp.insforge.dev/`), the emitted value becomes `...dev//.well-known/...`, which strict clients reject. The comment and CodeRabbit summary ("Prevented malformed metadata URLs containing duplicate slashes") both imply a protection the code doesn't have. Low blast radius today (the documented config convention is no trailing slash), so non-blocking — but either strip a trailing slash in the builder (`publicUrl.replace(/\/$/, '')`) or feed the test an actual trailing-slash input so it exercises what its name claims.

#### Information
- **Functionality — token-present branches emit only a bare Bearer challenge.** For `project_binding_required` and `invalid_credentials` (`src/http/server.ts:771,794,950`) a credential *was* supplied. RFC 6750 §3 permits an `error`/`error_description` parameter on the challenge. Discovery (the PR's goal) is fully served without it, and for `project_binding_required` the token is actually valid so `error="invalid_token"` would be wrong — so this is genuinely optional, noted only for completeness.
- **Design — good testability call.** Extracting the helper into `src/http/auth-challenge.ts` rather than `server.ts` (which runs `startServer()` at import) is the right move and lets the four new unit tests run in isolation. Convention-compliant: colocated `*.test.ts`, vitest, ESM `.js` import specifiers all match the surrounding code.
- **Out of scope, correctly deferred.** The PR notes the deployed server reports `version: 1.0.0` and serves `resource: "https://mcp.insforge.dev"` (vs `${baseUrl}/mcp` on `master`). Agreed this is a separate deployment/metadata issue and correctly not folded in here.

---

### Dimension coverage
- **Software engineering:** ✅ 4 new unit tests cover header content, status, body, and header-before-body ordering; follows repo conventions. One test is vacuous (see Suggestion).
- **Functionality:** ✅ Solves the stated bug; all five MCP-resource 401 branches covered; additive (status/body unchanged) so honoring clients gain discovery and ignoring clients are unaffected.
- **Security:** No security-relevant weakening. The header exposes only the already-public metadata URL — no secrets/PII/tokens leaked; no auth check removed or loosened; no new dependencies.
- **Performance:** No concern — a single `setHeader` on an early-return error path; no new I/O, loops, or allocations.

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings; the one Suggestion and the Information notes are all non-blocking.

jwfing (commented):
## Review: fix(oauth): make OAuth discoverable — WWW-Authenticate + per-endpoint resource metadata

**Summary:** A focused, well-scoped, RFC-grounded fix that adds `WWW-Authenticate: Bearer resource_metadata=…` to the MCP-resource 401s and serves a self-consistent protected-resource document per endpoint — no correctness, security, or performance blockers.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the git history on the branch, and the cited specs (RFC 9728 §3.1/§3.3/§5.1, RFC 8414). The PR's own claims were verified against the code.

I could not execute the `vitest` / `tsc` / `eslint` gates (the shallow review clone has no `node_modules`, and I stayed read-only), so the "30/30, tsc/eslint clean" claim rests on the PR statement plus a static read — which is consistent with what I see.

---

### Critical
(none)

### Suggestion

**Software engineering — route wiring is not covered by a PR-time test** (`src/http/server.ts:176-193`, `:777-808`, `:957-967`)
The new pure helpers in `auth-challenge.ts` are thoroughly unit-tested (12 tests), which is the right place for the logic. But the wiring in `server.ts` — that the metadata routes register at `/.well-known/oauth-protected-resource/mcp` and `/sse`, and that each 401 site passes the *matching* `resourcePath` — is exercised only by `scripts/verify-deploy.mjs`, which runs on a daily schedule against a deployed server, not as a PR gate. So a future edit that, say, passed the wrong `resourcePath` at one 401 site would pass CI. The `startServer()`-at-import constraint (correctly called out in the PR) makes a full supertest harness awkward, but extracting the Express route registration into an importable `registerRoutes(app)` function would let a test assert the wiring without booting the listener. Non-blocking.

### Information

- **Security — unquoted workflow input** (`.github/workflows/verify-deploy.yml:46`): `${{ github.event.inputs.expected_version }}` is interpolated unquoted into a `run:` shell block — the classic GitHub Actions script-injection shape. Blast radius is minimal (only a collaborator with `workflow_dispatch` can set it, and they could edit the workflow anyway), but quoting it or passing it via an `env:` var is the hygienic pattern.
- **Functionality — `scopes_supported` divergence** (`src/http/auth-challenge.ts:34`): the protected-resource document declares `['mcp:read','mcp:write']`, while `OAUTH_CONFIG.supportedScopes` (`config.ts:64`) and the AS metadata carry `['mcp:read','mcp:write','project:select']`. This is pre-existing (the old inline document had the same two scopes), not introduced here — but since the PR is now the single source for this document, it's a good moment to reconcile the lists or leave a note on why the resource intentionally advertises a subset.
- **Functionality — challenge on `project_binding_required`** (`src/http/server.ts:785-790`, `:963-967`): attaching a discovery challenge to a 401 where the token is *valid but unbound* is defensible (the client still needs to finish the OAuth/project-selection flow) and the status/body are unchanged, so no client regresses. Flagging only so it's a conscious choice rather than an accident of the blanket conversion.

---

### Dimension coverage
- **Software engineering:** Strong. `.js` ESM imports, `vitest` + `vi` mocks, and RFC-cited comments all match repo conventions; helpers extracted to a testable module. Only gap is the untested route wiring above.
- **Functionality:** Solves what it claims. All unauthenticated-entry 401 sites are covered — 5 MCP-resource sites converted, 2 `/api/projects` sites deliberately kept plain (they are the platform's own API, not the MCP resource). `GET`/`DELETE /mcp` and `POST /messages` act on already-established sessions and correctly emit no 401, so the "seven sites" accounting is complete. RFC 9728 URL insertion and the §3.3 identifier-equality invariant are implemented correctly and asserted in tests.
- **Security:** No weakening. Metadata is built from `SERVER_CONFIG.publicUrl`, never the `Host` header (host-spoofing-safe); `resourcePath` values are fixed constants, not user input; no secrets or PII are logged or added to responses; only a header is added to existing 401s, so clients ignoring it are unaffected. The one note is the unquoted workflow input above.
- **Performance:** No concerns. Metadata handlers are O(1) object construction; `sendUnauthorized` adds a single `setHeader`. No new DB queries, loops, allocations, or blocking I/O.

### Verdict
**approved** (informational — posted as a comment; explicit GitHub approval remains a separate human action). Zero Critical findings. The Suggestion and Information items are worth a look but none block merge. Note the PR's own caveat stands: the fix only takes effect once `mcp.insforge.dev` is deployed current.

jwfing (commented):
## Review: fix(oauth) — WWW-Authenticate + per-endpoint resource metadata

**Summary:** A correct, well-scoped, and thoroughly-tested fix that makes OAuth discoverable by adding RFC 9728 §5.1 `WWW-Authenticate` challenges on the MCP-resource 401s and serving per-endpoint, self-consistent protected-resource metadata documents.

### Requirements context
No spec/plan directory exists in this repo (`docs/superpowers/`, `docs/`, and `.claude/skills/` are all absent), so I assessed against the PR description and the RFCs it cites. The interpretation of RFC 9728 §3.1 (insert the well-known segment between host and resource path), §3.3 (`resource` must equal the identifier the URL was derived from), and §5.1 (`resource_metadata` parameter on the Bearer challenge) is accurate, and the implementation matches it. Verified in the workspace: `tsc --noEmit` clean, `eslint` clean for the changed files (the 6 warnings are all pre-existing in untouched files), and the 12 new `auth-challenge.test.ts` cases pass (12/12).

### Findings

#### Critical
(none)

#### Suggestion
- **Security — `.github/workflows/verify-deploy.yml:44-46` (GitHub Actions expression injection).** The `workflow_dispatch` inputs are interpolated directly into the `run:` shell: `"${{ github.event.inputs.url || '...' }}"` and — unquoted — `${{ github.event.inputs.expected_version }}`. This is the classic Actions script-injection anti-pattern; the unquoted `expected_version` in particular lets a dispatcher inject extra shell tokens. Blast radius is limited (only users with write access can dispatch, and the script is read-only), which is why this is non-blocking rather than Critical, but the recommended hardening is to pass inputs via `env:` and reference `"$URL"` / `"$EXPECTED_VERSION"` in the script body rather than templating them into the command line.
- **Software engineering — no CI-run test covers the `server.ts` route wiring.** The helpers in `auth-challenge.ts` are exercised well, but the actual Express registration (`server.ts:176-193`) and the `sendUnauthorized(..., <path>)` call sites (`server.ts:777,785,808,957,964`) have no automated coverage that runs in CI — a regression such as passing the wrong `resourcePath` to the SSE handler would only be caught by `verify-deploy.mjs`, which runs on a schedule against a live deploy, not on PRs. The PR's rationale (server.ts calls `startServer()` at import time and can't be imported by a test) is fair and refactoring that is out of scope here, so this is just a note that the wiring itself is currently untested.

#### Information
- **Functionality — trailing-slash on `MCP_SERVER_URL`.** `SERVER_CONFIG.publicUrl` (`config.ts:31`) is used verbatim in `protectedResourceMetadataUrl`/`protectedResourceMetadata`; a configured value with a trailing slash would yield `https://host//.well-known/...` and a doubled origin in `resource`. This is pre-existing behavior (the old code had the same shape) and `verify-deploy.mjs` already strips the trailing slash from its `base`, so it's informational only — worth a one-line normalization at config load if you want to close it.
- **Scope / conventions.** Nicely disciplined: the two `/api/projects` 401s (`server.ts:677,701`) are correctly left as plain 401s since they are the platform's own API rather than the MCP resource; no new runtime dependencies are added (only a `verify:deploy` script and a scoped eslint globals block for `scripts/**/*.mjs`); import style (`.js` extensions), config-constant usage, and JSDoc all match the surrounding code.

**Performance:** No performance-relevant concerns — the change only sets a response header and serves static JSON; no new queries, loops, allocations, or blocking I/O in any hot path.

**Security (summary):** The metadata uses the canonical `SERVER_CONFIG.publicUrl` rather than the request host header (explicitly avoiding host-header spoofing), `resourcePath` is sourced from server constants and never from user input, the 401 status/body are unchanged (so no new information disclosure), and `verify-deploy.mjs` is read-only and registers nothing. The only security note is the Actions-injection Suggestion above.

### Verdict
**approved** (informational; no Critical findings). The two Suggestions and the Information notes are non-blocking. As the PR body itself flags, the benefit only lands once `mcp.insforge.dev` is deployed current — the merge alone doesn't change production behavior, and existing clients that ignore the header are unaffected.

tonychang04:
## QA — OAuth discovery: what passes, and the recovery gap reproduced

Measured against this branch built and running locally (`node dist/http-server.js`, live Redis 7.0.15, `INSFORGE_CLIENT_ID`/`SECRET` set). Not measured against prod.

### What holds up

- `npm test` 32/32, `npm run lint` 0 errors (6 pre-existing `no-explicit-any` warnings).
- `npm run verify:deploy http://127.0.0.1:3113 1.2.10` → **22/22 PASS**.
- **The check has teeth.** Run against a build of this repo *without* this PR (a `fix/session-leak-and-sse-keepalive` build, which is off master), it fails on exactly the defects the PR fixes:

```
FAIL  401 carries WWW-Authenticate  — null
FAIL  challenge names a resource_metadata document  — null
FAIL  derived document for /mcp is 200  — status 404
FAIL  derived document for /sse is 200  — status 404
FAIL  origin document is self-consistent  — resource=http://127.0.0.1:3111/mcp
9/18 checks passed
```

That is a gate that goes red on the bug it was written for, which is more than most post-deploy checks manage.

### The recovery gap — reproduced, and it is on a 30-day clock

The 401 path is fixed. The stale-registration path is not, and it is not an edge case:

**`src/http/server.ts:233` stores each registration with a 30-day TTL** (`setex mcp:oauth:client:<id>, 30*24*60*60`). Access-token bindings get the same 30 days, and `/oauth/token` answers `unsupported_grant_type` for `refresh_token` — so there is no refresh path that could renew anything. Every OAuth client that has been installed for 30 days arrives back at `/oauth/authorize` with a `client_id` the server has already forgotten:

```json
{
  "registrationTtlSeconds": 2592000,
  "authorizeWithLiveClient":  { "status": 302, "location": "https://api.insforge.dev/api/oauth/v1/authorize?..." },
  "authorizeWithStaleClient": { "status": 400, "wwwAuthenticate": null,
    "body": "{\"error\":\"invalid_client\",\"error_description\":\"Unknown client_id. Register client first via /oauth/register.\"}" }
}
```

**And the client cannot recover on its own.** I drove the real `@modelcontextprotocol/sdk@1.28.0` `auth()` flow against the server with a stored-but-unknown `client_id`:

```json
{
  "storedClientIdUsed": "mcp_staleclientthatnolongerexists",
  "reRegistered": false,
  "events": ["redirectToAuthorization(mcp_staleclientthatnolongerexists)", "auth() -> REDIRECT"],
  "whatTheUsersBrowserGets": { "status": 400, "contentType": "application/json",
    "body": "{\"error\":\"invalid_client\", ...}" }
}
```

Control, same script with no stored client information: the SDK registers, `reRegistered: true`, and the browser gets the `302` to the platform. So the harness has teeth and the stored registration is the trigger.

The SDK *does* have recovery for this: `auth()` catches `InvalidClientError` → `invalidateCredentials('all')` → re-register (`client/auth.js:151-157`). It never fires here, for two compounding reasons:

1. `/oauth/authorize` is a **browser navigation** — the 400 lands in the user's browser as raw JSON, and the SDK never sees a response to classify.
2. `/oauth/token` never validates `client_id` against the registry at all (`server.ts:575`), so the one endpoint whose errors the SDK *does* see can never produce `invalid_client` either.

Net: the user sees a JSON error page, and the only fix is to remove and re-add the connector.

Cheapest changes that would close it, roughly in order of cost:

- **Stop expiring registrations** (or make the TTL sliding on use). A public client registration is not a session; the 30-day expiry is what turns "host moved" into "every install, every month".
- Serve **HTML rather than JSON** on the `/oauth/authorize` error path, saying "this connector's registration has expired, remove and re-add it". Does not fix it, but it stops the dead end being unreadable.
- Validate `client_id` at `/oauth/token` and return `invalid_client` — that is the error shape the SDK already knows how to recover from, and it covers the sub-case where the client reaches the token exchange.

Repro scripts are in the thread; happy to open them as a PR against `scripts/` if you want them to outlive this comment.

*Not tested:* anything against prod `mcp.insforge.dev` (read-only measurement only), the platform's own `/authorize`, and a real browser-driven end-to-end login (no credentials for the platform AS here).

— Quinn

tonychang04:
## QA — re-run on `ae0e438` (my earlier pass was on `b3d17e7`)

**26/26** with the checker from #73, and `274ecec` does what it claims: with `MCP_SERVER_URL` ending in `/`, the protected-resource documents come out normalised —

```json
{"resource":"http://127.0.0.1:3124","authorization_servers":["http://127.0.0.1:3124"]}
{"resource":"http://127.0.0.1:3124/mcp","authorization_servers":["http://127.0.0.1:3124"]}
```

(and it was genuinely broken before: `authorization_servers: [publicUrl]` at `b3d17e7` vs `[origin(publicUrl)]` now).

### The same bug is still live in the authorization-server document, and it breaks sign-in

`server.ts:154-158` builds that document from the raw `baseUrl`, so a trailing slash survives:

```
issuer                  http://host/
authorization_endpoint  http://host//oauth/authorize
token_endpoint          http://host//oauth/token
registration_endpoint   http://host//oauth/register
```

Express doesn't match the double-slash path, so a client cannot register at all:

```
FAILED at [discovery + registration]: HTTP 404 … Cannot POST //oauth/register
```

Everything else stays green — health, both resource documents, the challenge — and only the handshake fails. `verify:deploy` catches half of it (`25/26`, issuer mismatch) but not the endpoint paths.

Reachability is the part that makes it worth fixing here rather than later: `MCP_SERVER_URL` is hand-entered configuration, and the Manufact deployment is being configured from a UI this week. A pasted URL with a trailing slash produces a deployment that looks healthy and can log nobody in.

Fix is the one already in this PR, applied one block over: run `baseUrl` through the same `origin()` helper before building that document.

*Not tested:* a real browser sign-in (the handshake script's one manual step), and prod.

— Quinn

jwfing (commented):
## Review: fix(oauth) — WWW-Authenticate + per-endpoint resource metadata

**Summary:** A correct, well-scoped, and well-tested fix that makes the MCP server's OAuth flow spec-discoverable (RFC 9728 §3.1/§3.3/§5.1); no blocking issues.

**Requirements context:** No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory, so this was assessed against the PR description, the RFC 9728 / MCP-authorization requirements it cites, and the surrounding code. The RFC citations were cross-checked against the implementation and are accurate.

I confirmed the claims in the PR body against the workspace:
- **7 total 401 sites accounted for.** Five MCP-resource 401s are now routed through `sendUnauthorized` (`src/http/server.ts:777,785,808` on `/mcp`; `957,964` on `/sse`); the two on `/api/projects` (`src/http/server.ts:677,701`) intentionally keep a plain 401. `POST /messages` (`server.ts:1066`) issues no 401, so nothing was missed.
- **RFC-derived metadata URLs are correct.** `/.well-known/oauth-protected-resource/mcp` → `resource: <base>/mcp`, `/sse` likewise, and the bare path → `resource: <base>` — each document's `resource` equals the identifier its URL was derived from (§3.3). Route registration order is fine (all three are exact paths).

---

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality / consistency — trailing-slash normalization is only half the discovery chain.** `src/http/auth-challenge.ts:10-11` normalizes the trailing slash for the resource-metadata document and `authorization_servers`, but the Authorization-Server metadata document those entries point at still uses the raw base: `src/http/server.ts:152-159` emits `issuer`/`authorization_endpoint`/… from unnormalized `SERVER_CONFIG.publicUrl` (same for `callbackUrl` at `src/http/config.ts:61` and the `oauth_authorize_url`/`projects_url` bodies at `server.ts:780,788-789`). With a misconfigured `MCP_SERVER_URL` ending in `/`, a strict client that chains PR-metadata → `authorization_servers: ["https://host"]` → AS-metadata would read `issuer: "https://host/"` and see a mismatch, plus `https://host//oauth/authorize`. Blast radius is limited to that misconfiguration (and this is exactly the class of bug the PR set out to kill), so it's non-blocking — but normalizing once at config load (so `SERVER_CONFIG.publicUrl` is canonical for *every* consumer) would close the gap more completely than normalizing inside one helper.

#### Information

- **Functionality — challenge attached to post-authentication 401s.** `project_binding_required` (`server.ts:785,964`) and `invalid_credentials` (`server.ts:808`) are states where the client already holds a valid OAuth token; attaching the `resource_metadata` discovery challenge there is harmless (bodies still carry the `oauth_authorize_url`/`projects_url` guidance), but a spec-strict client could re-run OAuth discovery rather than following the bind guidance. Author's call — flagging, not requesting.
- **Software engineering — helper design & test quality are good.** Extracting the helpers into `auth-challenge.ts` because `server.ts` self-starts on import is the right call, and the unit suite (`auth-challenge.test.ts`) covers URL derivation, per-endpoint `resource` identity, header-before-body ordering, and — notably — now feeds actual `/` and `//` inputs so the trailing-slash guarantee is genuinely exercised rather than asserted against an input that never had a slash.
- **Testing — route wiring is unit-covered only.** The three `app.get` handlers in `server.ts` have no integration test (server self-starts on import, as documented), so their behavior is verified indirectly through the helper tests. Acceptable given the constraint and consistent with the repo.
- **Security — no new exposure.** `resourcePath` values are fixed constants and `publicUrl` comes from env, so there's no user input reaching the header; the header is set before the body; no secrets/PII added to responses or logs. No auth checks were weakened — status codes and bodies are unchanged for clients that ignore the header.
- **Performance — nothing hot-path relevant.** Pure string construction on the 401 path; no new I/O, queries, or allocations of note.

---

Zero Critical findings. Posted as a COMMENT — the explicit GitHub approval remains a separate human action. Worth echoing the PR's own caveat for the record: this lands its benefit on **deploy**, not on merge (`mcp.insforge.dev` still reports `version: 1.0.0`), so a green merge should not be read as "clients can discover OAuth now."

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review: fix(oauth): make OAuth discoverable — WWW-Authenticate + per-endpoint resource metadata

**Summary:** A tight, well-tested, correctly-scoped fix that adds the RFC 9728 `WWW-Authenticate` discovery challenge to the MCP-resource 401s and splits protected-resource metadata into per-endpoint documents so each is self-consistent under §3.3.

**Requirements context:** `insforge-mcp` has no `/docs/superpowers/` (or any `docs/`) spec directory — assessing against the PR description, the cited RFCs (9728 §3.1/§3.3/§5.1, 6750), and surrounding code. The RFC constructions were verified directly against the source.

---

### Findings

#### Critical
(none) — no correctness bugs, no security regressions, no broken existing behavior.

#### Suggestion

- **Functionality — token-present 401s could carry an `error` parameter.** `src/http/server.ts:785` (`project_binding_required`) and `:808` (`invalid_credentials`) fire when a bearer token *was* supplied but is insufficient. RFC 6750 §3 recommends the `WWW-Authenticate` challenge include `error="invalid_token"` + `error_description` in that situation, which helps clients distinguish "no token" from "token rejected." The current minimal `Bearer resource_metadata="…"` is spec-valid for discovery, so this is a refinement, not a defect — low blast radius.

#### Information

- **Partial trailing-slash normalization.** The new `origin()` helper (`src/http/auth-challenge.ts:11`) correctly normalizes the derived URL, `resource`, and `authorization_servers`, matching the PR's claim that those three can't diverge. However, sibling code paths still interpolate `SERVER_CONFIG.publicUrl` raw: the authorization-server metadata route (`src/http/server.ts:152-159`, `issuer`/`*_endpoint`) and the `oauth_authorize_url`/`projects_url` fields in the 401 bodies (`server.ts:780`, `:788-789`). A `MCP_SERVER_URL` ending in `/` would still yield `https://host//oauth/authorize` there. Pre-existing and out of this PR's scope (prod uses a no-slash URL), but noting it since the PR introduces the normalization concept — a follow-up could route these through the same helper.
- **`scopes_supported` divergence.** The resource documents advertise `['mcp:read','mcp:write']` (`auth-challenge.ts:52`) while the AS metadata advertises `OAUTH_CONFIG.supportedScopes` = `['mcp:read','mcp:write','project:select']` (`server.ts:163`). Pre-existing (`master` had the same two-scope list), not introduced here.

---

### Dimension coverage

- **Software engineering:** Strong. New logic extracted to `auth-challenge.ts` specifically so it's importable without triggering `startServer()` — a deliberate, well-reasoned seam. `auth-challenge.test.ts` (134 lines) exercises URL construction, §3.3 self-consistency, endpoint scoping, header-before-body ordering, and the trailing-slash edge case (and correctly calls out that the *old* trailing-slash test asserted a guarantee it never fed). Import style (`.js` extensions), Express typing, and naming all match repo conventions. All 5 MCP-resource 401 sites migrated to `sendUnauthorized`; the two `/api/projects` 401s (`server.ts:677`,`:701`) are correctly left as plain 401s per the PR's stated rationale.
- **Functionality:** Solves the stated problem. RFC 9728 §3.1 URL construction (`/.well-known/oauth-protected-resource/mcp`) and §3.3 `resource`↔URL identity were verified by hand for all three documents (bare, `/mcp`, `/sse`) — each round-trips. The three new exact-match routes don't collide with each other or existing routes. `authorization_servers: [<base>]` is consistent with the AS metadata `issuer` (`server.ts:155`). No existing test depended on the old single-document shape.
- **Security:** Positive. Discovery URLs and `resource` are derived from `SERVER_CONFIG.publicUrl` (env), not the request `Host` header, preserving the existing anti-spoofing property. No secrets/PII newly logged or returned. No new dependencies. Status codes and bodies for header-ignoring clients are unchanged.
- **Performance:** Negligible — a few string operations per 401; no new I/O, queries, or allocations in any hot path.

---

Zero Critical findings — the two items above are a non-blocking suggestion and minor notes the author may take or leave. Clean, spec-faithful, and honestly documented (including the accurate "lands its benefit on deploy, not on merge" caveat). Posting as a COMMENT; explicit GitHub approval remains a human action.

<sub>🤖 Automated review by the InsForge review bot.</sub>

tonychang04:
`d29b115` — @manager is right, and the blast radius is wider than either of us said. **Ten consumers, not two.**

`origin()` only guarded the protected-resource documents. Everything else took `SERVER_CONFIG.publicUrl` raw:

```
config.ts   OAUTH_CONFIG.callbackUrl        registered with the platform
server.ts   AS metadata baseUrl             issuer, authorize, token, revoke,
                                            registration_endpoint
server.ts   project-selection redirect
server.ts   two oauth_authorize_url values in 401 bodies
server.ts   projects_url
server.ts   the startup banner, twice
```

A configured `MCP_SERVER_URL` ending in `/` gives every one a doubled slash, and Express matches none of those routes. **A deploy that set the variable that way would advertise `//oauth/register`, register a doubled callback with the platform, and look entirely healthy while no client could log in** — the same "green but broken" shape as the gateway `/health` placeholder.

Fixed where `publicUrl` is defined, as an exported `normaliseBaseUrl()` so the rule has a name and a test rather than living inline. `origin()` is now redundant rather than load-bearing; left in place because it takes an explicit `publicUrl` in its own tests.

**One thing worth knowing for the next person:** the test calls the pure function rather than re-importing `config.ts`. That module builds a commander program at import time, so `vi.resetModules()` + re-import throws on the duplicate `--version` flag. My first attempt did exactly that and failed — config is not testable by re-import.

Verified both directions with a real trailing slash before writing the test:

```
MCP_SERVER_URL='https://mcp.insforge.dev/'
  publicUrl    https://mcp.insforge.dev
  callbackUrl  https://mcp.insforge.dev/oauth/callback
  registration https://mcp.insforge.dev/oauth/register

MCP_SERVER_URL='https://mcp.insforge.dev'   (unchanged)
```

33 tests, lint clean. This dismisses the approval as expected — sending a fresh review request.

tonychang04:
`a7dc73e` — @qa-bot's requested test, plus the guard that matters more.

**Three assertions over every advertised URL** given a slashed public URL: the protected-resource documents (`resource` and `authorization_servers`), the metadata URL a client derives, and the `resource_metadata` inside the `WWW-Authenticate` challenge. Doubled slashes checked *after* the scheme so `https://` isn't a false hit.

**And the guard that's stronger than any of them:** nothing outside `config.ts` may read `process.env.MCP_SERVER_URL`. Since `publicUrl` is normalised where it's defined, every consumer is safe by construction — the only way to reintroduce this is to read the raw variable somewhere else. That's now a test failure. Verified by adding a module that does exactly that: the guard fails, and passes again when it's removed.

### One honest note on coverage

It nearly slipped past me, and it's worth stating rather than letting the commit imply otherwise.

**The three advertised-URL tests pass a slashed `publicUrl` explicitly, so they exercise `origin()` in `auth-challenge.ts` — not the config-level normalisation.** I found this by neutering both and watching which tests moved: with `origin()` stubbed out they fail *while the config fix is still in place*.

So the actual coverage map is:

```
normaliseBaseUrl unit test   → the config-level fix          ✔
advertised-URL tests         → origin() in the helpers       ✔
bypass guard                 → reintroduction via raw env    ✔
```

Together that's complete, but no single one of them covers everything, and the three Quinn asked for aren't the ones covering the fix he asked them about. Same shape as the `globalThis.fetch` mock and the handshake success path — a test that looks like it validates the change while validating something adjacent.

37 tests across the suite, lint clean. @qa-bot — chain is ready for your re-verify; the trailing-slash end-to-end run is the one I can't do here, since booting the server needs Redis.

jwfing (commented):
## Review: fix(oauth) — WWW-Authenticate + per-endpoint resource metadata

**Summary:** A clean, well-tested, spec-grounded fix that makes OAuth discoverable to spec-compliant MCP clients; no blocking issues, one worthwhile follow-up for browser-based clients.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (nor `docs/specs/`) directory. Assessed against the PR description, the referenced RFCs (9728 §3.1/§3.3/§5.1, 6750), and surrounding code. I verified the changes locally: `tsc --noEmit` clean, `eslint` clean, **32/32 vitest pass** (14 of them new in `auth-challenge.test.ts`) — the verification claims in the PR body hold.

I confirmed the scope claims against the tree:
- All five 401 sites at MCP-resource endpoints (`server.ts:777, 785, 808, 957, 964`) now route through `sendUnauthorized`.
- The two 401s on `/api/projects` (`server.ts:677, 701`) correctly keep their plain 401 — not MCP resource endpoints.
- `GET /mcp`, `DELETE /mcp`, and `POST /messages` have no auth-based 401 sites (they gate on session id → 400/404), so nothing was missed there.
- The three metadata routes (`server.ts:176-194`) are exact-match paths — no route shadowing between `/.well-known/oauth-protected-resource/{mcp,sse}` and the bare path.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — browser MCP clients can't read the new header (CORS).** `src/http/server.ts:51` sets `Access-Control-Expose-Headers: 'Mcp-Session-Id'`. `WWW-Authenticate` is not a CORS-safelisted response header, so any browser/JS MCP client (the middleware sets `Access-Control-Allow-Origin: *`, so browser clients are clearly in scope) will get the 401 but `response.headers.get('WWW-Authenticate')` returns `null` — the very discovery signal this PR adds is invisible to that client class. Native clients (Claude Desktop, CLI) are unaffected and do get the benefit. Suggested one-line fix: add `WWW-Authenticate` to the expose list. Non-blocking because the primary consumers today are native.
- **Software engineering — no test exercises the actual route wiring.** The 14 new tests cover the pure helpers thoroughly, but nothing asserts that `GET /.well-known/oauth-protected-resource/mcp` is actually registered and returns 200 with a matching `resource`. A typo in the route-path concatenation at `server.ts:176-189` would pass all unit tests. I understand `server.ts` can't be imported (it calls `startServer()` at import time, as noted); a small supertest-style harness or extracting the route table would close the gap. Low priority given the on-the-wire verification in the PR body.

#### Information
- **RFC 6750 `error` parameter on token-present 401s.** For `project_binding_required` (`server.ts:785, 964`) and `invalid_credentials` (`server.ts:808`) a bearer token *was* supplied, so RFC 6750 §3 suggests the challenge also carry `error="invalid_token"` / `insufficient_scope`. The `resource_metadata` parameter alone satisfies discovery, so this is purely a nice-to-have for stricter clients; the no-token `authentication_required` case correctly omits `error`.
- **Security — good.** `WWW-Authenticate` and every `resource`/`authorization_servers` value derive from `SERVER_CONFIG.publicUrl` (env), never the request `Host` header, so the earlier host-spoofing concern is preserved. Status codes and bodies are unchanged, no secrets or tokens are logged or added to responses, and no auth check is weakened. Header is set before the body in `sendUnauthorized`, and the ordering is regression-tested.
- **Performance — no concerns.** Pure string construction on the 401 path; no new I/O, queries, or allocations of note.
- Nice touch: the trailing-slash normalization funnels the derived URL, `resource`, and `authorization_servers` through one `origin()` helper so they can't diverge, and the rewritten trailing-slash test now actually feeds `/` and `//` (the prior test asserted the guarantee without exercising it).

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the two Suggestions are worth a follow-up but do not block merge. Note the PR's own caveat stands: the benefit lands on deploy, not on merge, since `mcp.insforge.dev` is not yet running this code.

PR #71, opened 2026-07-28
fix(http): reap orphaned runtime sessions and keep SSE streams alive

tonychang04:
Fixes the memory growth behind `/health` reporting **510 active sessions against 4443 held in memory** on the deployed box — the 4GB instance that also serves customer Postgres. Assigned out of the port thread; deliberately **not** folded into PR2, because PR2 is gated on the deploy question and this shouldn't wait for an engine swap.

## 1. Session leak (the OOM risk)

Redis owns session lifetime via `SESSION_TTL`, but the in-memory `McpServer` + transport are only dropped on an explicit `DELETE /mcp`.

- SSE cleans up properly on `res.on('close')`.
- **Streamable HTTP has no equivalent signal**, and clients rarely send `DELETE`. So every session created since process start stays resident, each holding a server, a transport, and a full 17-tool registry. There was no sweeper and no idle eviction — `deleteSession` was reachable only from `DELETE /mcp`, the SSE close handler, and shutdown.

That is exactly the observed shape: Redis expires records on TTL, memory never does, so `memorySessionCount` climbs away from `activeSessions`.

**Fix:** Redis is already the source of truth, so read from it. Every 5 minutes, pipeline `EXISTS` across the resident session ids and close the ones whose record has gone. A failed *or absent* reply counts as **alive**, so a transient Redis error can never be read as "reap these live sessions". Timer is `unref()`'d and stopped on shutdown.

## 2. SSE keepalive

An idle SSE stream carries no bytes, so the load balancer closes it on its idle timeout (60s by default on an ALB) and the client sees a dropped connection. There was no keepalive of any kind. Now a comment frame every 25s — ignored by the event-stream parser and by the transport's own framing, so it's invisible to clients.

## Verification

Against a **real Redis**, using the built server:

```
session created                    → {activeSessions: 1, memorySessionCount: 1}
record expires, no DELETE sent     → {activeSessions: 0, memorySessionCount: 1}   ← the production signature, reproduced
after one sweep                    → {activeSessions: 0, memorySessionCount: 0}
[SessionManager] Swept 1 orphaned session(s); 0 remain in memory
```

Idle `/sse` stream over 4s emits `: keepalive` frames as expected.

28/28 vitest (10 new), `tsc` and `eslint` clean, build succeeds. The sweep tests cover the cases that matter: only-expired-are-reaped, Redis errors reap nothing, an empty pipeline reply reaps nothing, 45-of-50 orphans reclaimed in one pass, and the timer being idempotent, stoppable and rejection-safe.

## Blast radius

The running HTTP server. The sweep only closes **streamable** sessions Redis has already expired; SSE is left alone, since it has its own disconnect signal and cleans up on `res` `close`. (An earlier version of this note claimed such a client would be rejected on its next request anyway — that was wrong: `getStreamableSession` reads memory first, so it would have been served without consulting Redis. Corrected after review.) Both intervals take an env override (`MCP_SESSION_SWEEP_MS`, `MCP_SSE_KEEPALIVE_MS`) so they can be tuned to an environment's LB timeout without a redeploy of code.

## Note on deploying it

Per the thread, `mcp.insforge.dev` is 139 days stale, has no deploy path, and shares an instance with customer Postgres on 5432. **This PR doesn't change when it's safe to deploy** — it just means the leak is fixed whenever that question is resolved.

---
## Summary by cubic
Fixes the runtime session leak, keeps SSE connections alive behind ALBs, and adds memory stats to `/health`. The reaper now gates on in‑memory activity, ignores Redis wipes, re‑checks before closing, and never closes sessions while a stream is open.

- **Bug Fixes**
  - Session reaper (streamable only): runs every 5m; pipelines `EXISTS`; error/empty/non‑integer replies count as alive; reaps only when the Redis record is missing AND in‑memory `lastSeenAt` is past TTL AND there are no open streams; re‑checks the live entry before closing; warns when records vanish inside TTL; timer is unref’d and stopped on shutdown. Tracks stream open/close on `GET /mcp`, touches on open to restart TTL, and resets the idle clock on close.
  - SSE: writes a “: keepalive” comment every 25s and clears on close; refreshes session TTL on `POST /messages` so active SSE sessions don’t expire; never swept by the reaper.
  - Config: intervals tunable via `MCP_SESSION_SWEEP_MS` and `MCP_SSE_KEEPALIVE_MS`; overrides must be positive integers; invalid values fall back to defaults.

- **New Features**
  - `/health` now reports `heapUsedMb`, `heapLimitMb`, `heapUsedPct`, and `rssMb`.

<sup>Written for commit a1b7abe0a5ddf92533e3e252daaada41be905229. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/71?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable legacy SSE keep-alives and startup/shutdown idle session sweeping via environment-driven timing constants.
  * Enhanced the health endpoint with heap/memory metrics.
* **Bug Fixes**
  * Implemented TTL-gated orphan reaping for streamable runtime sessions when Redis records disappear, without affecting active SSE sessions.
  * Refreshes session activity on message arrival in a non-blocking manner.
* **Tests**
  * Added Vitest coverage for env parsing, SSE keepalive behavior, and session reaping/idle sweep timer controls.

### DIFF
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -98,6 +98,19 @@ export const REDIS_CONFIG = {
 // Session Configuration
 // ============================================================================
 
+/**
+ * How often to write a comment frame on an idle SSE stream. Must stay well
+ * under the load balancer's idle timeout (60s by default on an AWS ALB) or the
+ * stream is closed underneath a connected client.
+ */
+export const SSE_KEEPALIVE_MS = parseInt(process.env.MCP_SSE_KEEPALIVE_MS || '') || 25 * 1000;
+
+/**
+ * How often to reap runtime sessions whose Redis record has expired.
+ * Overridable so the reaper can be exercised without waiting minutes.
+ */
+export const SESSION_SWEEP_MS = parseInt(process.env.MCP_SESSION_SWEEP_MS || '') || 5 * 60 * 1000;
+
 export const SESSION_CONFIG = {
   /** Session TTL in seconds (24 hours) */
   ttl: 24 * 60 * 60,
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -99,17 +99,27 @@ export const REDIS_CONFIG = {
 // ============================================================================
 
 /**
+ * Positive-integer env override. A zero, a negative or an unparseable value
+ * falls back to the default rather than reaching setInterval, which clamps a
+ * non-positive delay to 1ms and would spin.
+ */
+export function positiveIntEnv(value: string | undefined, fallback: number): number {
+  const parsed = parseInt(value || '', 10);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+}
+
+/**
  * How often to write a comment frame on an idle SSE stream. Must stay well
  * under the load balancer's idle timeout (60s by default on an AWS ALB) or the
  * stream is closed underneath a connected client.
  */
-export const SSE_KEEPALIVE_MS = parseInt(process.env.MCP_SSE_KEEPALIVE_MS || '') || 25 * 1000;
+export const SSE_KEEPALIVE_MS = positiveIntEnv(process.env.MCP_SSE_KEEPALIVE_MS, 25 * 1000);
 
 /**
  * How often to reap runtime sessions whose Redis record has expired.
  * Overridable so the reaper can be exercised without waiting minutes.
  */
-export const SESSION_SWEEP_MS = parseInt(process.env.MCP_SESSION_SWEEP_MS || '') || 5 * 60 * 1000;
+export const SESSION_SWEEP_MS = positiveIntEnv(process.env.MCP_SESSION_SWEEP_MS, 5 * 60 * 1000);
 
 export const SESSION_CONFIG = {
   /** Session TTL in seconds (24 hours) */
--- a/src/http/keepalive.test.ts
+++ b/src/http/keepalive.test.ts
@@ -1,0 +1,128 @@
+import { describe, it, expect, vi, afterEach, type Mock } from 'vitest';
+import { positiveIntEnv } from './config.js';
+
+/**
+ * The keepalive itself is a few lines inline in the SSE handler, and server.ts
+ * calls startServer() at import time so it can't be imported here. These cover
+ * the behaviour that matters — write on tick, stop on close, never write to a
+ * finished response — against the same shape the handler uses, plus the env
+ * parsing that decides the interval.
+ */
+
+type FakeRes = {
+  writableEnded: boolean;
+  write: Mock<(chunk: string) => void>;
+  handlers: Record<string, () => void>;
+  on: (event: string, fn: () => void) => void;
+  emit: (event: string) => void;
+};
+
+function fakeRes(): FakeRes {
+  const res: FakeRes = {
+    writableEnded: false,
+    write: vi.fn<(chunk: string) => void>(),
+    handlers: {},
+    on(event, fn) {
+      res.handlers[event] = fn;
+    },
+    emit(event) {
+      res.handlers[event]?.();
+    },
+  };
+  return res;
+}
+
+/** Mirrors the handler: start an interval, clear it when the response closes. */
+function attachKeepAlive(res: FakeRes, intervalMs: number) {
+  const timer = setInterval(() => {
+    if (res.writableEnded) return;
+    try {
+      res.write(': keepalive\n\n');
+    } catch {
+      /* handler logs; irrelevant here */
+    }
+  }, intervalMs);
+  timer.unref?.();
+  res.on('close', () => clearInterval(timer));
+  return timer;
+}
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+describe('SSE keepalive', () => {
+  it('writes a comment frame on each tick', () => {
+    vi.useFakeTimers();
+    const res = fakeRes();
+    attachKeepAlive(res, 1000);
+
+    vi.advanceTimersByTime(3000);
+
+    expect(res.write).toHaveBeenCalledTimes(3);
+    // A comment frame: ignored by the event-stream parser, so it cannot be
+    // mistaken for protocol traffic by a client.
+    expect(res.write).toHaveBeenCalledWith(': keepalive\n\n');
+  });
+
+  it('stops writing once the response closes', () => {
+    vi.useFakeTimers();
+    const res = fakeRes();
+    attachKeepAlive(res, 1000);
+
+    vi.advanceTimersByTime(2000);
+    expect(res.write).toHaveBeenCalledTimes(2);
+
+    res.emit('close');
+    vi.advanceTimersByTime(10000);
+
+    expect(res.write).toHaveBeenCalledTimes(2);
+  });
+
+  it('does not write to a response that has already ended', () => {
+    vi.useFakeTimers();
+    const res = fakeRes();
+    attachKeepAlive(res, 1000);
+
+    res.writableEnded = true;
+    vi.advanceTimersByTime(5000);
+
+    expect(res.write).not.toHaveBeenCalled();
+  });
+
+  it('keeps ticking when a write throws', () => {
+    vi.useFakeTimers();
+    const res = fakeRes();
+    res.write.mockImplementation(() => {
+      throw new Error('EPIPE');
+    });
+    attachKeepAlive(res, 1000);
+
+    expect(() => vi.advanceTimersByTime(3000)).not.toThrow();
+    expect(res.write).toHaveBeenCalledTimes(3);
+  });
+});
+
+describe('positiveIntEnv', () => {
+  it('uses the fallback when unset', () => {
+    expect(positiveIntEnv(undefined, 25_000)).toBe(25_000);
+  });
+
+  it('accepts a positive override', () => {
+    expect(positiveIntEnv('1500', 25_000)).toBe(1500);
+  });
+
+  it.each(['-1', '0', 'abc', '', '  '])(
+    'falls back for %j rather than passing it to setInterval',
+    (value) => {
+      // setInterval clamps a non-positive delay to 1ms, which would spin
+      // writing keepalive frames as fast as the event loop allows.
+      expect(positiveIntEnv(value, 25_000)).toBe(25_000);
+    }
+  );
+
+  it('ignores trailing junk the way parseInt does, but still demands a number', () => {
+    expect(positiveIntEnv('1500ms', 25_000)).toBe(1500);
+    expect(positiveIntEnv('ms1500', 25_000)).toBe(25_000);
+  });
+});
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -1083,6 +1083,16 @@ app.post(SSE_ENDPOINTS.messages, async (req: Request, res: Response) => {
     });
   }
 
+  // Streamable HTTP refreshes the Redis record on every request; the SSE path
+  // did not, so an SSE record lapsed 24h after creation however active the
+  // client was, and /health under-reported live SSE sessions. Fire-and-forget:
+  // a failed refresh must not fail the message.
+  getSessionManager()
+    .touchSession(sessionId)
+    .catch((error) => {
+      console.error(`[SSE] Failed to refresh session ${sessionId}:`, error);
+    });
+
   await transport.handlePostMessage(req, res, req.body);
 });
 
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -1,6 +1,7 @@
 import 'dotenv/config';
 import express, { Request, Response, NextFunction } from 'express';
 import { randomUUID, createHash } from 'crypto';
+import v8 from 'node:v8';
 
 // Transport imports
 import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
@@ -126,6 +127,25 @@ function extractLegacyHeaders(req: Request): { apiKey?: string; apiBaseUrl?: str
   };
 }
 
+/**
+ * Heap numbers for the health payload.
+ *
+ * The session leak had to be diagnosed by counting objects and multiplying by
+ * a separately-measured size, because the process publishes no memory metric
+ * at all — so "how long until it dies" could only ever be an estimate. V8's
+ * own limit is the number that actually decides that, and exposing it costs
+ * nothing.
+ */
+function heapStats() {
+  const { heap_size_limit, used_heap_size } = v8.getHeapStatistics();
+  return {
+    heapUsedMb: Math.round(used_heap_size / 1024 / 1024),
+    heapLimitMb: Math.round(heap_size_limit / 1024 / 1024),
+    heapUsedPct: Math.round((used_heap_size / heap_size_limit) * 100),
+    rssMb: Math.round(process.memoryUsage().rss / 1024 / 1024),
+  };
+}
+
 // ============================================================================
 // Health & Discovery Endpoints
 // ============================================================================
@@ -143,6 +163,7 @@ app.get(API_ENDPOINTS.health, async (_req: Request, res: Response) => {
       sse: '2024-11-05 (deprecated)',
     },
     sessions: stats,
+    memory: heapStats(),
     authentication: 'OAuth Bearer Token',
   });
 });
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -16,6 +16,8 @@ import {
   OAUTH_CONFIG,
   STREAMABLE_HTTP_ENDPOINTS,
   SSE_ENDPOINTS,
+  SSE_KEEPALIVE_MS,
+  SESSION_SWEEP_MS,
   OAUTH_ENDPOINTS,
   API_ENDPOINTS,
   isOAuthConfigured,
@@ -978,9 +980,24 @@ app.get(SSE_ENDPOINTS.sse, async (req: Request, res: Response) => {
 
   console.log(`[SSE] Session created: ${transport.sessionId}, Project: ${validProjectInfo.projectName}`);
 
+  // An idle SSE stream carries no bytes, so the load balancer closes it on its
+  // idle timeout and the client sees the connection drop. A comment frame is
+  // ignored by the event-stream parser and by the transport's own framing, so
+  // it keeps the connection warm without being visible to the client.
+  const keepAlive = setInterval(() => {
+    if (res.writableEnded) return;
+    try {
+      res.write(': keepalive\n\n');
+    } catch (error) {
+      console.error(`[SSE] Keepalive write failed for ${transport.sessionId}:`, error);
+    }
+  }, SSE_KEEPALIVE_MS);
+  keepAlive.unref();
+
   // Clean up on close
   res.on('close', () => {
     console.log(`[SSE] Session closed: ${transport.sessionId}`);
+    clearInterval(keepAlive);
     sseTransports.delete(transport.sessionId);
 
     // Clean up the session from SessionManager (async with error handling)
@@ -1083,6 +1100,10 @@ async function startServer() {
     await redis.ping();
     console.log('[Redis] Connection verified');
 
+    // Runtime sessions are only dropped on an explicit DELETE, which Streamable
+    // HTTP clients rarely send. Reap the ones Redis has already expired.
+    getSessionManager().startIdleSweep(SESSION_SWEEP_MS);
+
     const server = app.listen(SERVER_CONFIG.port, SERVER_CONFIG.host, () => {
       const redisConfig = getRedisConfig();
       console.log(`
@@ -1166,6 +1187,7 @@ async function startServer() {
         // Close all MCP sessions
         try {
           const sessionManager = getSessionManager();
+          sessionManager.stopIdleSweep();
           await sessionManager.closeAllSessions();
         } catch (error) {
           console.error('[Shutdown] Error closing sessions:', error);
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -899,6 +899,13 @@ app.get(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
     });
   }
 
+  // This stream is the only sign of life for a client that opens it and then
+  // sends nothing. Hold the session for as long as it is open, and restart the
+  // Redis record's clock now so a restart can still restore the session.
+  sessionManager.openStream(sessionId);
+  res.on('close', () => sessionManager.closeStream(sessionId));
+  await sessionManager.touchSession(sessionId);
+
   await runtime.transport.handleRequest(req, res, req.body);
 });
 
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -1118,7 +1118,7 @@ app.post(SSE_ENDPOINTS.messages, async (req: Request, res: Response) => {
   getSessionManager()
     .touchSession(sessionId)
     .catch((error) => {
-      console.error(`[SSE] Failed to refresh session ${sessionId}:`, error);
+      console.error(`[SSE] Failed to refresh session ${tokenFingerprint(sessionId)}`, error);
     });
 
   await transport.handlePostMessage(req, res, req.body);
--- a/src/http/session-manager.test.ts
+++ b/src/http/session-manager.test.ts
@@ -15,7 +15,13 @@ vi.mock('./redis.js', () => ({
   }),
 }));
 
-const { SessionManager, selectOrphanedSessions, SESSION_TTL_MS, monotonicNow } = await import('./session-manager.js');
+const {
+  SessionManager,
+  selectOrphanedSessions,
+  selectPrematurelyMissing,
+  SESSION_TTL_MS,
+  monotonicNow,
+} = await import('./session-manager.js');
 
 /** A session whose last traffic is old enough that its record should have expired. */
 const STALE = -(SESSION_TTL_MS + 1000);
@@ -29,6 +35,7 @@ function fakeRuntime(transportType: 'streamable' | 'sse' = 'streamable', ageOffs
     transport: { close: vi.fn().mockResolvedValue(undefined) },
     transportType,
     lastSeenAt: monotonicNow() + ageOffsetMs,
+    openStreams: 0,
   };
 }
 
@@ -52,7 +59,7 @@ describe('selectOrphanedSessions', () => {
   const now = 1_800_000_000_000;
   /** Sessions whose records should already have expired on their own. */
   const aged = (...ids: string[]) =>
-    ids.map((sessionId) => ({ sessionId, lastSeenAt: now - SESSION_TTL_MS - 1000 }));
+    ids.map((sessionId) => ({ sessionId, lastSeenAt: now - SESSION_TTL_MS - 1000, openStreams: 0 }));
 
   it('picks exactly the ids whose Redis record is gone', () => {
     expect(selectOrphanedSessions(aged('a', 'b', 'c'), [1, 0, 1], now)).toEqual(['b']);
@@ -68,21 +75,63 @@ describe('selectOrphanedSessions', () => {
 
   it('holds a session whose record vanished well inside its TTL', () => {
     // Redis losing data, not Redis expiring a session.
-    const active = [{ sessionId: 'busy', lastSeenAt: now - 60_000 }];
+    const active = [{ sessionId: 'busy', lastSeenAt: now - 60_000, openStreams: 0 }];
     expect(selectOrphanedSessions(active, [0], now)).toEqual([]);
   });
 
   it('separates the wiped-but-active from the genuinely abandoned', () => {
     const mixed = [
-      { sessionId: 'active', lastSeenAt: now - 60_000 },
-      { sessionId: 'abandoned', lastSeenAt: now - SESSION_TTL_MS - 1 },
+      { sessionId: 'active', lastSeenAt: now - 60_000, openStreams: 0 },
+      { sessionId: 'abandoned', lastSeenAt: now - SESSION_TTL_MS - 1, openStreams: 0 },
     ];
     expect(selectOrphanedSessions(mixed, [0, 0], now)).toEqual(['abandoned']);
   });
 
+  it('holds a session with an open stream, however old both clocks are', () => {
+    // The client is plainly connected — it just has nothing to say.
+    const streaming = [
+      { sessionId: 'listening', lastSeenAt: now - SESSION_TTL_MS * 10, openStreams: 1 },
+    ];
+    expect(selectOrphanedSessions(streaming, [0], now)).toEqual([]);
+  });
+
+  it('reaps once the last stream has closed', () => {
+    const closed = [
+      { sessionId: 'was-listening', lastSeenAt: now - SESSION_TTL_MS - 1, openStreams: 0 },
+    ];
+    expect(selectOrphanedSessions(closed, [0], now)).toEqual(['was-listening']);
+  });
+
   it('reaps a session exactly at the TTL boundary', () => {
-    const boundary = [{ sessionId: 'edge', lastSeenAt: now - SESSION_TTL_MS }];
+    const boundary = [{ sessionId: 'edge', lastSeenAt: now - SESSION_TTL_MS, openStreams: 0 }];
     expect(selectOrphanedSessions(boundary, [0], now)).toEqual(['edge']);
+  });
+});
+
+describe('selectPrematurelyMissing', () => {
+  const now = 1_800_000_000_000;
+
+  it('flags a record that vanished while the session was still inside its TTL', () => {
+    const s = [{ sessionId: 'busy', lastSeenAt: now - 60_000, openStreams: 0 }];
+    expect(selectPrematurelyMissing(s, [0], now)).toEqual(['busy']);
+  });
+
+  it('says nothing about a record that expired on schedule', () => {
+    // Normal expiry is not evidence of anything and must not send an operator
+    // looking for a Redis fault.
+    const s = [{ sessionId: 'old', lastSeenAt: now - SESSION_TTL_MS - 1, openStreams: 0 }];
+    expect(selectPrematurelyMissing(s, [0], now)).toEqual([]);
+  });
+
+  it('does not flag a long-open stream whose record expired normally', () => {
+    // Held by its stream, not by the TTL gate — so it is not a Redis symptom.
+    const s = [{ sessionId: 'listening', lastSeenAt: now - SESSION_TTL_MS - 1, openStreams: 1 }];
+    expect(selectPrematurelyMissing(s, [0], now)).toEqual([]);
+  });
+
+  it('says nothing when the records are present', () => {
+    const s = [{ sessionId: 'fine', lastSeenAt: now - 60_000, openStreams: 0 }];
+    expect(selectPrematurelyMissing(s, [1], now)).toEqual([]);
   });
 });
 
@@ -181,6 +230,59 @@ describe('SessionManager.sweepOrphanedSessions', () => {
 
     await expect(manager.sweepOrphanedSessions()).resolves.toBe(45);
     expect(manager.getActiveSessionIds()).toHaveLength(5);
+  });
+
+  it('never reaps a streamable session while its stream is open', async () => {
+    // Quinn's case: the client holds GET /mcp open and sends no POSTs, so the
+    // Redis record lapses and lastSeenAt is frozen at the moment it opened.
+    // Both clocks agree to reap a client that is still connected.
+    const manager = new SessionManager();
+    const runtimes = seed(manager, ['listening'], 'streamable', STALE);
+    manager.openStream('listening');
+    pipelineExec.mockResolvedValue([[null, 0]]);
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+    expect(manager.getActiveSessionIds()).toEqual(['listening']);
+    expect(runtimes.get('listening')!.transport.close).not.toHaveBeenCalled();
+  });
+
+  it('collects the session once the stream closes and it goes idle', async () => {
+    const manager = new SessionManager();
+    seed(manager, ['listening'], 'streamable', STALE);
+    manager.openStream('listening');
+    manager.closeStream('listening');
+    // closeStream restarts the idle clock, so it is not immediately collectable.
+    pipelineExec.mockResolvedValue([[null, 0]]);
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+
+    // Age it past the TTL again and it goes, exactly as an idle session should.
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (manager as any).runtimeSessions.get('listening').lastSeenAt = monotonicNow() + STALE;
+    pipelineExec.mockResolvedValue([[null, 0]]);
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(1);
+    expect(manager.getActiveSessionIds()).toEqual([]);
+  });
+
+  it('counts overlapping streams so a reconnect cannot unprotect the session', async () => {
+    // A client may open its replacement stream before the old response has
+    // finished closing. A boolean flag cleared by the departing one would drop
+    // the guard while a live stream is still attached.
+    const manager = new SessionManager();
+    seed(manager, ['reconnecting'], 'streamable', STALE);
+    manager.openStream('reconnecting');
+    manager.openStream('reconnecting');
+    manager.closeStream('reconnecting');
+    expect(manager.getOpenStreamCount('reconnecting')).toBe(1);
+
+    pipelineExec.mockResolvedValue([[null, 0]]);
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+  });
+
+  it('ignores a stream close for a session that is already gone', async () => {
+    // Close events land after a reap or an explicit DELETE all the time.
+    const manager = new SessionManager();
+    expect(() => manager.closeStream('never-existed')).not.toThrow();
+    expect(manager.getOpenStreamCount('never-existed')).toBe(0);
   });
 
   it('keeps every session when Redis comes back empty', async () => {
--- a/src/http/session-manager.test.ts
+++ b/src/http/session-manager.test.ts
@@ -14,19 +14,19 @@ vi.mock('./redis.js', () => ({
 const { SessionManager, selectOrphanedSessions } = await import('./session-manager.js');
 
 /** Stand-in for a runtime entry; only close() is exercised by the sweep. */
-function fakeRuntime() {
+function fakeRuntime(transportType: 'streamable' | 'sse' = 'streamable') {
   return {
     server: { close: vi.fn().mockResolvedValue(undefined) },
     transport: { close: vi.fn().mockResolvedValue(undefined) },
-    transportType: 'streamable' as const,
+    transportType,
   };
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-function seed(manager: any, ids: string[]) {
+function seed(manager: any, ids: string[], transportType: 'streamable' | 'sse' = 'streamable') {
   const runtimes = new Map<string, ReturnType<typeof fakeRuntime>>();
   for (const id of ids) {
-    const rt = fakeRuntime();
+    const rt = fakeRuntime(transportType);
     runtimes.set(id, rt);
     manager.runtimeSessions.set(id, rt);
   }
@@ -97,6 +97,38 @@ describe('SessionManager.sweepOrphanedSessions', () => {
 
     await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
     expect(manager.getActiveSessionIds()).toEqual(['a']);
+  });
+
+  it('never reaps an SSE session, even with no Redis record', async () => {
+    // SSE cleans up on res 'close' and nothing on its message path used to
+    // refresh the record, so an SSE record lapses at TTL from creation while
+    // the connection is still open and keepalive-warm. Sweeping it would drop
+    // a live client.
+    const manager = new SessionManager();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const sse = seed(manager as any, ['sse-live'], 'sse');
+    pipelineExec.mockResolvedValue([[null, 0]]);
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+    expect(manager.getActiveSessionIds()).toContain('sse-live');
+    expect(sse.get('sse-live')!.transport.close).not.toHaveBeenCalled();
+    // Nothing to ask Redis about, so no pipeline should have run at all.
+    expect(pipelineExec).not.toHaveBeenCalled();
+  });
+
+  it('reaps an expired streamable session while leaving an SSE one alone', async () => {
+    const manager = new SessionManager();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const streamable = seed(manager as any, ['http-dead'], 'streamable');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const sse = seed(manager as any, ['sse-live'], 'sse');
+    // Only the streamable id is queried, so a single reply is correct.
+    pipelineExec.mockResolvedValue([[null, 0]]);
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(1);
+    expect(manager.getActiveSessionIds()).toEqual(['sse-live']);
+    expect(streamable.get('http-dead')!.transport.close).toHaveBeenCalled();
+    expect(sse.get('sse-live')!.transport.close).not.toHaveBeenCalled();
   });
 
   it('reclaims an accumulation of orphans in one pass', async () => {
--- a/src/http/session-manager.test.ts
+++ b/src/http/session-manager.test.ts
@@ -380,3 +380,61 @@ describe('SessionManager idle sweep timer', () => {
     logged.mockRestore();
   });
 });
+
+describe('EXISTS replies that are not a number', () => {
+  beforeEach(() => {
+    pipelineExec.mockReset();
+    redisDel.mockReset().mockResolvedValue(1);
+    redisGet.mockReset().mockResolvedValue(null);
+    redisSetex.mockReset().mockResolvedValue('OK');
+  });
+
+  it('treats a null reply as alive, not as an expired session', async () => {
+    // Number(null) is 0, so a bare cast turns "Redis said something odd" into
+    // "reap this". EXISTS answers with an integer or nothing useful at all.
+    const manager = new SessionManager();
+    seed(manager, ['a', 'b'], 'streamable', STALE);
+    pipelineExec.mockResolvedValue([
+      [null, null],
+      [null, undefined],
+    ]);
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+    expect(manager.getActiveSessionIds()).toEqual(['a', 'b']);
+  });
+
+  it('still reaps on an honest zero', async () => {
+    const manager = new SessionManager();
+    seed(manager, ['gone'], 'streamable', STALE);
+    pipelineExec.mockResolvedValue([[null, 0]]);
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(1);
+  });
+});
+
+describe('a session touched between the EXISTS round trip and the delete', () => {
+  beforeEach(() => {
+    pipelineExec.mockReset();
+    redisDel.mockReset().mockResolvedValue(1);
+    redisGet.mockReset().mockResolvedValue(null);
+    redisSetex.mockReset().mockResolvedValue('OK');
+  });
+
+  it('is not closed, because the live entry is re-checked', async () => {
+    // The candidate list is a snapshot taken before the pipeline. A request
+    // arriving in that window stamps lastSeenAt and rewrites the record, so
+    // acting on the snapshot would close a session that just proved it is live.
+    const manager = new SessionManager();
+    const runtimes = seed(manager, ['racing'], 'streamable', STALE);
+
+    pipelineExec.mockImplementation(async () => {
+      // The request lands while Redis is answering.
+      await manager.touchSession('racing');
+      return [[null, 0]];
+    });
+
+    await expect(manager.sweepOrphanedSessions()).resolves.toBe(0);
+    expect(manager.getActiveSessionIds()).toEqual(['racing']);
+    expect(runtimes.get('racing')!.transport.close).not.toHaveBeenCalled();
+  });
+});
--- a/src/http/session-manager.test.ts
+++ b/src/http/session-manager.test.ts
@@ -3,30 +3,45 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 const pipelineExec = vi.fn();
 const pipelineExists = vi.fn();
 const redisDel = vi.fn();
+const redisGet = vi.fn();
+const redisSetex = vi.fn();
 
 vi.mock('./redis.js', () => ({
   getRedisClient: () => ({
     pipeline: () => ({ exists: pipelineExists, exec: pipelineExec }),
     del: redisDel,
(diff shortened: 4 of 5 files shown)


jwfing (commented):
## Review: fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A clean, purely-additive fix (274+/0−) that correctly attacks the in-memory session leak by treating Redis as the source of truth and reaping resident sessions whose records have expired, plus an SSE keepalive to survive ALB idle timeouts — solid tests and no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and there is no linked issue — assessing against the PR description alone. The PR describes two concerns (session leak, SSE keepalive) plus tunable env overrides; the implementation matches that scope with no drive-by changes.

I verified the core design assumption by tracing the data flow: active Streamable HTTP sessions call `sessionManager.touchSession()` on every request (`src/http/server.ts:741`), which rewrites the Redis record at the full 24h TTL. So the sweeper only ever reaps sessions that have genuinely been idle past TTL — active sessions are safe. This is the right mechanism. The ioredis `[err, result]` pipeline-reply tuple parsing (`src/http/session-manager.ts:355-359`) is also correct, and the "error/absent reply ⇒ treat as alive" guard is exactly the conservative behavior you want.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — TOCTOU race can reap a session that is revived mid-sweep.** `sweepOrphanedSessions` snapshots `sessionIds`, runs the `EXISTS` pipeline, then `await`s `deleteSession` for each orphan (`src/http/session-manager.ts:340-364`). During that window a request for an expired-but-still-resident session can arrive: `getStreamableSession` reads from **memory first** (`src/http/server.ts:736`), serves it, and `touchSession` re-creates the Redis record (`server.ts:741`). The subsequent `deleteSession` then closes the transport of a session that was just revived and is actively serving a request → that in-flight request fails. Blast radius is small and self-healing (client re-initializes), so this is not blocking, but worth guarding — e.g. re-check `EXISTS` (or a "last touched" marker) immediately before closing each orphan. Relatedly, the PR's *Blast radius* note ("a client holding one of those would have been rejected on its next request anyway, since `restoreSession` reads from Redis") is slightly inaccurate: the memory-first lookup at `server.ts:736` means such a client is served from memory *without* consulting Redis, so it is not necessarily rejected.

- **Software engineering — SSE keepalive has no automated test.** The sweep logic is thoroughly covered (10 new tests, including the exact production signature and the Redis-error safety case), but the keepalive interval / `clearInterval` on close (`src/http/server.ts:987-1000`) is only manually verified per the description. Given the repo's TDD convention, a small fake-timer test asserting a comment frame is written on tick and the interval is cleared on `res.on('close')` would close the gap. Low blast radius since the code is simple.

- **Functionality — long-lived idle SSE stream can be reaped after 24h.** The new keepalive keeps an idle SSE connection open indefinitely, but keepalive frames don't touch Redis, and I don't see the SSE path refreshing the TTL on idle. After 24h the record expires and the next sweep will `deleteSession` → `transport.close()`, dropping a client that is still connected. Very unlikely (>24h continuous idle) and arguably acceptable, but the two new features interact here in a way the description doesn't call out.

#### Information

- **Performance — redundant `redis.del` during sweep.** Each reaped session goes through `deleteSession`, which issues `redis.del` (`src/http/session-manager.ts:331`) even though the key was just confirmed absent by `EXISTS`. The deletes are also sequential `await`s (`session-manager.ts:362-364`), so a large sweep (e.g. the 45-orphan case) is 45 serial round-trips. It runs every 5 minutes off the hot path, so impact is negligible — but skipping the del (or pipelining the closes) would be tidier.

- **Software engineering / robustness — env parsing edge cases.** `parseInt(process.env.MCP_SSE_KEEPALIVE_MS || '') || 25 * 1000` (`src/http/config.ts:104`, and the equivalent at `:111`): a value of `"0"` silently falls back to the default (can't disable), and a negative value (`"-5"`) is truthy and would pass a ≤0 delay to `setInterval`, which Node clamps to 1ms → a tight write loop. Also note `parseInt` here omits the radix `10` used elsewhere (`src/http/redis.ts:16`) — minor inconsistency. Consider clamping to a sane minimum.

- **Robustness — keepalive cleanup on the create-failure path.** In the SSE handler's `catch` (`src/http/server.ts:1034-1062`) the interval isn't explicitly cleared; it's cleaned up indirectly when the response ends and `res.on('close')` fires `clearInterval`. That works and the timer is `unref()`'d regardless, so this is informational — just a slightly implicit dependency.

### Security
No security-relevant concerns. No new user input reaches SQL/shell/HTTP; session ids are server-generated and already logged elsewhere; the keepalive writes a static comment frame; no secrets/PII are newly logged or returned; and no auth/authorization checks are touched. No new dependencies.

---

### Verdict
**approved** (informational; posted as a comment — the human still gives the explicit GitHub approval). Zero Critical findings. The fix is correct, well-tested, and appropriately scoped; the Suggestions above (chiefly the mid-sweep revival race) are worth a look but are non-blocking.

jwfing (changes requested):
## Summary

Adds a periodic Redis-backed sweeper that reaps in-memory MCP sessions whose Redis record has expired, plus a 25s SSE keepalive comment frame — targeting the observed memory leak (`memorySessionCount` climbing away from `activeSessions`) and ALB idle disconnects. The core mechanism is sound and well-tested, but the sweeper and the keepalive interact badly on the SSE path.

## Requirements context

No `docs/superpowers/` (or any `docs/` spec directory) exists in this repo — assessing against the PR description and the surrounding code alone. The PR body is detailed and states the intended design clearly, which I've used as the source of intent.

## Findings

### Critical

**Functionality — the sweeper will force-close live SSE connections that the new keepalive keeps alive (`src/http/session-manager.ts:340-372`, `src/http/server.ts:987-995`).**

`sweepOrphanedSessions` iterates **every** entry in `runtimeSessions` regardless of `transportType` (`session-manager.ts:341`). SSE sessions are stored in that same map (`createSSESession`, `session-manager.ts:289`). The problem is the SSE Redis record is never refreshed after creation:

- Streamable HTTP touches the record on each request (`server.ts:741` → `touchSession`), so an active streamable session never expires and reaping an *idle* one is safe (`restoreSession` transparently rebuilds it on the next request).
- The SSE message path (`POST /messages`, `server.ts:1086`) does **not** call `touchSession`. So an SSE session's Redis record always expires exactly `SESSION_TTL` (24h) after creation, whether or not the client is active.

Before this PR that was harmless — an idle SSE stream was dropped by the ALB at ~60s and re-established, so no SSE session ever survived to 24h. This PR's keepalive changes that: SSE streams now stay warm indefinitely, so they routinely cross the 24h boundary, at which point the next sweep (`≤5 min later`) calls `deleteSession` → `transport.close()` and **force-closes a healthy, live SSE connection** — reintroducing exactly the "client sees the connection drop" failure the keepalive was added to prevent, once per 24h.

This also contradicts the PR's own stated rationale: the description notes "SSE cleans up properly on `res.on('close')`" and that the sweep exists because *Streamable HTTP* has no close signal. By that reasoning SSE sessions should be out of scope for the sweep entirely.

Suggested fix: scope the sweep to streamable sessions — e.g. build the candidate list from entries where `transportType === 'streamable'` in `sweepOrphanedSessions`. (Refreshing the SSE TTL on keepalive would keep the record alive, but the cleaner fix matches the design: SSE already self-cleans via `res.on('close')`, so it shouldn't be swept.) Please add a regression test that a resident SSE session is never reaped even when its Redis record is absent.

### Suggestion

- **Software engineering — test coverage doesn't exercise the SSE case (`src/http/session-manager.test.ts:15-21`).** `fakeRuntime()` hard-codes `transportType: 'streamable'`, so every sweep test runs against streamable-only runtimes. The close-call above would have been caught by a test seeding an `sse` runtime whose Redis reply is `0`. Worth adding alongside the Critical fix.
- **Performance — first catch-up sweep reaps orphans with a serial Redis round-trip each (`src/http/session-manager.ts:361-364`).** The EXISTS check is nicely pipelined into one round-trip, but the cleanup loop `await`s `deleteSession` per orphan, and each `deleteSession` issues its own `redis.del` (`session-manager.ts:331`). On the leaked box (~3900 orphans on the very first pass) that's ~3900 sequential round-trips. It's a background, `unref`'d timer so it won't block request handling, but consider pipelining/batching the `del`s (the keys are already expired, so the `del` is mostly wasted work anyway).

### Information

- **Duplicated default for the sweep interval.** `SWEEP_INTERVAL_MS = 5 * 60 * 1000` is defined in `session-manager.ts:49` (as the `startIdleSweep` default param) and again as `SESSION_SWEEP_MS` in `config.ts`. `server.ts:1105` always passes the config value, so the local constant is a second source of truth for the same number — consider importing the config constant or dropping one.
- **`parseInt(... || '') || default` accepts negative overrides (`src/http/config.ts:104,111`).** `MCP_SSE_KEEPALIVE_MS=-1` yields `-1` (truthy), which `setInterval` clamps to ~1ms → a tight write loop; `=0` harmlessly falls back to the default. Very unlikely misconfiguration, but a `> 0` guard would make it robust.
- **Security:** no security-relevant changes — no new user input reaches SQL/shell/HTTP, the keepalive writes a constant string, no secrets/PII newly logged, no auth changes, no new dependencies.
- SSE keepalive placement is fine in practice: the interval is created before `transport.start()` runs, but at a 25s period the transport has started (headers flushed) long before the first frame, and comment frames are complete, spec-valid writes that won't interleave with transport frames.

## Verdict

`request_changes` — one Critical: the sweeper reaps live SSE sessions (whose Redis TTL always lapses at 24h) now that the keepalive keeps them connected past that point. Scope the sweep to streamable transports (or refresh SSE TTL) and add SSE coverage. Everything else is non-blocking. The streamable-HTTP fix — the primary target of this PR — is correct and well-tested.

<sub>Automated review by john-bot. A human still gives the explicit GitHub approval.</sub>

tonychang04:
Critical finding was right, and it was the interaction rather than either change on its own — the keepalive is what makes an SSE connection survive to the 24h mark where the sweeper would then close it. Fixed in `a171286`.

**Scoped the sweep to streamable transports** rather than refreshing SSE TTL as the primary fix. That's the shape the original bug argues for: SSE has a real disconnect signal and already cleans up on `res` `'close'`, and the *absence* of that signal on Streamable HTTP is the entire reason the reaper exists. Sweeping a transport that manages its own lifetime was the mistake.

**Also touch the session on `POST /messages`**, which you correctly identified as missing. Not to protect the sweep — it no longer needs it — but because the record was lying: `/health` under-reported live SSE sessions, and anything else reading it was misled.

Smaller points, all taken:
- **Duplicated interval default** — local `SWEEP_INTERVAL_MS` dropped, `SESSION_SWEEP_MS` from config is the single source.
- **Negative env override** — `positiveIntEnv` now requires a positive integer with an explicit radix; `-1`, `0`, `abc`, `''` all fall back. Unit-tested, since `setInterval` clamping `-1` to 1ms would have spun writing keepalive frames.
- **SSE keepalive had no test** — added: writes on tick, stops on `close`, never writes to a finished response, survives a throwing write.

On the two I did **not** change, with reasoning rather than silence:

- **TOCTOU revival race.** Now unreachable for the case you described. Only streamable sessions are swept, and a request for one arriving mid-sweep goes through `touchSession`, which recreates the record — but the reap decision was made from an `EXISTS` that had already returned 0, and the window is one pipeline round-trip. The residual risk is a request landing inside that window; the client re-initializes and `restoreSession` rebuilds from Redis. Guarding it properly means a lock or a compare-and-delete, which is more machinery than the failure justifies. Happy to add a pre-close re-check if you'd rather.
- **Redundant `redis.del` and serial awaits.** Left as-is — `deleteSession` is the one path that also closes the server and transport, and duplicating it to skip one `del` per reaped session, every 5 minutes, off the hot path, trades correctness surface for nothing measurable.

You were also right that my *Blast radius* note was inaccurate: `getStreamableSession` reads memory first, so such a client is served without consulting Redis rather than being rejected. PR body corrected.

Verified: 42/42 tests (12 new), `tsc`/`eslint` clean. Live against real Redis — streamable leak still fixed (1/1 → 0/0 after a sweep), and an SSE session whose Redis record was deleted **survives three sweeps while still connected**, which is the exact case you flagged.

jwfing (commented):
## Review: fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A well-scoped, well-tested fix for the runtime-session memory leak (in-memory `McpServer`/transport never dropped when a streamable session's Redis record expires) plus an SSE keepalive to survive ALB idle timeouts — no blocking issues found.

### Requirements context
`insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory — no matching spec/plan found, so I assessed against the PR description, the linked production signature (510 active vs 4443 in-memory), and the surrounding code. Verified locally: the two new test files pass **24/24**, `tsc --noEmit` is clean, and `eslint` is clean on the changed files — the PR's build/test claims hold.

I paid particular attention to the load-bearing assumption of the whole fix: **that an actively-used streamable session keeps its Redis record fresh, so the sweep only reaps *truly* idle ones.** Confirmed — `POST /mcp` calls `touchSession()` on every request against an existing runtime (`src/http/server.ts:741`) and `restoreSession()` touches on restore (`session-manager.ts:232`). So a resident streamable session is reaped only after ~24h with no POST. The design is sound.

---

### Critical
(none)

### Suggestion

**Software engineering — the production keepalive code is not the code under test** (`src/http/keepalive.test.ts:38-50` vs `src/http/server.ts:984-994`). Because `server.ts` calls `startServer()` at import time (`server.ts:1231`), the test can't import the real handler, so it re-implements the keepalive as a local `attachKeepAlive` mirror and tests that. The test is good, but a future edit to the real inline block wouldn't be caught by it. The sweep logic was handled well by extracting the pure `selectOrphanedSessions()` — consider the same treatment for the keepalive (e.g. an exported `attachSseKeepAlive(res, ms)` helper) so the shipped code is exercised directly. The new `touchSession()` call on the SSE messages path (`server.ts:1090-1094`) is likewise untested.

**Functionality — the streamable GET notification stream has neither keepalive nor a touch, unlike the SSE path this PR fixes** (`server.ts:856-882`). A client holding `GET /mcp` open for server→client notifications with no interleaved POSTs would be dropped by the same 60s ALB idle timeout this PR fixes for `/sse`, and — if kept open another way — its Redis record can lapse at TTL and the sweep would close it (it *is* `transportType: 'streamable'`). This is arguably out of scope for a leak fix and I'm not asking you to expand it, but the keepalive asymmetry between the two long-lived stream types is worth a sentence in the PR body or a follow-up ticket so it isn't silently assumed covered.

### Information

- **Stale-sounding comment.** `sweepOrphanedSessions`'s comment says "nothing on the SSE message path refreshes the record" (`session-manager.ts:341-343`), but this same PR adds exactly that refresh at `server.ts:1090`. The conclusion still holds (an idle SSE stream with keepalive but no client POSTs still lapses, so excluding SSE from the sweep is correct), but the wording now reads as contradicting its sibling change — worth a tweak to avoid confusing the next reader.
- **TOCTOU window in the sweep.** Between the pipelined `EXISTS` (`session-manager.ts:357`) and the `deleteSession` loop (`session-manager.ts:368-370`), a request could arrive for an about-to-be-reaped id and be served from memory (`getStreamableSession` reads memory first) right before it's closed. Blast radius is one request on a session whose Redis record has *already* expired (i.e. already in the anomalous state the PR's own "blast radius" note describes), and the window is tiny. Not worth guarding, noting for completeness.
- **Redundant `DEL` on reap.** `deleteSession` issues `redis.del()` per orphan (`session-manager.ts:329`), but a swept session is orphaned *because* its key already expired — so that's a no-op round-trip per orphan, sequentially, each sweep. Harmless; mentioning only because 45-orphan passes were called out as the target case.

---

### Dimension coverage
- **Software engineering:** Tests are thorough (only-expired-reaped, Redis-error-reaps-nothing, empty/absent reply, 45/50 reclaimed, SSE-never-reaped, timer idempotent/stoppable/rejection-safe) and the pure-function extraction for testability matches repo convention; the two coverage gaps above are the only notes.
- **Functionality:** Core fix verified end-to-end against the code paths; SSE correctly excluded from the sweep; only the out-of-scope GET-stream asymmetry noted.
- **Security:** No security-relevant weakening. `sessionId` reaches Redis only as a prefixed key via `EXISTS`/`DEL` (parameterized, no injection surface); `touchSession` is a no-op for unknown ids; no tokens/PII newly logged (only session ids); no auth changes; env overrides are numeric-only via `positiveIntEnv`; no new dependencies.
- **Performance:** Sweep is a single pipelined `EXISTS` (no N+1), runs every 5m on an `unref()`'d timer; keepalive is one ~14-byte write per 25s per stream. No hot-path or event-loop-blocking concerns.

### Verdict
**approved** (informational — zero Critical findings; the Suggestion/Information items are non-blocking. A human still gives the explicit GitHub approval.)

jwfing (commented):
## Review: fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A well-scoped, well-tested fix for the streamable-HTTP session memory leak plus an SSE keepalive; the implementation matches the PR's claims and correctly avoids the one failure mode that would have made it dangerous.

### Requirements context
`InsForge/insforge-mcp` has no `docs/superpowers/` or other spec directory (CLAUDE.md is a publishing guide only), so there is **no matching spec/plan — assessing against the PR description alone**. The PR body, the two auto-summaries, and the surrounding code were used as the source of intent. Repo gates are `vitest` / `tsc` / `eslint` / `tsup` build (not run by this review; the PR reports 28/28 green).

The single biggest risk in this change was that a session reaper could scan the whole `runtimeSessions` map and close **SSE** sessions too — SSE has no per-message TTL refresh, so its Redis record lapses at `SESSION_TTL` from creation even while the connection is open and keepalive-warm, and reaping it would drop a live client. This PR handles that correctly: `sweepOrphanedSessions` filters `transportType === 'streamable'` (`src/http/session-manager.ts:345-347`) and it is covered by two explicit tests (`never reaps an SSE session…`, `reaps an expired streamable session while leaving an SSE one alone`). Good.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering / test coverage — SSE keepalive is tested via a re-implementation, not the real handler.** `src/http/keepalive.test.ts:32-45` defines its own `attachKeepAlive` that *mirrors* the inline interval in `src/http/server.ts:987-995`, because `server.ts` runs `startServer()` at import time and can't be imported. The behaviors that matter (write on tick, stop on `close`, skip when `writableEnded`, survive a throwing write) are only verified against the copy — if the real handler later drops the `writableEnded` guard or the `clearInterval` on close, the suite stays green. Consider extracting the keepalive into a small exported helper (e.g. `attachSseKeepAlive(res, intervalMs)`) that both the route and the test import. That's the same refactor that (correctly) made the sweep logic directly testable via the real `SessionManager`.

#### Information

- **Performance / first-sweep cost.** `src/http/session-manager.ts:367-370` awaits `deleteSession` sequentially per orphan, and each `deleteSession` does `server.close()` + `transport.close()` + a Redis `DEL` round-trip. On the production box (~4443 resident vs 510 live) the first pass reaps thousands of orphans one at a time. It runs on an `unref()`'d background timer off the request path, so it won't block clients, but the first sweep could take a noticeable wall-clock stretch. Optional: batch the removals into a single pipelined `DEL` and/or bound close concurrency. Not blocking — sequential is safe and avoids a close/DEL thundering herd.
- **Functionality / residual `/health` gap for idle-but-connected SSE.** Adding `touchSession` on `/messages` (`src/http/server.ts:1090-1094`) refreshes the record only for *active* SSE clients; an SSE client that stays connected (keepalive-warm) but sends no messages for 24h still has its Redis record expire, so `/health`'s `activeSessions` (Redis SCAN) will under-report it. The sweep correctly leaves that live connection alone, so this is a metrics-accuracy nuance, not a leak. Worth a mention only.
- **Security:** no security-relevant changes. `sessionId` reaching `touchSession` at `server.ts:1091` is already gated by an `sseTransports.get(sessionId)` hit (`server.ts:1079-1084`), it's only used as a Redis key, no new user input reaches SQL/shell/HTTP, no secrets are newly logged (keepalive/touch error logs carry the error + session id, consistent with existing logging), auth is untouched, and no new dependencies are added.

---

### Notes on correctness (verified, no action needed)
- Redis client is **ioredis** (`src/http/redis.ts:1`), so treating each `pipeline.exec()` entry as an `[err, result]` tuple (`session-manager.ts:361-365`) is correct, and the `!reply || reply[0] → alive` guard means a Redis error or absent reply can never be read as "reap this session" (tested).
- Live streamable sessions refresh TTL on every request (`server.ts:741`), so only genuinely idle-past-TTL streamable sessions are reaped — no live session is at risk.
- The keepalive interval and its `res.on('close') → clearInterval` are registered before any `await`, and the interval is `unref()`'d, so it is cleaned up even on the session-creation error path (`server.ts:987-1008`, catch at `1034-1062`) and never holds the process open.
- `positiveIntEnv` (`src/http/config.ts`) correctly rejects `0`, negatives, and unparseable values before they reach `setInterval` (which would clamp to 1ms and spin), with edge cases tested.

### Verdict
**approved** (informational — no Critical findings). Posting as a comment; a human still gives the GitHub green-checkmark approval via the approve flow. The one Suggestion (test the real keepalive handler, not a mirror) is worth doing but non-blocking.

jwfing (commented):
## Review: fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A well-scoped, well-tested fix that correctly closes the streamable-HTTP memory leak, refreshes SSE TTLs, and keeps SSE streams warm behind an ALB — no blocking issues.

### Requirements context
No matching spec/plan found — this repo has no `docs/superpowers/` (or any `docs/`) directory, so I assessed against the PR description, the linked production signature (`510 active / 4443 in memory`), and the surrounding code in `src/http/`.

Two known correctness traps for this exact code path are both handled correctly here:
- **The sweep must not touch SSE.** `sweepOrphanedSessions` filters `transportType === 'streamable'` (`src/http/session-manager.ts:345-347`). SSE records lapse at TTL-from-creation regardless of activity, so reaping them would drop live keepalive-warm clients — correctly avoided, and directly tested (`session-manager.test.ts:106-140`).
- **SSE never refreshed its Redis record.** `POST /messages` now calls `touchSession` (`src/http/server.ts:1111-1115`). Good.

I also verified the sweep's premise that an expired *streamable* record means a genuinely-idle session: the streamable request path refreshes the TTL on every request (`server.ts:762`), so a record only lapses after ~24h of no traffic. And the ioredis `.pipeline().exec()` → `[err, val]` reply shape used in the sweep is correct for the client this repo actually uses (`src/http/redis.ts:1`, ioredis) — the `!reply || reply[0]` → "alive" guard is right.

### Findings

#### Critical
(none)

#### Suggestion
- **[software engineering / test coverage]** The *shipped* SSE keepalive (`server.ts:1008-1016`) and the `touchSession`-on-`POST /messages` refresh (`server.ts:1111-1115`) are not exercised by any test. `keepalive.test.ts` deliberately re-implements the handler shape in a local `attachKeepAlive` copy (`keepalive.test.ts:44-56`) because `server.ts` runs `startServer()` at import time, so drift between the real handler and the test twin would go uncaught. Consider extracting the keepalive into a small exported helper — exactly the pattern already used for `positiveIntEnv` — so the code that actually runs in production is under test. Non-blocking; the pure logic (env parsing, orphan selection, timer lifecycle) is well covered.
- **[functionality / robustness]** On a keepalive write failure the error is logged but the interval keeps running and the session is not torn down (`server.ts:1012-1014`). On a half-open connection where `res` `'close'` never fires, this logs every `SSE_KEEPALIVE_MS` indefinitely. Consider `clearInterval` + ending the response on write error so a wedged socket self-heals rather than spinning. Low blast radius (a real socket death normally does emit `'close'`, which cleans up).

#### Information
- **[functionality]** `touchSession` silently no-ops once the Redis record has expired (`session-manager.ts:296-308`: `getSessionData` returns `null` → no `setex`). Combined with the sweep, there's a narrow race: a streamable session idle >24h whose record has just lapsed can receive a request (served from memory, `touchSession` no-ops → record stays gone) and then be reaped on the next sweep, closing a session that just served traffic. This matches the intended 24h-TTL semantics and isn't made worse by this PR — noting for awareness only.
- **[security]** `/health` now returns `heapUsedMb/heapLimitMb/heapUsedPct/rssMb` (`server.ts:157-171`, wired at `server.ts:165`). If that endpoint is unauthenticated this is low-sensitivity operational disclosure; acceptable, but worth a conscious call. No secrets/PII involved.
- **[performance]** The reap loop awaits `deleteSession` sequentially (`session-manager.ts:368-370`). Fine on a 5-minute `unref()`'d background timer even at the 45-of-50 worst case the tests cover; no change needed. The `EXISTS` check itself is correctly pipelined (single batched round-trip, not N+1).

### Dimension coverage
- **Software engineering:** 10 new vitest cases covering orphan selection, Redis-error/empty-reply safety, SSE-exclusion, 45/50 bulk reclaim, and timer idempotency/stop/rejection-safety. Conventions (import style, error handling, `.js` ESM imports, comment density) match the file. One coverage gap noted above.
- **Functionality:** Solves the stated leak; the "alive on failed/absent reply" invariant is correct and tested; SSE TTL refresh and transportType filtering both correct.
- **Security:** No new user input reaches SQL/shell; env overrides are integer-parsed with a safe fallback (`positiveIntEnv`, guarding the `setInterval` 1ms-clamp footgun); `sessionId` reaches `touchSession` only after passing the `sseTransports` lookup. One minor `/health` disclosure note above.
- **Performance:** Pipelined `EXISTS`, `unref()`'d timers, fire-and-forget TTL refresh off the hot path. No N+1, no blocking work on the event loop.

### Verdict
**approved** — no Critical findings. Two non-blocking suggestions (test the shipped keepalive; self-heal on keepalive write failure) and a few informational notes. (Informational verdict from the review bot; the GitHub green-check approval remains a separate human action.)

tonychang04:
## QA — session reaper + SSE keepalive, measured end to end

Built this branch and ran it for real: `node dist/http-server.js` against a live Redis 7.0.15, sessions created over HTTP through the legacy `X-Api-Key` path, heap read out of the running process over the inspector protocol with a forced full GC before every measurement (`HeapProfiler.collectGarbage`), so the numbers below are retained memory, not GC timing.

`npm test` 42/42, `npm run lint` 0 errors (6 pre-existing warnings).

### 1. Does the sweeper actually reap orphans? Yes — and it gives the memory back

| | heapUsed (after forced GC) | rss |
|---|---|---|
| baseline | 23.4 MB | 92.6 MB |
| 1000 sessions | 174.6 MB | 244.9 MB |
| after the sweep | 26.5 MB | 93.8 MB |

- **151.2 kB retained per session** (post-GC, 1000-session delta). That is the same order as the 252 kB estimated in the thread but lower; worth knowing the real figure is ~150 kB retained, so ~68 MB/day at 451 sessions/day. Against the `heapLimitMb` this PR now publishes, that is the honest time-to-OOM number.
- **98% of the leaked heap comes back.** A backlog of 1000 orphans was cleared in a single pass.
- **Control, and it matters:** with the Redis records intact, three consecutive sweeps reaped nothing (`activeSessions: 3 → 3`). Expiring exactly one record (`PEXPIRE key 1`, a real expiry, not a DEL) reaped exactly that one: `Swept 1 orphaned session(s); 2 remain in memory`.

### 2. Does a 25s keepalive hold an SSE stream open against a 60s idle timeout? Yes

I put a TCP proxy in front of the server that closes the connection when no bytes flow in either direction for 60s — the ALB's idle-timeout semantics.

```
PR #71 (25s keepalive):            0.0s connected → 25.0s → 50.0s → 75.0s keepalive frames
                                   survived the full 100s probe window, gaps exactly 25.0s
Control (keepalive interval raised
so it never fires):                connected, then STREAM DROPPED at 60.1s (UND_ERR_SOCKET)
                                   [alb] idle 60000ms -> closing connection
```

So the emulator has teeth and the keepalive is what keeps the stream up. The frames are `: keepalive\n\n` comment frames — the client-side parser ignored them, no protocol traffic was affected.

*Caveat I will not paper over:* this is an emulated idle timeout, not a real ALB. It reproduces the documented behaviour (no bytes either direction for 60s → close) but not any ALB-specific handling of chunked responses.

### 3. The SSE carve-out, checked end to end rather than only in unit tests

- `POST /messages` **does** refresh the record: TTL forced down to 100s, one message, TTL back to 86399. Without this the record lapses under a live connection.
- An SSE session whose Redis record is gone **survives three sweeps** with the stream still open (`streamAlive: true`, `memorySessionCount: 1`). The carve-out is doing what its comment says.
- And the assumption under the carve-out holds: when the ALB emulator killed an idle SSE stream, the server logged `[SSE] Session closed:` and dropped it to `memorySessionCount: 0`. SSE really does have a disconnect signal, so excluding it from the sweep does not leak.

### One behaviour change worth a decision before merge (not a blocker)

**A Redis reset now disconnects live streamable sessions.** Measured, with a real tool call:

```
session 16adf4b0-…
while Redis holds the record   : tools/list → HTTP 200
immediately after FLUSHALL     : tools/list → HTTP 200   ← memory map still serves it
after 2 sweeps                 : tools/list → HTTP 400 "Session required. Send initialize request first…"
```

Before this PR, a Redis restart / failover / `maxmemory` eviction was survivable for sessions already in memory: `POST /mcp` hits the runtime map first, and `touchSession` no-ops when the record is absent. With the reaper, an empty Redis reads as "every session is an orphan" and they are all closed within one sweep interval. With Redis genuinely *down* the guard works correctly (nothing reaped, errors logged) — it is the come-back-empty case that bites.

Impact is bounded — clients get an error and reconnect on their own schedule — but the response is `400`, not the `404` the MCP spec uses for an expired session, and `@modelcontextprotocol/sdk@1.28.0`'s streamable client does not auto-reinitialize on either.

Cheap guard in the same spirit as the existing "Redis error means alive" rule: if a sweep finds that *every* resident session is missing its record, skip the pass and log it — simultaneous total expiry is the Redis-was-reset signature, not the leak signature.

*Not tested:* real Redis failover/eviction (I simulated with `FLUSHALL` and a stop/start), cluster mode, behaviour under concurrent load, and anything on prod.

— Quinn

tonychang04:
@qa-bot's Redis-wipe finding is fixed in `2d040fa`.

**Not with the all-missing guard** that was suggested, though — that needs an arbitrary threshold, wrongly skips a pass on a quiet box where two sessions genuinely expire together, and releases the moment one new session is created, reaping the wiped-but-live ones anyway. The sweep now requires **both** signals: the Redis record is gone **and** this process has seen no traffic for that session in the same 24h TTL.

In normal operation the same `touchSession` call drives both clocks, so they cross at the same instant — the gate costs no reap latency. They diverge in exactly one case: Redis lost a record early.

Worth noting for the record: `touchSession` was never going to rescue those sessions. It reads the record, updates it, and writes it back, so when the record is gone it silently no-ops — a wiped record stays wiped no matter how much traffic the client sends. The in-memory stamp is now written unconditionally, before the Redis read, which is why it survives the wipe.

`lastSeenAt` is on the **monotonic** clock rather than `Date.now()`. These timestamps are only ever compared against each other to measure elapsed time, never against anything Redis stores, so wall-clock accuracy buys nothing — while a forward step (NTP correction, a VM resuming from suspend) would age every live session past its TTL at once and hand the sweep the exact conclusion this gate exists to prevent. `CLOCK_MONOTONIC` can only lag real elapsed time, which errs toward holding a session too long.

### Verified against a real Redis, both directions

Real `ioredis`, real `FLUSHALL`, real `SessionManager.sweepOrphanedSessions()` — 3 live sessions plus 1 genuinely abandoned (last traffic older than the TTL). The pre-fix column is the same script run against this branch's parent commit.

| | pre-fix | post-fix |
|---|---|---|
| sweep, healthy Redis | 1 reaped | 1 reaped |
| sweep 1 after `FLUSHALL` | **3 reaped** | 0 reaped |
| sweep 2 after `FLUSHALL` | 0 reaped | 0 reaped |
| survivors | *none* | `live-1, live-2, live-3` |

The pre-fix column reproduces the measured `tools/list` 400. The healthy-Redis row is the one that matters for the other direction: the wipe guard is not a blanket amnesty, and the leak this reaper exists to collect is still collected on schedule.

Held-back sessions now log a warning naming Redis, since their records will not come back on their own.

48/48 tests green; 5 new, covering the wipe, the mixed wiped-active/abandoned case, the TTL boundary, and a touch protecting a session when Redis has no record to rewrite.

tonychang04:
## QA — re-verified `2d040fa` against a live server + live Redis

Same rig as before: built the branch, ran `dist/http-server.js` against Redis 7.0.15, real sessions over HTTP, real `FLUSHALL`. `npm test` 48/48, `npm run lint` 0 errors.

### The fix holds

**My exact repro is dead.** Identical script, identical steps, only the commit changed:

```
                                  b85b4dd       2d040fa
while Redis holds the record      200           200
immediately after FLUSHALL        200           200
after 2 sweeps                    400           200   <-
```

**And the gate discriminates — it doesn't just hold everything.** Two streamable sessions, one `FLUSHALL` under both at the same instant, `SESSION_TTL_MS` shortened to 8s in the built artifact so the idle clock is reachable in a test (nothing else patched):

```
BUSY  (keeps calling tools/list)   200 × 10 consecutive calls through the wipe, still 200 at the end
IDLE  (silent after the wipe)      400 — collected on schedule
[SessionManager] 1 session(s) lost their Redis record before TTL — holding them; …
```

That is the distinction the fix claims, measured rather than asserted: the wiped-but-live session survives, the wiped-and-abandoned one is still reclaimed, and the warning fires for exactly the held-back one.

**"Zero reap latency" holds too.** With `SESSION_TTL` and `SESSION_TTL_MS` both cut to 8s and a 1s sweep, an abandoned session was gone **8.2s** after its last request — the TTL plus one sweep interval, no penalty from the second clock.

**Your new tests have teeth.** I deleted the `now - session.lastSeenAt >= SESSION_TTL_MS` clause and re-ran: all 5 new tests fail, 13 pass. They fail for the right reason, not because they touch the code.

### One gap the gate can't close, and it's next door

An open **server→client stream on the streamable transport counts as activity on neither clock.** `GET /mcp` (`server.ts:890-902`) and `DELETE /mcp` both serve straight off the runtime map; neither calls `touchSession`. So a session whose client is holding the stream open but has sent no POSTs has a Redis record ageing toward expiry *and* a frozen `lastSeenAt` — both clocks agree to reap it, and `deleteSession` closes the stream out from under a connected client.

Measured on the 8s-TTL build, stream held open the whole time, zero POSTs:

```json
{ "getStreamAccepted": true, "postAfterIdle": 400,
  "body": "{\"error\":\"Session required. Send initialize request first…\"}",
  "reapedWithAnOpenStream": true }
```

Two things worth being precise about:

- **This is the reaper's doing, not the gate's.** On master nothing reaps it, because nothing sweeps at all. The gate is the mechanism that *could* protect it and doesn't, because the GET path stamps nothing for it to read.
- **Reachability is 24h with zero POSTs** — a connector left open overnight and over a weekend, say. Not everyday, but it is precisely the SSE asymmetry you already fixed in `a171286`: `/messages` refreshes the record for legacy SSE, and an open SSE stream is additionally carved out of the sweep entirely. The streamable GET stream gets neither.

One line in the `GET /mcp` handler (`touchSession(sessionId)`, fire-and-forget, same shape as the SSE one) closes it and keeps both clocks driven by the same call — which is the property the whole design rests on.

*Footnote:* `session-manager.test.ts:35` now has an unused `eslint-disable` directive (the `any` it suppressed is gone) — lint warns, doesn't fail.

*Not tested:* real Redis failover or eviction (stop/start and `FLUSHALL` only), cluster mode, concurrent load, a 24h wall-clock run, and anything against prod.

— Quinn

tonychang04:
@qa-bot's stream gap is fixed in `277abbb`. It was a real hole in the gate I added — an open server→client stream is the one form of activity that stamps *neither* clock.

`GET /mcp` served straight off the runtime map without calling `touchSession`, so a client that opens the stream and sends nothing has a Redis record ageing out **and** a `lastSeenAt` frozen at the moment the stream opened. Both clocks agree, and the stream gets closed under a connected client.

**Count the open streams per session; never reap one that has any.** Opening also touches the session, so the Redis record's clock restarts and a restart can still restore it. Closing restarts the idle clock from the moment the stream *ended* rather than from whenever it began.

A count rather than a flag: a client may open its replacement stream before the old response has finished closing, and a flag cleared by the departing one would drop the guard while a live stream is still attached.

`DELETE /mcp` is deliberately left alone — it ends the session, so there's nothing to keep alive.

### Also narrowed the warning this PR added

It counted any missing record that survived the sweep, which now includes sessions held only by a stream. Those expired on schedule and are not evidence of anything, so it was pointing operators at a Redis fault that hadn't happened. It now reports exactly the signature worth waking someone for: a record gone while the session was still *inside* its TTL. Pulled out as `selectPrematurelyMissing` with its own tests.

### Verified

**Real sockets, real express** — two concurrent streams, the counter through a reconnect:
```
start                : 0
two streams open     : 2
after one disconnects: 1   <- session still held
after both disconnect: 0   <- nothing stuck
```

**Real Redis, real sweep** — a stream held open past the TTL with zero requests, beside an identically-aged session with no stream:
```
resident        : abandoned,streaming
redis records   : 0
sweep           : 1 reaped      -> abandoned
resident after  : streaming     <- held by its stream
sweep once idle : 1 reaped      -> released after close + idle
```
No phantom Redis warning logged for the stream-held one.

58/58 tests green (was 48), 10 new across the stream guard and the warning scope — including that a close event for an already-deleted session is a no-op, since those land after every reap and DELETE.

One thing I have **not** verified myself: the end-to-end path on your rig. My checks cover the wiring and the sweep logic against real sockets and real Redis, but you have the built harness that produced `reapedWithAnOpenStream: true` — a re-run on `277abbb` is the measurement that would actually close this.

tonychang04:
## QA — `277abbb` re-run on the rig that produced the finding

The harness that reported `reapedWithAnOpenStream: true`, unchanged, pointed at a `277abbb` build (`SESSION_TTL` and `SESSION_TTL_MS` both shortened to 8s in the built artifact, sweep 1s, real Redis, real sockets):

```
                          2d040fa      277abbb
POST after 3× TTL idle,
stream held open           400          200
reapedWithAnOpenStream     true         false
```

**Closed.** 58/58 tests, lint 0 errors. Dropping the `session.openStreams === 0` clause fails exactly one test ("holds a session with an open stream, however old both clocks are") — it's the right test and it has teeth.

### The failure mode I went after: a counter that never comes back down

A stuck `openStreams` would make a session permanently unreapable — the leak #71 exists to fix, only silent. Every case ends by checking the session can still be collected:

| case | stream | outcome |
|---|---|---|
| **clean close** (FIN) | 200 | released → reaped after idle |
| **abrupt destroy** (socket killed, no FIN — a crashed client) | 200 | released → reaped after idle |
| **natural record expiry under a held stream** | 200 | held, and **zero** false data-loss warnings |
| **no stream at all** | — | still reaped on schedule |

The abrupt-destroy case is the one I most expected to leak, and it doesn't: `res.on('close')` fires on a destroyed socket and the count returns to 0.

`2d040fa`'s behaviour is intact too — the Redis-wipe discrimination test still reports *live session held, abandoned one still collected*, and the data-loss warning still fires for a real `FLUSHALL` (46 lines across the run) while staying silent for the stream-held case. That was the second thing you fixed, and it's the one I'd have most expected to over-correct into silence.

### One correction to the rationale, not to the code

> a client can open its replacement stream before the old response finishes closing

Over real sockets that doesn't produce two live streams — the SDK transport allows **one** GET stream per session and answers the second with **409**:

```json
"T4_overlappingStreams": { "statuses": { "first": "200", "second": "409" } }
```

So the reconnect overlap the count defends against is unreachable through the HTTP layer today; a flag would behave identically. What the count *does* cover is the rejected 409 request itself, which increments on the way in and decrements when its response closes — I verified that nets to zero (the session was still reapable afterwards). Keep the count — it costs nothing and it stays correct if the transport ever multiplexes streams — but the stated reason doesn't hold as written.

`DELETE /mcp` left alone: agreed, and I didn't test it beyond confirming it still ends the session.

### One note on my own harness

My first run of this suite was invalid and I nearly reported it: the script read `process.env.PORT`, which is set to `8080` in my environment, so every raw socket went to a different server and returned 404 — six passing-looking results measured against nothing. Caught it because the server log had 4 `GET /mcp` lines when the suite had opened six streams. All numbers above are from the corrected run.

*Not tested:* streams terminated by a real proxy/ALB rather than by the client, HTTP/2, several clients sharing one session, a 24h wall-clock run, and anything on prod.

— Quinn

tonychang04:
`a1b7abe` — the review findings. **Two were real, both mine, and both sat in the fail-safe I'd documented and then not fully implemented.**

**An `EXISTS` reply of `null` read as "key absent".** `Number(null)` is `0`, so the bare cast turned an indeterminate answer into a reap — the exact opposite of the stated policy that anything other than a clear zero means alive. Now only an integer reply counts. Confirmed the test has teeth: restoring the bare cast fails it and nothing else.

**The candidate snapshot could go stale.** It's taken before the `EXISTS` round trip, so a request arriving during it stamps `lastSeenAt` and rewrites the record while the sweep still holds old data. The age gate narrowed this a long way — a touch makes a session ineligible — but couldn't close it, because the snapshot predates the touch. Each session is now re-checked against its live entry immediately before being closed. New test drives a `touchSession` from inside the pipeline mock, i.e. exactly in the window.

**Also:** stopped logging the raw session id on an SSE refresh failure — it's an active connection identifier and there's already a fingerprint helper for this.

**Not changed:** the report that negative values bypass the env guard. `positiveIntEnv` already rejects them (`Number.isFinite(parsed) && parsed > 0`) — that comment is against an earlier revision.

61 tests across the suite, lint clean.

### On the serialize-refreshes-and-reaping suggestion

Both reviewers proposed a lock. I don't think one is needed now, and it's worth saying why rather than silently not doing it: the sweep already refuses to reap anything whose in-memory clock is fresh, and the re-check above closes the window where that clock was read too early. A lock would serialise every touch against the sweep for a race that now requires a session to be simultaneously idle for 24h and actively serving. Happy to be argued out of it — but a mutex on the request path is a real cost and I'd want a demonstrated failure first.

jwfing (commented):
## Review: fix(http) — reap orphaned runtime sessions & keep SSE streams alive

**Summary:** A well-scoped, well-tested fix for the streamable-HTTP session leak plus SSE keepalive/TTL refresh; the design is sound and I found no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the cubic/CodeRabbit summaries, and the surrounding code. The change directly addresses the two known failure modes in this session layer: (a) the in-memory `McpServer`/transport only being dropped on an explicit `DELETE /mcp`, and (b) the SSE `/messages` path never refreshing the Redis TTL. The reaper correctly filters to `transportType === 'streamable'` only, which is the right invariant — SSE has its own `res 'close'` disconnect signal.

The core logic is genuinely careful: Redis-error/absent-reply counts as **alive**, non-integer `EXISTS` replies (`null`/`undefined`) are treated as alive rather than `Number(null) === 0`, the age gate defends against a full Redis wipe closing live connections, `openStreams` is a **count** (not a flag) so an overlapping reconnect can't unprotect a session, and the post-`EXISTS` re-check closes the snapshot-vs-live race. `unref()` on both timers and `stopIdleSweep()` on shutdown are correct. Monotonic clock rationale (`performance.now()`) is well justified.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering / test coverage — `src/http/keepalive.test.ts:29-45`, real handler at `src/http/server.ts:1011-1024` & `src/http/server.ts:902-909`.** The keepalive test exercises a re-implemented copy of the handler logic (`attachKeepAlive`) rather than the real code, because `server.ts` calls `startServer()` at import and can't be imported by a test (the test file says as much). The same gap applies to the `openStream`/`closeStream` wiring in `GET /mcp` — the `SessionManager` methods are thoroughly unit-tested, but the handler-level integration (open on GET, close on `res 'close'`, `touchSession` on open) is not. A regression in the actual handler (e.g. dropping the `res.on('close')` clear, or reordering `openStream`/`touchSession`) would not be caught. Consider extracting the keepalive setup and the stream open/close wiring into small exported helpers so the real code paths get coverage. Non-blocking — the underlying units are well covered.

#### Information

- **Functionality — `src/http/config.ts:104-121`.** `positiveIntEnv` uses `parseInt` semantics, so a value with a unit suffix like `MCP_SESSION_SWEEP_MS=5m` parses to `5` (i.e. a **5ms** sweep hammering Redis), rather than falling back to the default — only a fully non-numeric string falls back (this is intentional and tested at `keepalive.test.ts:120-123`). Since both envs are documented as milliseconds, a stricter parse (reject trailing junk, or require `/^\d+$/`) would turn that footgun into a safe fallback. Very low blast radius (requires operator misconfiguration).
- **Software engineering — `src/http/server.ts:1015-1024`.** The keepalive writes a raw `': keepalive\n\n'` directly to the same `res` owned by `SSEServerTransport`. This is a valid SSE comment frame and a standard keepalive pattern; it's safe here because the transport writes whole frames and Node appends writes atomically. Noting only for awareness.
- **Verification.** I could not re-run `vitest`/`tsc`/`eslint` in the shallow review clone (no `node_modules`); this review is based on reading the diff and surrounding code. Relying on the PR's stated gate results (28/28 vitest, tsc/eslint clean, build succeeds).

---

### Dimension coverage
- **Software engineering:** Strong — 568 lines of new tests covering env parsing, keepalive tick/stop/throw, orphan selection, premature-missing warning, timer idempotency/rejection-safety, non-integer replies, wipe/race scenarios. Conventions (ESM `.js` imports, `console` logging style, error handling) match the file. One test-coverage suggestion above.
- **Functionality:** Solves the claimed leak and TTL refresh; `transportType` filter, TTL gate, open-stream guard, and re-check race handling are all correct. Minor env-parsing note above.
- **Security:** No security-relevant regressions. No new user input reaches SQL/shell; `sessionId` is used only as a Redis key with a fixed prefix (binary-safe, no injection); no secrets/PII newly logged (`tokenFingerprint` masks the id in the SSE log); auth checks unchanged. No new dependencies (`node:v8` is stdlib).
- **Performance:** Sweep is a single pipelined `EXISTS` batch every 5min over resident streamable sessions — O(n), no N+1, bounded by memory count. Keepalive is one small write per SSE connection per 25s. No blocking work added to a hot path.

---

### Verdict
**approved** (informational — no Critical findings; the human GitHub approval is a separate action). Posting as a COMMENT per the review policy. The two suggestions above are worth a look but none block merge.

jwfing (commented):
## Review: fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A well-scoped, well-tested fix for the streamable-HTTP runtime-session leak plus SSE keepalive; the reaper correctly limits itself to streamable sessions and gates on both Redis state and an in-memory monotonic clock, so I found no blocking issues.

### Requirements context
`insforge-mcp` has no `docs/superpowers/` or other spec directory (CLAUDE.md is a publishing guide only), so — **no matching spec/plan found; assessed against the PR description, the linked production signature (510 active / 4443 resident sessions on the 4GB box), and the surrounding code.** The core design constraint here is repo-specific and known: a background sweeper over `runtimeSessions` **must not** touch SSE sessions, because the SSE message path never refreshed the Redis record, so an SSE record lapses at TTL-from-creation while the connection is still live. This PR honors that constraint (`sweepOrphanedSessions` filters `transportType === 'streamable'`, `session-manager.ts:461-467`) and covers it with a dedicated test ("never reaps an SSE session, even with no Redis record").

---

### Critical
(none)

### Suggestion

**Software engineering — the real keepalive + stream-tracking wiring in `server.ts` is untested.**
`keepalive.test.ts` deliberately tests a *mirror* of the handler (`attachKeepAlive`) rather than the code that ships, because `server.ts` calls `startServer()` at import time and can't be imported (`keepalive.test.ts:9-14`). As a result the actual inline SSE keepalive (`server.ts:1011-1023`) and the `openStream`/`closeStream` reference-counting wired into `GET /mcp` (`server.ts:905-906`) have **no direct test** — only the pure sweep/select functions and `positiveIntEnv` are covered. The pure-function extraction in `session-manager.ts` (`selectOrphanedSessions`, `selectPrematurelyMissing`) is exactly the right pattern; consider giving the keepalive the same treatment (extract a tiny exported `attachKeepAlive(res, ms)` helper into a module `server.ts` can import) so the shipped code — not a copy — is what the suite exercises. Low blast radius today, but a mirror test silently rots when the handler and the copy drift.

**Functionality — `GET /mcp` streamable notification streams get stream-tracking but no keepalive.**
The PR adds `openStream`/`closeStream` around the streamable server→client stream (`server.ts:902-909`) specifically because such streams stay open with no other trace, and the reaper now protects them (`openStreams > 0` ⇒ held). But no `: keepalive` frame is written on that stream — only the legacy `/sse` path gets one (`server.ts:1011-1023`). So an idle streamable notification stream behind the same ALB is still dropped at the 60s idle timeout, which is the exact failure the SSE keepalive fixes. The reaper's open-stream protection can't really hold for longer than the LB timeout without it. This is outside the PR's stated scope ("keep **SSE** streams alive"), hence a suggestion rather than a blocker — but adding the same 25s comment frame to the `GET /mcp` handler would make the new open-stream guard meaningful in production.

### Information

**Security — no security-relevant changes.** No new user input reaches SQL/shell/HTTP; no auth/authz path is touched. `heapStats()` (`server.ts:139-148`) adds `heapUsedMb/heapLimitMb/heapUsedPct/rssMb` to the already-public `/health` payload — low-sensitivity infra numbers, no secrets/PII. The new SSE-refresh failure log uses `tokenFingerprint(sessionId)` (`server.ts:1118`), consistent with existing hygiene. Nothing to flag.

**Functionality — SSE keepalive shares the response with the SDK's transport writes.** The keepalive does `res.write(': keepalive\n\n')` on the same `res` the `SSEServerTransport` writes events to. This is safe **as long as** the SDK emits each event in a single `res.write` call (a comment frame between two complete events is valid SSE); it would only corrupt framing if the SDK ever split an event across multiple writes. Current SDK behavior is a single write per event, so this holds — noting the assumption since it's load-bearing and undocumented at the call site.

**Software engineering — `positiveIntEnv('1500ms')` returns `1500`.** `parseInt` trailing-junk tolerance means `MCP_SESSION_SWEEP_MS=1500ms` is silently accepted as `1500`. Explicitly covered by a test and harmless; noting only.

**Performance — clean.** The sweep is a single pipelined `EXISTS` batch every 5 min (O(resident streamable sessions), one round-trip), timers are `unref()`'d and stopped on shutdown, and the added `touchSession` calls on `POST /messages` / `GET /mcp` are fire-and-forget on the SSE path. No N+1, no unbounded loop, no blocking work on the event loop, no per-request large allocations. The monotonic-clock rationale (`monotonicNow`, `session-manager.ts:60-72`) is the correct call — a forward wall-clock step would otherwise age every session past TTL at once and defeat the whole gate.

---

No Critical findings. The design is careful where it counts: alive-on-error/alive-on-non-integer EXISTS handling, the snapshot re-check before close, the reconnect-safe stream **count** (not flag), and the premature-missing operator warning are all correct and thoroughly tested (28/28 vitest, 10 new). The two suggestions (test the shipped keepalive wiring; extend the keepalive to `GET /mcp` streams) are worth a follow-up but do not block merge.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review — fix(http): reap orphaned runtime sessions and keep SSE streams alive

**Summary:** A well-scoped, thoroughly-tested fix that correctly reads Redis as the source of truth to reap orphaned streamable sessions and adds SSE keepalive + a `lastSeenAt`/`openStreams` activity model; no blocking issues found for the standalone-Redis target it addresses.

### Requirements context
No `/docs/superpowers/` (or any `docs/`) directory exists in this repo — no matching spec/plan found, so I assessed against the PR description, the cubic/CodeRabbit summaries, and the surrounding code. The two problems this repo has historically had here — SSE `POST /messages` never refreshing the record (so SSE sessions lapsed at TTL-from-creation and `/health` under-reported them) and a sweeper that must not touch SSE — are both directly and correctly addressed (`server.ts:1114-1122`, `session-manager.ts:461-469`).

### Findings

#### Critical
(none)

#### Suggestion

**1. Functionality — the sweep's single cross-slot pipeline will reject under multi-shard Redis Cluster** (`src/http/session-manager.ts:479-484`)
`sweepOrphanedSessions` builds one `redis.pipeline()` and queues `EXISTS mcp:session:<uuid>` for every resident session. `redis.ts:32-44` supports `REDIS_CLUSTER=true` explicitly "for AWS ElastiCache Serverless". In ioredis Cluster mode, `Pipeline.exec()` runs `generateMultiWithNodes` and rejects unless *every* key in the pipeline is served by the same node allocation group (confirmed against ioredis `lib/Pipeline.ts`). Session ids are random UUIDs with no hash tag, so on a multi-shard cluster they scatter across nodes and `exec()` throws `"All keys in the pipeline should belong to the same slots allocation group"`. That rejection is caught by `startIdleSweep`'s `.catch` and logged — so the reaper would silently reap nothing and log an error every interval, i.e. the leak fix is a no-op in that topology. Standalone Redis and single-shard clusters are unaffected, and the PR body describes a co-located standalone Redis on the 4GB box, so the immediate target is fine. But since Serverless (which shards) is a documented supported config, consider replacing the single pipeline with per-key `exists` via `Promise.all` (ioredis Cluster routes each key to its owning node) or grouping keys by node. **If the production Redis behind `mcp.insforge.dev` is ElastiCache Serverless / multi-shard, treat this as blocking.**

**2. Functionality — the streamable `GET /mcp` notification stream has no keepalive** (`src/http/server.ts:902-909`)
The PR adds a 25s comment-frame keepalive to the legacy `/sse` stream, and correctly guards the streamable session from reaping while `openStreams > 0`. But the streamable server→client stream established by `GET /mcp` is itself a long-lived idle HTTP response that an ALB will drop on its 60s idle timeout, exactly like the SSE case — the session survives, but the client's stream still gets torn down. Consider attaching the same keepalive to `GET /mcp` for symmetry.

**3. Software engineering — `keepalive.test.ts` exercises a mirror of the handler, not the handler** (`src/http/keepalive.test.ts:37-49`)
Because `server.ts` calls `startServer()` at import time it can't be imported under test, so `attachKeepAlive` re-implements the SSE handler's interval/close logic and the tests assert against the copy. The comment is honest about this, but it means the real `server.ts:1012-1024` keepalive code has no direct coverage and could drift from the tested mirror. Extracting the keepalive into an exported helper that both the handler and the test call would close the gap. (`positiveIntEnv` is well covered — good use of `it.each` for the `-1/0/abc/''` fallbacks.)

#### Information

- **Security — no material concerns.** No new dependencies (only `node:v8` built-in + global `performance`), no new user input reaching SQL/shell, no auth changes. Session ids in the new SSE refresh log are hashed via `tokenFingerprint` (`server.ts:1121`). The only new external surface is `/health` now returning `heapUsedMb/heapLimitMb/heapUsedPct/rssMb` (`server.ts:163`) — trivial info disclosure if that endpoint is public, acceptable for an ops signal.
- **Performance — no hot-path concerns.** The sweep is an unref'd 5-minute timer doing one pipelined round trip; `touchSession` on the SSE message path is fire-and-forget. The reap loop `await`s `deleteSession` sequentially (`session-manager.ts:509-517`) — fine at the ~45-orphans-per-pass scale described; only worth batching if single-pass reclaim counts grow into the thousands.
- **Style — stale JSDoc placement** (`src/http/session-manager.ts:375-377`): the `/** Update last accessed time and refresh TTL */` block that documents `touchSession` now sits directly above `openStream`, with `openStream`/`closeStream`/`getOpenStreamCount` inserted between it and `touchSession`. Cosmetic.

### Verdict
`approved` (informational — a human still gives the GitHub approval). Zero Critical findings: the correctness reasoning is careful (fail-alive on Redis errors/non-integer replies, TTL age-gate against a cold Redis wipe, monotonic clock to resist forward clock steps, snapshot re-check before close, count-not-flag stream tracking), and test coverage for those cases is excellent. The suggestions above — chiefly the cluster-mode pipeline behavior — are worth resolving but do not block the standalone deployment this PR targets.

PR #72, opened 2026-07-28
fix(redis): actually send the password, and accept a REDIS_URL

tonychang04:
Blocks the trial deploy right now. Found by @Iris while diagnosing the crash; handed to me rather than cutting across #71's branch.

## The bug

`REDIS_PASSWORD` was read into a `REDIS_CONFIG` object in `config.ts` **that nothing imported** — one reference in the whole repo, its own definition. The real client in `redis.ts` built its own options from `{host, port, tls, cluster}` and never passed a password to ioredis.

So the server cannot connect to any Redis that requires AUTH, while the environment variable sits there looking correct. Nearly every managed Redis is password-protected, so this blocks running anywhere except our own ElastiCache — which authorises by security group rather than AUTH, which is why it has never bitten us.

Measured against a password-protected Redis with `REDIS_PASSWORD` correctly set:

```
master:    [Redis] Connection error: NOAUTH Authentication required.
           Failed to start server: ReplyError: NOAUTH Authentication required.
           exit code 1                       ← indistinguishable from "no Redis at all"
this PR:   /health → ok
```

## What changed

- **Password and username are threaded through** for standalone and cluster (cluster takes them via `redisOptions`).
- **`REDIS_URL` is supported** — the shape managed Redis is normally handed over in: one variable with credentials embedded rather than five. ioredis parses user, password, host, port and the `rediss://` TLS switch out of it. A URL wins over the discrete fields, but does **not** override cluster mode, which is configured by node list.
- **The duplicate config is deleted, not fixed.** A second unimported copy of the Redis settings is exactly what made a missing feature look like a present one. `redis.ts` is now the single place Redis config is read, next to the client it configures.
- **Options are built in `resolveRedisClientSpec`**, so what ioredis is handed can be asserted without opening a socket. The previous inline shape couldn't be tested at all, which is why a password that never arrived went unnoticed.

## Verification

11 new tests covering: password reaches ioredis, username for ACL auth, credentials omitted when unset, TLS on/off, URL used verbatim, URL beating discrete fields, URL *not* hijacking cluster mode, and cluster credentials landing in `redisOptions`.

End-to-end against a real password-protected Redis (`redis-server --requirepass`): `master` exits 1 with `NOAUTH`; this branch serves `/health` both via `REDIS_PASSWORD` and via a single `REDIS_URL`.

29/29 suite, `tsc` and `eslint` clean.

## Blast radius

Redis connection setup — every deployment. Behaviour with our current env is unchanged: no `REDIS_URL`, no `REDIS_PASSWORD`, `REDIS_CLUSTER=true` resolves to the same cluster client with the same node, TLS and timeouts as before. The change only adds paths that previously didn't work.

Worth knowing for the trial: `REDIS_CLUSTER=true` is right for our ElastiCache and wrong for a single-node platform Redis.

---
## Summary by cubic
Fix Redis auth by sending credentials to `ioredis` and adding `REDIS_URL` support. Removes the unused config so managed Redis works and trial deploys are unblocked.

- **Bug Fixes**
  - Thread `username`/`password` through for standalone and cluster (via `redisOptions`); TLS still supported.
  - Remove the unused Redis config in `src/http/config.ts` that read `REDIS_PASSWORD` but was never used.

- **New Features**
  - Support `REDIS_URL` (e.g., `rediss://user:pass@host:port`), parsed by `ioredis`.
  - URL takes precedence over discrete host/port/TLS, but does not override cluster mode.

<sup>Written for commit 69a03a3f0bd9da3229b5ffaebef53b1c2b2491bf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/72?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring Redis connections with a full URL and optional username/password credentials.
  * Improved TLS and cluster connection configuration.

* **Bug Fixes**
  * Corrected credential handling for standalone and clustered Redis connections.
  * Preserved retry settings across supported connection modes.
  * Ensured URL-based configuration takes precedence where applicable.

* **Tests**
  * Added comprehensive coverage for standalone, URL, TLS, credential, and cluster connection scenarios.

### DIFF
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -76,23 +76,11 @@ export const OAUTH_CONFIG = {
 // ============================================================================
 // Redis Configuration
 // ============================================================================
-
-export const REDIS_CONFIG = {
-  /** Redis host */
-  host: process.env.REDIS_HOST || 'localhost',
-
-  /** Redis port */
-  port: parseInt(process.env.REDIS_PORT || '6379'),
-
-  /** Redis password */
-  password: process.env.REDIS_PASSWORD || undefined,
-
-  /** Use TLS for Redis connection */
-  tls: process.env.REDIS_TLS === 'true',
-
-  /** Use Redis cluster mode */
-  cluster: process.env.REDIS_CLUSTER === 'true',
-} as const;
+//
+// Deliberately not here. Redis config is read in redis.ts, next to the client
+// it configures. A second copy lived here, unimported, and it advertised a
+// REDIS_PASSWORD that the real client never sent — so a password-protected
+// Redis failed to connect with the variable sitting there looking correct.
 
 // ============================================================================
 // Session Configuration
--- a/src/http/redis.test.ts
+++ b/src/http/redis.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect } from 'vitest';
+import { resolveRedisClientSpec, type RedisConfig } from './redis.js';
+
+/**
+ * These exist because of a bug that was invisible: REDIS_PASSWORD was read into
+ * a config object nothing imported, while the real client built its own options
+ * and never sent a password. Every managed Redis requires AUTH, so the server
+ * could not connect to one — and the env var sat there looking correct.
+ *
+ * The assertions are on the resolved spec rather than a live client, so they
+ * check what ioredis is actually handed without opening a socket.
+ */
+
+const base: RedisConfig = {
+  host: 'redis.internal',
+  port: 6379,
+  tls: false,
+  cluster: false,
+};
+
+describe('resolveRedisClientSpec — standalone', () => {
+  it('passes the password through to ioredis', () => {
+    const spec = resolveRedisClientSpec({ ...base, password: 's3cret' });
+    expect(spec.kind).toBe('standalone');
+    expect(spec.options.password).toBe('s3cret');
+  });
+
+  it('passes a username through for ACL-style auth', () => {
+    const spec = resolveRedisClientSpec({ ...base, username: 'app', password: 's3cret' });
+    expect(spec.options.username).toBe('app');
+    expect(spec.options.password).toBe('s3cret');
+  });
+
+  it('omits credentials entirely when none are configured', () => {
+    const spec = resolveRedisClientSpec(base);
+    expect(spec.options).not.toHaveProperty('password');
+    expect(spec.options).not.toHaveProperty('username');
+  });
+
+  it('keeps host, port and the retry policy', () => {
+    const spec = resolveRedisClientSpec(base);
+    expect(spec.options.host).toBe('redis.internal');
+    expect(spec.options.port).toBe(6379);
+    expect(spec.options.maxRetriesPerRequest).toBe(3);
+  });
+
+  it('enables TLS when asked', () => {
+    expect(resolveRedisClientSpec({ ...base, tls: true }).options.tls).toEqual({});
+    expect(resolveRedisClientSpec(base).options).not.toHaveProperty('tls');
+  });
+});
+
+describe('resolveRedisClientSpec — URL', () => {
+  // The shape managed Redis is normally handed over in: one variable with the
+  // credentials already embedded.
+  it('is used verbatim so ioredis can parse credentials out of it', () => {
+    const spec = resolveRedisClientSpec({
+      ...base,
+      url: 'redis://app:s3cret@redis.example:6380',
+    });
+    expect(spec).toMatchObject({ kind: 'url', url: 'redis://app:s3cret@redis.example:6380' });
+  });
+
+  it('carries the retry policy alongside the URL', () => {
+    const spec = resolveRedisClientSpec({ ...base, url: 'rediss://redis.example:6380' });
+    expect(spec.options.maxRetriesPerRequest).toBe(3);
+  });
+
+  it('takes precedence over the discrete host and port', () => {
+    const spec = resolveRedisClientSpec({
+      ...base,
+      host: 'ignored.example',
+      url: 'redis://redis.example:6380',
+    });
+    expect(spec.kind).toBe('url');
+    expect(spec.options).not.toHaveProperty('host');
+  });
+
+  it('does not hijack cluster mode', () => {
+    // Cluster is configured by node list, not by URL — a URL alongside it must
+    // not silently downgrade the client to standalone.
+    const spec = resolveRedisClientSpec({ ...base, cluster: true, url: 'redis://x:6379' });
+    expect(spec.kind).toBe('cluster');
+  });
+});
+
+describe('resolveRedisClientSpec — cluster', () => {
+  it('sends credentials via redisOptions, where the cluster client reads them', () => {
+    const spec = resolveRedisClientSpec({
+      ...base,
+      cluster: true,
+      username: 'app',
+      password: 's3cret',
+    });
+    expect(spec.kind).toBe('cluster');
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const redisOptions = (spec.options as any).redisOptions;
+    expect(redisOptions.password).toBe('s3cret');
+    expect(redisOptions.username).toBe('app');
+  });
+
+  it('still targets the configured node and enables TLS', () => {
+    const spec = resolveRedisClientSpec({ ...base, cluster: true, tls: true });
+    expect(spec).toMatchObject({ nodes: [{ host: 'redis.internal', port: 6379 }] });
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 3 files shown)


jwfing (commented):
## Review: fix(redis): send the password, accept a REDIS_URL

**Summary:** A correct, well-tested fix for a real bug — `REDIS_PASSWORD` was read into an unimported `REDIS_CONFIG` and never reached ioredis; this threads credentials through, adds `REDIS_URL` support, and makes option-building testable. No blocking issues.

**Requirements context:** No `/docs/superpowers/` or `docs/specs/` directory exists in this repo (consistent with insforge-mcp's known shape), so this is assessed against the PR description and linked context (#71 mentioned, not blocking). No matching spec/plan found.

I verified the ioredis usage against the ioredis v5 docs via context7: `new Redis(url, options)` (string-path + options overload), `rediss://` auto-enabling TLS, empty `tls: {}`, and cluster credentials via `redisOptions` are all correct — no stale or hallucinated API.

---

### Critical
(none)

### Suggestion

**Functionality — cluster mode silently drops URL-embedded credentials** (`src/http/redis.ts:64-83`)
The `if (cluster)` branch returns before the `if (url)` branch, so when both `REDIS_CLUSTER=true` and `REDIS_URL=redis://user:pass@…` are set, the URL — including its embedded credentials — is ignored entirely. Cluster auth then comes *only* from the discrete `REDIS_USERNAME`/`REDIS_PASSWORD`. The PR body documents "a URL does not override cluster mode," and the test at `redis.test.ts:80-83` asserts the mode isn't downgraded, but neither makes it obvious to an operator that URL credentials are being discarded. Given this is precisely the class of bug the PR fixes (a credential that looks configured but never arrives), consider a one-line `console.warn` when `url && cluster` are both set, so a misconfiguration surfaces at startup instead of as a `NOAUTH` at connect time. Low blast radius (your ElastiCache uses security-group auth, per the PR), hence non-blocking.

### Information

- **Startup banner misreports the endpoint under `REDIS_URL`** (`src/http/server.ts:1138`) — the banner prints `${redisConfig.host}:${redisConfig.port}`, which stays at the `localhost:6379` defaults when the connection is actually driven by `REDIS_URL`. Cosmetic/observability only, but could mislead during the trial deploy this PR unblocks. (Correctly, it does *not* print the password or URL — no secret leak.)
- **Discrete `REDIS_TLS` is ignored in URL mode** (`src/http/redis.ts:79-83`) — TLS in URL mode depends solely on the `rediss://` scheme; a `redis://` URL plus `REDIS_TLS=true` yields no TLS. This is the documented "URL wins" behavior and is reasonable; noting it for completeness.
- **New env vars aren't in `.env.example`** — `REDIS_URL` and `REDIS_USERNAME` are new, but `.env.example` currently documents no Redis vars at all, so this is a pre-existing gap rather than a regression.
- **Cluster path doesn't carry `BASE_OPTIONS`** (`src/http/redis.ts:66-76`) — `maxRetriesPerRequest`/`retryStrategy` are applied to standalone and URL specs but not to the cluster `redisOptions`. This matches the pre-PR behavior (the old inline cluster options had none either), so it's not a regression — flagging only in case parity was intended.

---

**Notes on the four dimensions:**
- *Software engineering:* Extracting `resolveRedisClientSpec` so the constructed options can be asserted without a socket is the right call, and the 11 new tests cover the meaningful branches (password/username through, credentials omitted, TLS on/off, URL verbatim, URL precedence, URL-not-hijacking-cluster, cluster credentials in `redisOptions`). Test assertions are consistent with the implementation. Good TDD-shaped design.
- *Functionality:* The fix does what it claims; API verified. Edge cases are documented, with the one cluster+URL nuance noted above.
- *Security:* Strengthens auth (password now actually sent); no secrets logged or returned; no new dependencies. No security-negative changes.
- *Performance:* Connection-setup only, behind a singleton — no hot-path, allocation, or event-loop concerns.

### Verdict: **approved** (informational; posted as a comment — a human still gives the GitHub approval)
Zero Critical findings. The Suggestion and Information items are non-blocking; the cluster+URL warning is the one worth a moment's thought before the trial deploy.

jwfing (commented):
## Review: fix(redis): actually send the password, and accept a REDIS_URL

**Summary:** A correct, well-tested fix for a real silent-misconfiguration bug — the password is now threaded to ioredis for standalone and cluster, `REDIS_URL` is supported, and the dead duplicate config is removed; no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (nor `docs/specs/`) directory. Assessed against the PR description, the linked bug narrative, and the surrounding code (`src/http/redis.ts`, `src/http/server.ts`, `src/http/config.ts`). The core claim in the PR body checks out: `REDIS_CONFIG` in `config.ts` had exactly one reference (its own definition — now deleted), and the live client in `redis.ts` never received a password. API usage was verified against the ioredis 5.x docs (URL-string constructor, `redisOptions` credentials for cluster, `tls: {}`) — all correct, nothing stale or hallucinated.

---

### Critical
(none) — the fix is correct and behavior for the existing env (no `REDIS_URL`, no `REDIS_PASSWORD`, `REDIS_CLUSTER=true`) is unchanged.

### Suggestion

**Functionality — startup banner misreports the endpoint when `REDIS_URL` is used** (`src/http/server.ts:1138`)
The banner prints `${redisConfig.host}:${redisConfig.port} (TLS: ${redisConfig.tls}, Cluster: ${redisConfig.cluster})`. When connecting via `REDIS_URL`, `host`/`port`/`tls` are still the discrete defaults (`localhost:6379`, TLS `false`) while the client actually connects to the URL's target. During exactly the trial-deploy diagnosis this PR targets, an operator reading the banner to confirm where the server connected would be misled. Consider deriving the displayed endpoint from the resolved spec / URL, or appending `(via REDIS_URL)` when a URL is in effect.

**Functionality — silent precedence when a URL and discrete credentials are both set** (`src/http/redis.ts:79-83`)
The `url` branch returns `{ ...BASE_OPTIONS }` only, dropping `credentials` and `tlsOptions`. This "URL wins" rule is documented and defensible, but it reproduces the same *silent* failure mode this PR fixes: setting `REDIS_URL=redis://host:6379` alongside `REDIS_PASSWORD=secret` (or `REDIS_TLS=true`) yields an un-authenticated / non-TLS connection with no signal. Similarly, cluster mode (`redis.ts:64-77`) ignores `REDIS_URL` entirely, including any credentials embedded in it. A one-line `console.warn` when discrete creds / URL are set but silently ignored would make the precedence visible and cheaply prevent a repeat of the exact bug class being fixed here.

**Software engineering — no test for the `createRedisClient` dispatch** (`src/http/redis.ts:95-107`)
`resolveRedisClientSpec` is thoroughly covered (11 cases), but the `switch` that turns a spec into a client — particularly the `url` arm calling `new Redis(spec.url, spec.options)` and the cluster arm — has no test. Low blast radius since the interesting logic is in the resolver, but a construction smoke test (e.g. asserting the client instance kind per spec) would close the loop and guard the `new Redis(url, options)` call shape.

### Information

- **Type modeling of the cluster options** (`src/http/redis.ts:75`, `101`): the cluster arm's `options` is really `ClusterOptions` (it carries `redisOptions`, `dnsLookup`, `slotsRefreshTimeout`, `enableReadyCheck`) but is typed as `RedisOptions` via `as RedisOptions`, then re-cast `as any` at construction. Functionally correct and matches the shape ioredis expects, but typing that arm as `ClusterOptions` would drop both casts and restore type safety. Non-blocking.
- **Confirmed: leaving `retryStrategy`/`maxRetriesPerRequest` out of the cluster `redisOptions` is correct** — ioredis explicitly excludes `retryStrategy` from `redisOptions`, so this matches both the pre-PR behavior and the library contract.
- **Security — no secret leakage:** the startup banner logs only host/port/TLS/cluster; password and full URL are never logged or returned in any API response. No new user-controlled input reaches the Redis config (all values are operator-set env vars). No security-relevant regressions.
- **Performance:** no concerns — changes are confined to one-time connection setup; no new per-request work, loops, or allocations.

---

### Verdict: **approved** (informational — a human still gives the GitHub approval)
No Critical findings. The bug is real, the fix is correct and verified against ioredis, and the new tests assert the resolved spec without opening a socket — a genuinely good testability improvement. The suggestions are about surfacing the (deliberate) precedence rules so this class of silent misconfiguration doesn't recur; none block merge.

jwfing (approved):
LGTM - approved.

PR #73, opened 2026-07-29
chore(ops): post-deploy discovery check (split out of #69)

tonychang04:
Split out of #69 on Max's call. It was part of that PR, drew **8 of its 10 review comments**, and dismissed an approval the discovery fix already had. #69 is now three files and the fix only.

This can take as many review rounds as it needs — and it should, since one finding was a real shell injection rather than style.

## What it is

```
npm run verify:deploy -- <url> [expected-version]
```

Read-only: fetches discovery documents and makes one unauthenticated MCP request. Registers nothing, writes nothing, exits non-zero on the first failed assertion so it can gate a deploy. Runs daily on a schedule and on demand against any URL.

It walks the whole chain: the 401 carries a challenge → the document it names exists → every derived URL resolves to a document whose `resource` matches the identifier it was derived from → the authorization servers it names resolve at their own RFC 8414 URLs with matching `issuer` → OAuth credentials are actually configured.

## Review finding already applied

**Shell injection (MAJOR)** — dispatch inputs were interpolated into `run:`; `expected_version` unquoted and `url` able to terminate its quotes. Both now go through `env:`, plus `permissions: contents: read` and `persist-credentials: false`.

## Still open, and why they belong here

Three findings from cubic/CodeRabbit remain, all hardening of this script:

- follow each endpoint document's own `authorization_servers` rather than only the origin's
- reject non-HTTPS issuers and treat URL-parse failures as assertion failures
- validate array shape with `Array.isArray` rather than a truthy `length`

I'll take these here. None of them should hold up a discovery fix that has users waiting.

## Discrimination

It has to fail on real problems to be worth anything:

- **8/15** against production today — missing `WWW-Authenticate`, both derived URLs 404
- **21/21** against a correct server
- **13/18** against `mcp.instacloud.com` — catches insta-platform#145 from the MCP side

---
## Summary by cubic
Adds a post-deploy verification tool and a real-client handshake to validate OAuth/MCP end-to-end on deployed servers. Tightens `/health` and authorize checks, adds per-request deadlines, and prevents scoring off-target hosts.

- **New Features**
  - `npm run verify:deploy -- <url> [expected-version]` checks health/version, `WWW-Authenticate`, well-known resource docs, AS metadata, and that `/oauth/authorize` validates without credentials; scheduled daily and manual workflow (manual requires `expected_version`); inputs via `env`.
  - `scripts/verify-handshake.mjs` runs a real `@modelcontextprotocol/sdk` client through discovery, dynamic registration, PKCE/code exchange, initialize, and `tools/list` with a one-time browser sign-in.

- **Bug Fixes**
  - Gate only the target: scope AS checks by the server’s self-identity; print OFF-TARGET for other hosts and exclude them from scoring; follow each document’s own `authorization_servers`; validate with `Array.isArray`; reject unparseable or non-HTTPS issuers (loopback allowed).
  - Robustness: treat network failure as a value and exit after `/health` when unreachable; add per-request timeouts via `AbortSignal.timeout` (default 10s, `VERIFY_TIMEOUT_MS` override); workflow runs with `contents: read` and `persist-credentials: false`.
  - Health: assert `/health` names `insforge-mcp` (not a platform placeholder) and always report a version; add a basic session leak check (resident vs active).
  - OAuth probe: require `/oauth/authorize` to return 400 (not 302), and fail when credentials are missing (500).
  - Handshake: fix the success path by reconnecting with a fresh client/transport after authorization; add callback listener error handling and keep rejections handled; refuse non-HTTPS targets unless loopback; add script globals for linting.

<sup>Written for commit 7dad9c4f7ecb2af2a290426bcdfe5c7011bc437b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/73?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automated post-deployment verification for service health, OAuth discovery behavior, protected-resource metadata, and authorization endpoint reachability.
  * Added a scheduled, manually triggerable workflow to run the deployed verification against a specified base URL.
  * Added local verification commands, including an end-to-end OAuth handshake check that confirms tool registration.
* **Reliability**
  * Produces automation-friendly PASS/FAIL output and fails when required target checks don’t succeed.
* **Chores**
  * Enhanced linting support for standalone Node ESM scripts and added a `verify:deploy` command.


jwfing (commented):
## Review: post-deploy discovery check (split out of #69)

**Summary:** A read-only post-deploy verification script (`scripts/verify-deploy.mjs`) plus a scheduled/dispatch workflow that walks the OAuth discovery chain and gates on the first failed assertion — well-scoped, and the previously-flagged shell-injection is correctly fixed; the three remaining hardening items you called out are real but non-blocking.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessing against the PR description, the RFC references it cites (RFC 9728 §3.1/§3.3, RFC 8414 §3.1/§3.3), and the actual server behavior in `src/http/server.ts`. I verified the script's key server-side assumptions against the checked-out head:
- `/oauth/authorize` does check credentials (`500 server_error`, `server.ts:244-249`) **before** validating parameters (`400 invalid_request`, `server.ts:251-256`) — so the `!== 500` credential probe is sound.
- `/health` returns `version` and `sessions: { activeSessions, memorySessionCount }` (`server.ts:135-145`, `session-manager.ts:339-362`) — the session-drift heuristic reads the right fields.
- The origin `/.well-known/oauth-protected-resource` currently returns `resource: ${baseUrl}/mcp` with a single origin route and no `/mcp`|`/sse` suffix routes (`server.ts:169-178`). The script's stricter checks are therefore **intentionally aspirational** — this is the RFC drift the tool exists to catch, consistent with "expect RED until deployed current." Not a script defect.

### Findings

#### Critical
(none) — the shell-injection fix (dispatch inputs via `env:`, quoted `"$DEPLOY_URL"`/`"$EXPECTED_VERSION"`, `permissions: contents: read`, `persist-credentials: false`) is correct, and nothing here blocks merge.

#### Suggestion

*Functionality*
- **`scripts/verify-deploy.mjs:167` — `asMetadataUrl(issuer)` throws outside the try/catch.** `new URL(issuer)` (line 158) runs at line 167, *before* the `try` at line 168. A malformed / non-URL entry in `authorization_servers` throws an uncaught exception and the script dies with a stack trace instead of a clean `FAIL`. It still exits non-zero (so the deploy is still gated), just not as an assertion. This is your acknowledged item #2 — folding in HTTPS-issuer rejection and treating URL-parse failure as a failed `check(...)` resolves both at once.
- **`scripts/verify-deploy.mjs:163-166` — `authorization_servers` shape not validated with `Array.isArray`.** `?? []` only defaults on null/undefined; a string value slips through, `authServers.length` becomes the string length, and `for (const issuer of authServers)` iterates characters into nonsense fetches. The named-document check at line ~140 already uses `Array.isArray` — worth making the origin path consistent. Your acknowledged item #3.
- **`scripts/verify-deploy.mjs:163` — only the origin document's `authorization_servers` are followed.** The AS-chain loop reads `originDoc.body?.authorization_servers`, so a per-endpoint (`/mcp`, `/sse`) document naming a different AS goes unchecked. Your acknowledged item #1.
- **`scripts/verify-deploy.mjs:91` — a cross-origin `resource_metadata` is silently re-pointed at `base`.** When `named` doesn't start with `base`, the code takes `new URL(named).pathname` and then `getJson(path)` fetches `base + path` — i.e. it checks the wrong host rather than the one the challenge actually named. Same-origin in practice, but a drifted deployment naming a foreign host (exactly the failure this tool targets) would be masked as a pass. Consider fetching the absolute `named` URL directly.

*Software engineering*
- **No tests for the pure helpers.** TDD is a stated repo convention and the URL-derivation logic (`asMetadataUrl`, the `named`→path derivation at line 91) is pure and trivially unit-testable with a `fetch` mock. A test over `asMetadataUrl` with a malformed issuer would have surfaced the throw-outside-try above. Understood that this repo has no `scripts/` test precedent, so non-blocking — but a small `scripts/verify-deploy.test.ts` around the pure bits would be cheap insurance.

#### Information

*Software engineering*
- `eslint.config.js:10-17` is the right fix: the base config uses `tseslint.configs.recommended` (non-type-checked), so the `.mjs` — excluded from `tsconfig.json` (`rootDir: ./src`) — parses syntax-only without a `parserOptions.project` error, and the explicit `process/console/fetch/URL` globals satisfy `no-undef` (ES built-ins like `JSON`/`Math` come from `ecmaVersion`). `npm run lint` should stay green.
- `.github/workflows/verify-deploy.yml:38` pins actions to `@v4` but `node-version: 'lts/*'` floats to the latest LTS. Fine for an ops check; noting only for reproducibility.

*Security* — no secrets are read or logged; the script only prints check names and non-sensitive detail (statuses, `resource`/`issuer` values). The workflow hardening is good.

*Performance* — a handful of small sequential fetches on a schedule; no hot-path, allocation, or N+1 concerns.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). No Critical findings. The three hardening items you flagged as open are legitimate Suggestions and worth taking in this PR as you planned, but none block merge.

jwfing (dismissed):
LGTM - approved.

tonychang04:
`d76c441` — @qa-bot found this gate scoring a server it wasn't asked about, and I've fixed it rather than documenting around it. **This does dismiss the approval; I think that's the right trade** — a gate that reports four passes earned by the machine we're replacing is worse in a cutover checklist than one more review round.

## The defect

The script followed `authorization_servers` wherever it pointed and counted the answers as its own. Before a DNS flip that's the wrong machine: the box under test advertises the **final** hostname as its authorization server, that hostname still resolves to the box being **replaced**, and four checks go green because the old machine is up.

Assertions now carry a scope. Anything describing a host other than the one on the command line prints as `OFF-TARGET`, never `PASS`, and is excluded from the score — with a closing note naming the host and saying that a re-run against the final hostname after propagation is what actually gates the deploy. An authorization server that **is** the target still counts normally.

## The three review findings from this PR, same pass

- **Follow every document's own `authorization_servers`,** not just the origin document's. A `/mcp` document pointing somewhere else is exactly the skew this script exists to catch, and it was invisible.
- **Validate with `Array.isArray`.** A bare string passed `.length > 0` and was then iterated character by character.
- **Reject an issuer that is unparseable or not https** instead of fetching it. Loopback is exempt so the script can be pointed at a local server. Previously a malformed issuer threw out of `new URL()` and killed the run, and a plaintext one produced a real outbound request to a host named by the document under test.

Also: it now says so when no expected version was passed. A container serving the right name at the wrong version passes every other assertion here — which is the failure this whole thread started from.

## Measured, two local servers, one standing in for the old box

Server on `:8393` under test, configured with `:8394` as its public URL — Quinn's Case 1 without touching prod.

| | result |
|---|---|
| before | `11/18`, including **4 PASSes earned by `:8394`** |
| after | `8/17`, those 4 printed `OFF-TARGET` and not counted |
| control — AS is the target itself | scored normally, `12/21` |
| control — plaintext issuer | fails the check, no request made |

```
OFF-TARGET  AS metadata for http://127.0.0.1:8394 is 200 at its derived URL
            — ok, but describes another host

8/17 checks passed against http://127.0.0.1:8393

4 check(s) describe http://127.0.0.1:8394 and were NOT counted.
```

**Not verified by me:** a real second deployment or DNS propagation — same limits Quinn stated. The two-server setup reproduces the document topology, not the DNS behaviour.

tonychang04:
`b338b27` — the version assertion is now **required** for `workflow_dispatch`.

It was an optional input defaulting to empty, so the one check that catches a container serving the right name at the wrong build could be skipped by leaving a form field blank. That's the same failure class @qa-bot just found: a gate that silently drops its strongest assertion. Production sitting 139 days stale while every other check stayed green is exactly what that check exists to catch.

The scheduled run still asserts no version — there's no single correct one on an arbitrary day — and the script now says plainly in its output when that check didn't run.

The input description names where to read the number, because the obvious sources disagree:

```
master package.json   1.2.10
npm latest            1.2.11
#70 branch            1.2.12   (draft, merges last)
```

Master is **behind** the published package, not ahead of it. A remembered number is as likely to produce a false green as no number at all, so the description points at the `package.json` of the commit being deployed rather than at any of these.

jwfing (commented):
## Review — `chore(ops): post-deploy discovery check` (#73)

**Summary:** A read-only post-deploy probe (`scripts/verify-deploy.mjs`) plus a scheduled/dispatch workflow that walks the full OAuth/MCP discovery chain against a deployed server; the implementation is correct, RFC-faithful, and the previously-flagged shell-injection is properly remediated.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, its parent #69, and the actual server implementation in `src/http/server.ts`. I cross-checked every assumption the script encodes against the live handlers.

### Verification against the codebase
The script's assertions match the server's real contracts:
- `/health` shape — `version` + `sessions.{activeSessions,memorySessionCount}` confirmed at `src/http/server.ts:135-144` and `src/http/session-manager.ts:361-362`.
- `/oauth/authorize` two-state logic — `500 server_error` when unconfigured vs `400 invalid_request` on a bare request — confirmed at `src/http/server.ts:244-256` (`isOAuthConfigured()` gate precedes parameter validation). The `!== 500` / `=== 400||302` split is exactly right.
- The stricter checks (per-endpoint `oauth-protected-resource/{mcp,sse}` docs, origin `resource === base`, a `WWW-Authenticate` challenge) intentionally diverge from today's server (`src/http/server.ts:169-177` returns only the origin doc with `resource: baseUrl/mcp`, and no `WWW-Authenticate` header is emitted). This is the drift the tool is built to catch and matches the PR's "expect RED until deployed current" framing — encoding target behavior, not current behavior, is the right call for an acceptance gate.

### Findings

#### Critical
(none)

#### Suggestion
- **[software engineering / robustness]** `scripts/verify-deploy.mjs:44-48, 57, 80, 103, 151` — the AS-metadata loop wraps `fetch` in try/catch (`:237-241`) and `parseIssuer` cleanly converts bad URLs into assertion failures, but the earlier calls (`getJson` for `/health`, the raw `/mcp` POST, the named doc, the derived docs, `/oauth/authorize`) do not. A DNS/TLS/connection-refused failure — the exact case when a host isn't deployed yet — throws an unhandled rejection with a stack trace instead of a clean `FAIL … — <reason>`. Same for `new URL(base).origin` at `:27` on a malformed argument. The process still exits non-zero, so the gate still "fails," but a probe designed to be pointed at not-yet-live hosts should degrade to a scored failure rather than a crash. Applying the try/catch pattern you already use in the AS loop to the top-level calls would make the whole run consistent.
- **[software engineering / test coverage]** `scripts/verify-deploy.mjs` — no automated test accompanies the script, and the repo gates on vitest. A live-server probe is inherently hard to unit-test, which is fair, but the *pure* helpers — `asMetadataUrl()` (RFC 8414 §3.1 well-known insertion), `parseIssuer()` (https/loopback policy), and the `derivedFrom`/`named` slicing logic (`:103`, `:113`) — are deterministic and are precisely the kind of RFC-derivation code that "every implementation got wrong in some direction" per your own comment. Extracting them and adding a small vitest would lock the derivation rules and guard against regressions; the PR body itself notes one earlier finding was "a real shell injection rather than style," underscoring that these ops scripts carry real risk and deserve the same test discipline as `src/`.

#### Information
- **[functionality]** `scripts/verify-deploy.mjs:103` — when `resource_metadata` is an absolute URL, the script extracts `new URL(named).pathname` and then refetches it against `base` (`getJson` = `fetch(base + path)`), so a cross-origin `resource_metadata` value would be silently re-fetched from the target host rather than the host it named. Likewise, a `base` that carries a path prefix (e.g. `https://host/prefix`) would mis-derive well-known paths via the `slice`. Both are non-issues for the origin-root, same-host deployments this targets today, but a one-line comment noting the same-origin assumption would help the next reader. Also note `new URL(named).pathname` drops any query string, which is fine for these well-known URLs but worth being aware of.
- **[security]** No security concerns. The workflow correctly routes both dispatch inputs through `env:` (`.github/workflows/verify-deploy.yml:50-54`), sets `permissions: contents: read` and `persist-credentials: false`, and the script is read-only with no secrets. The prior shell-injection is fully resolved.
- **[performance]** No concerns — bounded set of sequential HTTP probes, one per assertion group; nothing hot-path or unbounded.
- **[software engineering]** `eslint.config.js:10-17` — scoping runtime globals to `scripts/**/*.mjs` rather than loosening them repo-wide is the right minimal approach; `process/console/fetch/URL` covers everything the script references.

### Verdict
`approved` (informational — a human still gives the GitHub approval). Zero Critical findings. The tooling is correct, safe, and does what the PR claims; the suggestions above are hardening and test-discipline improvements, none blocking.

tonychang04:
## QA — `d76c441`, driven into its own failure modes

Since this script is now the cutover precondition, I tested the tool rather than the server: a stub that serves a discovery topology I control (`/health`, the 401 + challenge, the three protected-resource documents, AS metadata), so `authorization_servers` can be set to anything a real deployment might publish. Tests 18/18, lint 0 errors.

### The scope fix does what it says

| case | result |
|---|---|
| **AS is the target itself** (post-flip shape) | scored normally — `25/25`, exit 0 |
| **pre-flip** (target `:8393`, documents advertise `:8394`, old box up) | 4 checks `OFF-TARGET`, excluded, `18/21`, **exit 1** |
| `authorization_servers` is a bare **string** | 4 clean FAILs, and **9 requests total** — no character-by-character iteration (a 25-char string would have sprayed ~25) |
| **unparseable** issuer (`"not a url"`) | clean FAIL, run completes |
| **plaintext** issuer, pointed at a host I control that logs hits | FAIL, and that host recorded **zero** requests from the script |
| documents naming **different** issuers | all followed, each attributed to the documents that named it |

The plaintext one I checked properly rather than by reading: `/etc/hosts` entry so the request *would* have succeeded had it been made, and the target's hit log stayed empty.

### One thing the scope rule can't currently tell apart

`external` means "a different origin from the URL on the command line". That is exactly right for the pre-flip case. It also describes **a resource server that delegates to a separate authorization server** — which is the architecture this thread is moving toward (`api.insforge.dev` as the AS, per #756/#757).

In that shape every AS check is off-target forever, and a broken chain is excluded from the score rather than failing it:

```
stub self-consistent with the target, authorization_servers -> another host
21/21 checks passed against http://127.0.0.1:8393
4 check(s) describe http://127.0.0.1:8394 and were NOT counted.
exit 0
```

And when that other host is simply unreachable, its four failures print as `OFF-TARGET … FAILED` and still don't count:

```
OFF-TARGET  AS metadata for https://ok.example is 200 at its derived URL  — FAILED, but describes another host: … → ERR TypeError: fetch failed
21/23 checks passed
```

Not reachable today — the server hardcodes `authorization_servers` to its own base URL, so a self-consistent deployment always names itself. It becomes reachable the day the rewrite lands, and it fails in the green direction: exit 0 and a "21/21 checks passed" summary line with the auth chain unverified.

**A discriminator that keeps your fix and closes it:** the pre-flip case has a signature beyond "different origin" — the AS is the host the box *advertises as itself*, which is precisely why the three self-consistency checks fail in that run. So scope on that instead: if the target's documents are self-consistent with the URL under test, score the AS chain wherever it points; only when they are not (we're knowingly talking to a box advertising another hostname) treat the chain as off-target. Pre-flip behaviour is unchanged — that run is red on self-consistency anyway — and post-rewrite the gate keeps checking the platform AS.

### Smaller

- An unreachable target exits 1, but via an unhandled `TypeError: fetch failed` stack trace rather than a `FAIL GET /health` line. On cutover day the first thing an operator runs against a box that hasn't started yet is this, and a stack trace reads like the tool is broken.
- The no-expected-version note is a good addition; I'd make the checklist pass it always, since that check is what catches the failure this thread started from.

*Not tested:* real DNS propagation, a real second deployment (I have no Manufact hostname either), TLS/certificate behaviour on a real https AS, and prod.

— Quinn

tonychang04:
`79f7964` — both of @qa-bot's findings. The scope gap was the important one and his discriminator is better than my rule, so I took it as proposed.

## Scope on self-identity, not on origin difference

"Different origin from the URL I was given" also describes a resource server that legitimately delegates to a separate AS — where this codebase is heading. Under that rule a real failure of a real dependency printed `OFF-TARGET` and didn't count, so **a delegated AS that was simply down let the gate pass.**

The pre-cutover case has a signature beyond a differing origin: **the box doesn't agree about where it lives.** Reached on a temporary URL it still publishes the FINAL hostname as its own identifier, and that same hostname is what it names as its AS — so the chain leads to the machine being replaced. When its documents *do* agree with the URL under test, whatever they point at is the client's genuine next hop and is scored wherever it lives.

Compared by **origin** rather than exact string, so it holds whichever identifier convention the resource document uses for its path — it works both against master's `<base>/mcp` convention and #69's `<base>`. That comparison is now a check in its own right, so a box that doesn't know where it lives says so on its own line instead of only by implication.

## Unreachable target

It read as an unhandled fetch rejection and a stack trace, which looks like a broken script rather than a down server — and a box that hasn't finished starting is the first thing this gets pointed at on cutover day. Network failure is a value now, and a target that doesn't answer at all exits after one clear line rather than burying it under twenty identical failures:

```
FAIL  GET /health is 200  — unreachable (ECONNREFUSED)

http://127.0.0.1:8399 did not answer at all. Nothing else was checked.
exit 1
```

## Driven through six topologies

Stub that publishes an arbitrary discovery shape, same approach you used:

| topology | result |
|---|---|
| AS is the target | 25/25, exit 0 |
| **box agrees, delegated AS is DOWN** | **21/25, exit 1** — was passing |
| pre-cutover, old box still up | 17/21, 4 `OFF-TARGET`, exit 1 |
| target unreachable | one line, exit 1, no stack trace |
| `authorization_servers` a bare string | clean FAILs, no character iteration |
| plaintext issuer | FAIL, never fetched |

The closing note now explains *why* checks were excluded — naming the identifier the box published and the origin it was reached at — and states that a box whose documents agree has its AS scored wherever it lives, so nobody reads `OFF-TARGET` as a general property of third-party authorization servers.

**Still not verified by me:** DNS propagation, a real second deployment, TLS against a real https AS, or prod — same limits you listed.

jwfing (commented):
## Review: post-deploy discovery check (#73)

**Summary:** A well-scoped, read-only ops tool (`scripts/verify-deploy.mjs` + a scheduled/dispatch workflow) that walks the OAuth/MCP discovery chain against a deployed server; the logic is RFC-faithful and the previously-flagged shell injection is fixed — no blocking issues.

### Requirements context
`insforge-mcp` has no `docs/superpowers/` or any spec directory (confirmed: `docs/` is empty in the workspace), so there's **no matching spec/plan** — assessed against the PR description, the sibling PR #69 narrative, and the surrounding code (RFC 9728 §3.1/§3.3, RFC 8414 §3.1/§3.3). Note that the checks this script asserts (path-suffixed `oauth-protected-resource` docs, a root `resource` identifier, a `WWW-Authenticate` challenge) are **not** served by current `master` — the server today registers a single exact `/.well-known/oauth-protected-resource` route returning `resource: baseUrl + '/mcp'` (`src/http/server.ts:169-178`) and its 401s carry no challenge header (`src/http/server.ts:662-949`). That's expected: this is the *verifier* for the discovery fix in #69, and the workflow header states it will be RED until the deployment is current.

Verified against the workspace: `/health` shape (`server.ts:131-146`) and the session fields `activeSessions`/`memorySessionCount` (`session-manager.ts:361-362`) match the script's reads; the `/oauth/authorize` ordering the script depends on — credentials `500` checked *before* param `400` — is correct (`server.ts:244-256`).

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — false-green if `/oauth/authorize` alone is unreachable** (`scripts/verify-deploy.mjs:216-222`). `getJson` returns a *string* status like `"unreachable (...)"` on a network error. The check `authorize.status !== 500` then passes, so "OAuth credentials are configured" prints **PASS** without verifying anything. The companion check (`status === 400 || 302`) still fails, so the gate isn't fooled — but the PASS line is misleading. Consider requiring a numeric status (e.g. `authorize.status === 400 || authorize.status === 302`) for the credentials assertion too. Low blast radius since `/health` already gates whole-server reachability.
- **Functionality — `/sse` protected-resource doc is a hard scored assertion** (`scripts/verify-deploy.mjs`, the `for (const path of ['/mcp', '/sse'])` loop). SSE is the deprecated transport (`server.ts:920`). If #69's discovery contract does not publish `/.well-known/oauth-protected-resource/sse` (e.g. SSE isn't OAuth-gated or is being retired), this becomes a permanent target-scoped failure that blocks the gate against an otherwise-correct server. Worth confirming this doc is part of the discovery model #69 ships.
- **Software engineering — the RFC derivation helpers aren't unit-testable.** `asMetadataUrl`, `parseIssuer`, the `derivedFrom` strip, and the origin comparison are the highest-risk logic here — the script's own comments note "every implementation looked at got that rule wrong in one direction or another." They're pure functions but live inside a top-level-`await` script with no exports, so the repo's vitest gate can't cover them. Extracting them into a small importable module with tests for the RFC 8414 §3.1 path-insertion and the `resource`/`issuer` identifier match would catch a derivation regression in CI rather than only against a live deploy. Not blocking — this is itself an ops/test harness.

#### Information
- **Security — clean.** Dispatch inputs go through `env:` not shell interpolation (injection fix verified in `.github/workflows/verify-deploy.yml:56-62`), `permissions: contents: read`, `persist-credentials: false`, the script is strictly read-only (registers/writes nothing), carries no secrets, and rejects non-HTTPS issuers with a loopback exemption (`parseIssuer`). No security-relevant concerns.
- **eslint change is correct.** The `scripts/**/*.mjs` override (`eslint.config.js:10-17`) declares exactly the four non-ECMAScript globals the script uses (`process`, `console`, `fetch`, `URL`); language built-ins (`JSON`, `Math`, `Set`, `Map`, `Array`) are provided automatically by flat config, so `no-undef` is satisfied.
- **Performance — n/a.** Scheduled ops script making a small, bounded number of sequential fetches; no product hot path, no runtime impact. The sequential `await` over issuers is fine (1–2 issuers).
- Health-unreachable branch (`verify-deploy.mjs:~96-101`) records a `check(...)` result and increments `failed`, then prints its own `FAIL` line and `process.exit(1)` before the report loop runs — so that recorded result is never printed and the increment is moot. Harmless redundancy.
- If the challenge's `resource_metadata` ever names a URL on a different origin, `path` falls back to `new URL(named).pathname` but `getJson(path)` still fetches `base + path` (the reached origin). Not realistic for a self-describing server; noting for completeness.

### Verdict
**approved** (informational; a human still gives the explicit GitHub approval). Zero Critical findings — the shell-injection fix and the three hardening items called out in the description are all present in the diff, and the discovery logic is RFC-faithful. The suggestions above are non-blocking robustness/testability improvements.

tonychang04:
`e1c4814` adds `scripts/verify-handshake.mjs` — the bar this work was originally set against, which nothing in the repo actually checked.

`verify-deploy.mjs` proves the discovery documents are correct. That's necessary and not sufficient: every document can be right and the handshake still fail at the callback, the code exchange, the project lookup, or tool registration. Those are different questions and only one of them was covered.

This drives the actual `@modelcontextprotocol/sdk` client — the same one a real client uses — through discovery → dynamic registration → PKCE → code exchange → `initialize` → `tools/list`. A pass means a real client works. A failure **names the stage**, so "it doesn't work" becomes "it dies at the code exchange".

```
node scripts/verify-handshake.mjs https://mcp.insforge.dev
```

One human step, the browser sign-in, because that's what OAuth is. Everything either side is automatic and the callback is captured on loopback. Nothing is persisted — client information, tokens and code verifier live in memory for the run.

**Verified against a local server:** real discovery, real DCR with a `client_id` minted by the server under test, real PKCE authorize URL carrying the RFC 8707 `resource` parameter. Failure naming checked separately — pointed at a dead port it reports `FAILED at [discovery + registration]` and exits 1 rather than a stack trace.

### Worth flagging: lint caught a real bug in this file

```js
new Promise((res, rej) => ((resolve = res), (rej = rej)))   // rej = rej
```

`reject` was never assigned, so both the timeout path and the authorization-error path would have thrown a `TypeError` instead of failing cleanly — the two paths you only reach when something is *already* wrong. Caught by `no-self-assign`, which was only reachable because this commit also adds the globals `scripts/**` needs. Before that those files linted with `no-undef` noise drowning everything else.

**Not verified:** a run against a deployed instance. Prod is 139 days stale and I don't hold platform OAuth credentials, so the first real run will be the first real answer. That's the point of it being one command.

tonychang04:
`3674b37` — all four review findings on `verify-handshake.mjs`, and **the P1 was a real bug of mine on the success path.**

**P1 — reconnect after authorization.** After the browser step the script called `client.connect()` a second time on the same `Client`, which throws `Already connected to a transport`. It would have failed at exactly the moment it was supposed to succeed. I never saw it because without platform OAuth credentials every local run stops at the browser step, so the line past it had never executed. Now closes the first transport and client and reconnects with a fresh pair.

That one's worth naming plainly: **I shipped a verification tool whose success path had never run.** Same class as verifying everything except the thing being claimed.

**P2 — callback listener had no error handler.** A busy port crashed with an uncaught `EADDRINUSE` stack trace, in a script whose entire purpose is naming failures. Two parts to the fix: an `error` handler, *and* keeping the rejection handled, because the listener can fail long before anything awaits it — without that it still died on an unhandled rejection. Verified with 8765 occupied: previously a stack trace at listen time, now it proceeds and fails by name.

**P2 — no scheme check.** The script obtains a real access token and sends it to whatever URL it's given, so over plaintext that's a credential on the wire. Now refused unless the host is loopback, which keeps local testing working. Same rule already applied to issuers in `verify-deploy.mjs`.

```
$ node scripts/verify-handshake.mjs http://example.com
refusing to send a bearer token to http://example.com over http:
use https, or point this at a loopback address for local testing.
```

**P3 — unused eslint globals.** Dropped `Headers` and `clearTimeout`. Correct catch: I added `Headers` and then removed the code that needed it. Unused globals stop eslint catching a misspelling.

18/18, lint clean.

### Older findings on `verify-deploy.mjs`, status

- **Follow each endpoint document's own `authorization_servers`** — fixed (`d76c441`).
- **`Array.isArray` on the metadata list** — fixed (`d76c441`).
- **Workflow shell injection** — fixed; inputs go through `env:`, plus `persist-credentials: false` and minimal `permissions`.
- **No fetch timeout** — still open, and legitimate: a stalled endpoint can hang the scheduled job to the runner timeout. Not fixed here to keep this commit to the handshake findings; happy to add an `AbortSignal.timeout` pass if you'd like it in this PR rather than a follow-up.
- **302 unreachable on the authorize probe** — agreed, `/oauth/authorize` returns 400 before any redirect when parameters are missing, so accepting 302 weakens the assertion. Also deferred with the timeout change rather than mixed in here.

tonychang04:
`49149c7` — the last two outstanding items, both of which I'd deferred openly rather than dropped. **#73 now has nothing outstanding.**

**Every request carries a deadline.** `AbortSignal.timeout`, 10s default, settable via `VERIFY_TIMEOUT_MS`. This runs on a schedule and as a deploy gate, so a stalled endpoint would otherwise hang to the runner's own timeout and the whole check report would be lost — a server that never answers is a failure worth reporting, not a reason to report nothing.

Verified against a server that **accepts the connection and never responds**, which is the case a refused connection doesn't exercise (my first attempt used an unroutable address and returned in 45ms without touching the timeout path at all):

```
VERIFY_TIMEOUT_MS=2000 node scripts/verify-deploy.mjs http://127.0.0.1:8499

FAIL  GET /health is 200  — unreachable (The operation was aborted due to timeout)
http://127.0.0.1:8499 did not answer at all. Nothing else was checked.
real  0m2.026s
```

**The authorize probe now accepts 400 only.** The 302 branch was unreachable — the handler validates its parameters before it can redirect, so a bare probe can't legitimately produce one. Accepting it would quietly excuse a proxy or route change that started answering there, which is exactly the class of drift this script exists to catch.

Also adds the `AbortSignal` global for `scripts/`, without which the new code linted as `no-undef`.

18/18, lint clean. Note this moves the head again, so the review request I sent a few minutes ago is now stale — sending a fresh one.

jwfing (commented):
## Review: post-deploy discovery check (`chore(ops)`)

**Summary:** Read-only post-deploy verification tooling (`verify-deploy.mjs` + a real-client `verify-handshake.mjs`, plus a scheduled/dispatch workflow and eslint/package wiring) — the discovery assertions match the server's actual behavior, the SDK usage is correct, and the previously-flagged shell injection is properly fixed; the remaining notes are non-blocking hardening.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/` at all), so this was assessed against the PR description, the split-out context from #69, and the actual runtime code the scripts assert against (`src/http/server.ts`, `src/http/session-manager.ts`).

Verified against the code on the head commit:
- `/health` returns `{ version, sessions: { activeSessions, memorySessionCount }, ... }` (`src/http/server.ts:131-146`, `src/http/session-manager.ts:338-364`) — matches what `verify-deploy.mjs` reads.
- `/oauth/authorize` returns **500** when `isOAuthConfigured()` is false and **400** on missing params *after* the credential check (`src/http/server.ts:244-256`) — exactly the 500-vs-400 discriminator the script relies on.
- MCP SDK OAuth flow (`@modelcontextprotocol/sdk@^1.27.1`): confirmed via upstream docs that `StreamableHTTPClientTransport` accepts an `OAuthClientProvider` as `authProvider`, and the documented interactive pattern is "after `UnauthorizedError`, redirect, then call `finishAuth(code)` before reconnecting" — which `verify-handshake.mjs` implements, including the fresh-client/transport reconnect. Import subpaths are correct for the 1.x line. No stale or hallucinated API.

---

### Critical
(none)

---

### Suggestion

**Functionality — `verify-deploy.mjs:27`: invalid base URL crashes with a raw stack trace.**
`const baseOrigin = new URL(base).origin;` runs at module top with no guard. Invoked as `verify-deploy.mjs mcp.insforge.dev` (no scheme — a realistic operator mistake), `new URL()` throws an unhandled `TypeError` before any friendly output. This is the exact failure mode the script is designed to avoid ("reads as a broken script rather than a down server"), and `verify-handshake.mjs:23-30` already guards its `new URL(target)` in try/catch. Mirror that guard here.

**Functionality — `verify-deploy.mjs:140`: cross-host `resource_metadata` is fetched from the target, not the named host.**
```js
const path = named.startsWith(base) ? named.slice(base.length) : new URL(named).pathname;
const doc = await getJson(path); // getJson → fetch(base + path)
```
In the else branch (a `resource_metadata` URL on a *different* origin) the code keeps only the pathname and re-fetches it from `base`, so the assertion "the named document … is 200" actually checks the target's copy of that path rather than the document a real client would fetch at the absolute `named` URL. In practice `resource_metadata` is same-origin so this rarely fires, but if it ever points off-origin the check silently validates the wrong host. Fetching `named` directly (with the same origin-scope logic used for authorization servers) would be more faithful.

**Software engineering — no automated tests for the pure RFC-derivation helpers.**
`asMetadataUrl` (`verify-deploy.mjs:208`) and `parseIssuer` (`verify-deploy.mjs:215`) are pure functions encoding the RFC 8414 §3.1/§3.3 metadata-URL and issuer rules — and the PR body itself notes "every implementation looked at while writing this got that rule wrong in one direction or another." These are trivially unit-testable (path-suffix insertion, trailing-slash handling, https/loopback acceptance, unparseable rejection) and the repo's convention is TDD (vitest). A small table-driven test would lock in the exact behavior these assertions depend on. Acknowledged that end-to-end network assertions in these scripts aren't practically unit-testable — this note is only about the extractable pure logic.

---

### Information

- **`verify-deploy.mjs:195`: the `=== 302` branch is effectively unreachable.** `getJson` uses `fetch` with default `redirect: 'follow'`, so a real 302 from `/oauth/authorize` would be followed and `res.status` would be the *final* status, never 302. The bare no-param request returns 400 anyway, so the check is correct in practice — the `|| authorize.status === 302` is just dead. Harmless; drop it or note why it's there.
- **Security — no concerns.** The workflow passes both dispatch inputs via `env:` and quotes them as argv (`.github/workflows/verify-deploy.yml:55-62`), closing the earlier shell-injection; `permissions: contents: read` and `persist-credentials: false` are set; both scripts refuse to send/derive over non-HTTPS (loopback exempted); the tooling registers nothing and writes nothing. No secrets are read or logged.
- **Performance — no concerns.** One-shot sequential fetches in a diagnostic script; no hot path, no added server-side work (the `/health` `SCAN` is pre-existing).
- **eslint config (`eslint.config.js`):** the `scripts/**/*.mjs` globals block covers every runtime global the two scripts use (`process`, `console`, `fetch`, `URL`, `setTimeout`); `http` is imported and the rest are ES built-ins. Fine.

---

### Verdict
**approved** (informational — zero Critical findings; the GitHub green-check remains a separate human action). The three Suggestions are worth a quick pass but none block merge; the tooling does what the PR claims and its assertions are grounded in the real server behavior.

jwfing (commented):
## Review: post-deploy discovery check (#73)

**Summary:** A clean, read-only, purely-additive ops tool — two Node ESM scripts plus a scheduled workflow — that encodes the OAuth/MCP discovery-chain checks by hand into a repeatable gate; no Critical findings.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/`) directory, so this is assessed against the PR description, the split-out history from #69, and the actual server contract in `src/http/server.ts`. I verified the scripts' assumptions against the runtime:
- `/health` shape matches (`server.ts:132-145` → `version`, `sessions.{activeSessions,memorySessionCount}` from `session-manager.ts:361-362`). ✅
- `/oauth/authorize` returns **500** when `isOAuthConfigured()` is false and **400** `invalid_request` when configured-but-missing-params — and the credentials check runs *before* param validation (`server.ts:241-256`), so the "500 → creds missing, 400 → reachable+validating" logic is accurate. ✅
- The `/mcp` 401 does not currently emit `WWW-Authenticate` (`server.ts:762-796`), which is exactly the production gap the script is designed to catch (RED against prod as the PR states) — the assertion is intentional, not a false assumption. ✅
- MCP SDK auth usage verified against `@modelcontextprotocol/typescript-sdk` v1.29.0 docs (context7): the `OAuthClientProvider` shape, `transport.finishAuth(code)`, and the catch-`UnauthorizedError` → `redirectToAuthorization` → `finishAuth` → reconnect-with-fresh-transport pattern all match the documented flow. No stale/hallucinated API. ✅

### Findings

#### Critical
(none)

#### Suggestion

- **[functionality] Cross-origin challenge document is fetched against the wrong host** — `scripts/verify-deploy.mjs:151`: when `resource_metadata` names a URL on a *different* origin, `new URL(named).pathname` drops the origin and the subsequent `getJson(path)` re-prepends `base`, so the "named document" check probes `base` + path rather than the host the challenge actually named. Same-origin (the normal case) is handled correctly via the `startsWith(base)` branch, so blast radius is low, but for a delegated resource this reports against the wrong host. Consider fetching the absolute `named` URL directly.

- **[functionality] "OAuth credentials are configured" can read green on non-500 errors** — `scripts/verify-deploy.mjs:201-202`: `authorize.status !== 500` is broadly true for an unreachable endpoint (status is a `unreachable (...)` string) or a route that 404s, so that specific line can PASS misleadingly. The paired `status === 400` check at `:209-210` catches most of these in aggregate, but the credentials line itself would benefit from asserting the positive (`=== 400`) or explicitly handling the unreachable/other-status case.

- **[software engineering] No automated tests for the new scripts** — correctness rests entirely on the manual runs documented in the PR body (8/15 prod, 21/21 correct server, 13/18 insta). The repo has no precedent for unit-testing ops scripts and these are read-only and out of the runtime path, so this is non-blocking. That said, the pure helpers (`asMetadataUrl`, `parseIssuer`, the `resource_metadata` regex, and the `derivedFrom` derivation) are cheap to cover with vitest and would guard the RFC 8414/9728 path-insertion logic against regression — which the PR itself notes "every implementation got wrong in one direction or another."

#### Information

- **[security]** Hardening confirmed good: workflow inputs pass through `env:` (never interpolated into `run:`), `permissions: contents: read`, `persist-credentials: false`; both scripts refuse to send/expose over non-HTTPS unless loopback; scripts register/write nothing. The previously-flagged shell injection is fully addressed. No secrets logged.
- **[performance]** No concerns — a bounded sequence of `fetch`es, each with an `AbortSignal.timeout` deadline (`VERIFY_TIMEOUT_MS`, default 10s); no loops over unbounded input, no runtime hot path touched.
- **[software engineering]** `scripts/verify-handshake.mjs:89-91`: the `no callback` timeout timer is never `clearTimeout`'d on the success path; it's `unref()`'d so it won't keep the process alive — purely cosmetic.
- **[software engineering]** The eslint block (`eslint.config.js`) is correct and necessary: `.mjs` files are linted as modules without the TS runtime globals in scope, so declaring `process/console/fetch/URL/AbortSignal/setTimeout` prevents `no-undef`. `fetch`/`AbortSignal` are only used by `verify-deploy`; harmless for `verify-handshake` (which imports `node:http`).
- **[functionality]** `verify:deploy` is wired as an npm script but `verify:handshake` is not — intentional given the handshake's human-in-the-loop browser step, worth a one-line note in the script header for discoverability.

### Verdict
**approved** (informational) — zero Critical findings. Posted as a COMMENT; the three Suggestions are worth a look but none block merge. Human approval via the GitHub approve flow is still a separate step.

tonychang04:
`7dad9c4` — @infra-bot found a false green in this gate against the live Manufact deployment, and it's the same class as the off-target scoring @qa-bot found: **a response proves something answered, not that the thing under test answered.**

Manufact's gateway returns `{"status":"healthy","timestamp":...}` with a 200 whether or not a container is running behind it. So the `/health` assertion was green against a deployment that has *never once started* — production 404s `/mcp` right now and this script would have reported the health check passing.

**Two changes:**

- **Assert the body names us.** Ours says `server: "insforge-mcp"`; a placeholder can't, because it doesn't know what it's standing in for. That's the check that distinguishes our container from anything else serving a 200 on that path.
- **Report the version unconditionally.** It was only asserted when an expected version was passed, so omitting the argument silently dropped the check that catches a container serving the right name at the wrong build — the failure that let production sit 139 days stale. With an argument it still compares; without one it at least requires a version to be present.

**Measured both directions:**

```
live Manufact gateway
  PASS  GET /health is 200
  FAIL  /health is our server, not a platform placeholder
        — server=undefined body={"status":"healthy","timestamp":"..."}
  FAIL  /health reports a version — got undefined
  2/18 checks passed

real local server
  PASS  GET /health is 200
  PASS  /health is our server, not a platform placeholder
  PASS  /health reports a version
```

The `GET /health is 200` line staying green in the first block is deliberate — it's true, and the point is that it isn't sufficient. The two lines under it are what the gate now rests on.

Head has moved, so the review request from earlier is stale again; sending a fresh one.

jwfing (commented):
## Review: post-deploy discovery check (split out of #69)

**Summary:** A read-only, dependency-free post-deploy verification tool (`verify-deploy.mjs`) plus a real-client OAuth handshake check (`verify-handshake.mjs`) and a scheduled/dispatch workflow — well-reasoned, security-conscious, and correct against the server it probes; no blocking issues.

### Requirements context
No matching spec/plan found — this repo has no `docs/superpowers/` (or `docs/`) directory, so I assessed against the PR description, the linked context (#69), and the actual server behavior in the workspace. I verified the script's assumptions against the live handlers rather than trusting the narrative:

- `/health` shape matches: `src/http/server.ts:135-145` returns `server: 'insforge-mcp'`, `version`, and `sessions: { activeSessions, memorySessionCount }` (`src/http/session-manager.ts:360-363`) — exactly what the script asserts.
- The 500-vs-400 discrimination is real and load-bearing: `isOAuthConfigured()` → `500 server_error` at `src/http/server.ts:244-249` runs *before* the missing-parameter → `400 invalid_request` check at `:251-256`. A bare `/oauth/authorize` probe genuinely separates "credentials missing" from "configured".
- The `@modelcontextprotocol/sdk` OAuth client surface used by the handshake (`StreamableHTTPClientTransport({ authProvider })`, the `OAuthClientProvider` method set, `transport.finishAuth(code)`) matches the SDK's real API for the pinned `^1.27.1` (verified against SDK docs) — no stale or hallucinated calls.

The script deliberately encodes the **RFC-correct target state** (per-path `oauth-protected-resource` documents, `resource` == derived identifier), which is stricter than what production currently serves — consistent with the PR's stated intent that this run RED until #69's fix is deployed.

---

### Findings

#### Critical
(none) — this is a read-only ops tool; it registers nothing, writes nothing, and adds no runtime code path.

#### Suggestion

- **Functionality — false green on an unreachable authorize endpoint** (`scripts/verify-deploy.mjs:210-215`). `getJson('/oauth/authorize')` returns `{ status: 'unreachable (...)' }` (a *string*) on network failure. The next line, `authorize.status !== 500`, then evaluates `'unreachable…' !== 500` → `true`, so **"OAuth credentials are configured" prints PASS** even though nothing answered. The sibling check `authorize.status === 400` (`:216-223`) does fail, so the aggregate gate still exits non-zero — blast radius is limited to one misleading report line. But given the script's explicit no-false-greens ethos, consider guarding on a numeric status first (e.g. treat a non-number `status` as a target failure for this assertion, mirroring the early-exit you already do for `/health` at `:110-118`).

- **Software engineering — no tests for the RFC-encoding helpers** (`scripts/verify-deploy.mjs:233-256`). `asMetadataUrl()` (well-known segment insertion) and `parseIssuer()` (https enforcement, loopback exemption, unparseable/non-string rejection) are pure, network-free functions encoding exactly the RFC 8414/9728 rules the PR says "every implementation got wrong in one direction or another." The repo gate already runs vitest; a small unit test locking in these two would be cheap and would guard the highest-risk logic in the diff against future regressions. Not blocking (the script is itself a test harness), but this is the one spot where a close-call bug could slip in silently.

#### Information

- **Functionality — cross-origin `resource_metadata` follow** (`scripts/verify-deploy.mjs:162`). When the challenge's `resource_metadata` URL doesn't start with `base`, `path` falls back to `new URL(named).pathname` and is then fetched as `base + path` (via `getJson`), which would hit the *wrong host*. Harmless today (the challenge points at the same origin), but if delegated resource metadata ever lands on another origin this check would probe the wrong server. Consider fetching the absolute `named` URL directly.

- **Functionality — intentional strictness vs. current server.** The origin-document expectation `resource === base` and the derived `/mcp` + `/sse` documents (`:174-197`) are stricter than the single document the server serves today (`src/http/server.ts:169-178` returns one doc with `resource: ${baseUrl}/mcp`). Flagging for reviewers only: this is by design — the script encodes the post-#69 RFC-conformant target, so a RED run against current production is expected, not a script bug.

- **Security — positive notes (no changes weakened).** The MAJOR shell-injection finding is correctly resolved: dispatch inputs flow through `env:` and are read as `"$DEPLOY_URL"`/`"$EXPECTED_VERSION"`, never interpolated into `run:` (`.github/workflows/verify-deploy.yml:57-62`). Workflow runs with `permissions: contents: read` and `persist-credentials: false`. `verify-handshake.mjs:23-33` refuses to send a bearer token over non-HTTPS (loopback exempt), and `parseIssuer` rejects non-HTTPS issuers. Every request carries an `AbortSignal.timeout` deadline (`:35`), so a stalled endpoint fails loudly instead of hanging CI. No secrets are read, logged, or returned.

- **Performance.** Sequential fetches only, no DB/hot-path work, all bounded by per-request deadlines — appropriate for a gate script. No concerns.

- **Software engineering — eslint globals** (`eslint.config.js:10-22`). The added `scripts/**/*.mjs` globals block cover both scripts' runtime dependencies (`fetch`, `URL`, `AbortSignal`, `setTimeout`, `process`, `console`); `verify-deploy.mjs` imports nothing and `verify-handshake.mjs` only imports `node:http` + the SDK, so the workflow correctly needs no `npm install`. Good.

---

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings: the tool is read-only, its assumptions check out against the actual handlers, and its security posture is solid. The two Suggestions (unreachable-authorize false green; unit tests for the RFC helpers) are worth a look but are non-blocking.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review: post-deploy discovery check (#73)

**Summary:** A read-only, well-documented post-deploy verification tool (`verify-deploy.mjs` + `verify-handshake.mjs`) whose server-side assumptions I verified against the actual source and whose prior shell-injection finding is confirmed fixed — no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or any `docs/`) directory. Assessed against the PR description, the split-out context from #69, and the actual server implementation in `src/http/server.ts`.

I cross-checked the script's assertions against the deployed server's real behavior:
- `/health` shape — `src/http/server.ts:135-145` returns `server: 'insforge-mcp'`, `version`, and `sessions` with `activeSessions`/`memorySessionCount` (`session-manager.ts:339-362`). Matches the script's expectations exactly. ✅
- `/oauth/authorize` 500-vs-400 discriminator — `src/http/server.ts:244-256`: `500 server_error` when `!isOAuthConfigured()`, `400 invalid_request` on missing params. The script's "credentials configured" (`status !== 500`) and "reachable and validating" (`status === 400`) checks are accurate. ✅
- The 401/`WWW-Authenticate` and per-path protected-resource assertions correctly encode RFC 9728 §3.1/§3.3. The current server emits no `WWW-Authenticate` header and serves a single `/.well-known/oauth-protected-resource` with `resource: base/mcp` (`server.ts:169-178`) — i.e. the script *correctly* reports the very drift the PR says it catches (8/15 today). This is the tool doing its job, not a script bug. ✅
- `verify-handshake.mjs` SDK usage checked against MCP TypeScript SDK docs (context7): `finishAuth(code)` with a string arg is correct for the pinned `@modelcontextprotocol/sdk ^1.27.1` (the v2 line later changed it to take the full callback `params`); the "reconnect on a fresh transport after auth" pattern matches the documented guidance. ✅

### Findings

#### Critical
(none) — the shell-injection fix from #69 is present and correct: dispatch inputs flow through `env:` (`verify-deploy.yml:57-62`), plus `permissions: contents: read` and `persist-credentials: false`. Read-only client, no secrets logged, no new dependencies.

#### Suggestion
- **Software engineering / test coverage** — `scripts/verify-deploy.mjs` contains deterministic, pure helpers that are unit-testable in isolation: `parseIssuer` (`verify-deploy.mjs:240-260`), `asMetadataUrl` (`verify-deploy.mjs:230-234`), and the self-identity/loopback scoping logic. The repo convention is TDD, and no tests accompany these. Non-blocking: the script's real value is integration-level against live infra, and the PR documents empirical discrimination (8/15 prod, 21/21 correct, 13/18 instacloud) as functional evidence — but extracting the URL/issuer helpers into a small `*.test.ts` would lock in the RFC-parsing edge cases (non-HTTPS rejection, loopback exemption, unparseable issuer).
- **Functionality / robustness** — `verify-handshake.mjs:171` detects the redirect-needed case via `error?.constructor?.name !== 'UnauthorizedError'` (string compare) rather than `instanceof UnauthorizedError`. The SDK exports `UnauthorizedError` from `@modelcontextprotocol/sdk/client/auth.js`; `instanceof` is more robust — a constructor-name compare breaks under bundling/minification or a subclass. It runs fine as a plain `node` ESM script today, so this is a resilience nicety, not a defect.

#### Information
- **Security / real-client fidelity** — the handshake's in-memory `provider` omits an OAuth `state()` CSRF binding and its callback (`verify-handshake.mjs:78-92`) doesn't validate `state`. The MCP SDK docs note the SDK does *not* validate `state` and the client should. This is acceptable for an operator-run loopback verification tool (no victim, operator initiates and completes on `127.0.0.1`), but worth noting given the script's stated bar — "a pass here means a real client works" — since a shipped client should validate `state`.
- **Functionality / discoverability** — `verify-handshake.mjs` is not wired into `package.json` scripts (only `verify:deploy` is). A `verify:handshake` alias would make it as discoverable as its companion.
- **Portability** — `verify-deploy.mjs` uses global `fetch` and `AbortSignal.timeout` (Node ≥18 / ≥17.3). CI is safe (`verify-deploy.yml` pins `lts/*`), but `npm run verify:deploy` on an older local Node would throw an opaque error; an `engines` hint or a friendlier guard is optional. Separately, the `finishAuth(code)` string arg is correct only for the pinned `^1.27.1` — flagging so a future SDK major bump revisits it.

#### Performance
No concerns — this is a scheduled/on-demand ops script with sequential fetches, each bounded by a per-request deadline (`AbortSignal.timeout`, default 10s, `VERIFY_TIMEOUT_MS` override). No hot path, no unbounded loops, no server-side load beyond a handful of GETs and one unauthenticated MCP POST.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the shell-injection hardening and RFC/server assumptions all check out. The Suggestions (unit tests for the pure helpers, `instanceof` for the SDK error) and Information notes are non-blocking polish the author can take or leave.

PR #74, opened 2026-07-29
Stop client registrations expiring while they are in use

tonychang04:
Fixes the 30-day registration expiry @qa-bot found and @manager scoped. Own PR rather than #71, since it shares nothing with the session leak.

## The bug

`server.ts` writes the registration with a 30-day TTL at `/oauth/register` and the read in `/oauth/authorize` never refreshes it. So this was never "stale after 30 days of disuse" — **every client stops working exactly 30 days after it registered, however heavily it has been used in between.** Some connected clients are already past that line, and the failure is silent from their side.

## Two changes, and nothing else

**1. Refresh the TTL once the registration has been read and validated.** That turns a fixed lifetime into an idle timeout for any client that keeps *authorizing*.

> **Correction (@qa-bot, measured):** I first wrote "an actively-used client stops expiring at all." That is wider than the change. Registrations slide only at `/oauth/authorize`; token bindings slide on every authenticated request. So the ordinary user — signs in once, then only calls tools — never returns to authorize, and their registration still dies 30 days after it was written. This PR strictly improves matters and the recovery path below is what they land on; it just does not cover that population. #75 does. A failed refresh is logged and swallowed — the worst case is the record keeping its old expiry, and the login itself must not fail on it.

**2. Answer the already-dead clients with something a person can act on.** Nothing here can recover automatically, and it's worth being precise about why:

- The MCP SDK only re-registers on an `InvalidClientError` from the **token** endpoint or from registration — `client/auth.js:152`, `invalidateCredentials('all')`.
- `/oauth/authorize` is a **browser navigation**. The SDK builds the URL and hands it to the browser; it never fetches or parses the response.
- Reaching `/oauth/token` requires an authorization code, which requires `/oauth/authorize` to have succeeded. So a dead registration can never produce the error that would heal it.

The only party who sees this response is the person looking at the tab, and a raw OAuth JSON body tells them nothing. They now get a page saying the connection needs adding again, naming the command. Programmatic callers get the identical JSON; a request with no `Accept` header gets JSON too.

### Deliberately not included

- **A `WWW-Authenticate` hint on the 400.** It has the same shape as #69's 401 but no consumer — per the above, the SDK never reads this response. A spec-shaped header nothing parses is worse than nothing, because it reads as a fix.
- **Validating `client_id` at `/oauth/token`.** It's the endpoint the SDK *does* watch, but the flow dies at authorize before ever reaching it, so it would add a rejection path that never fires in this scenario.
- **Any registration-lifecycle redesign.** Registrations become a signed, stateless value in the platform-OAuth rewrite. Both changes here sit on routes that rewrite deletes outright, so neither is work to un-build.

## Measured, both sides

Real built server, real Redis, real `/oauth/register` + `/oauth/authorize`. Pre-fix column is the same script against `master`.

| | pre-fix | post-fix |
|---|---|---|
| TTL at registration | 2592000 | 2592000 |
| TTL after aging the record to 120s | 120 | 120 |
| **TTL after a successful `/oauth/authorize`** | **120** — unchanged, still counting down | **2592000** — refreshed |
| authorize result | 302 | 302 |
| browser sees on a dead registration | raw JSON `invalid_client` | an actionable page, `400 text/html` |
| program sees on a dead registration | same JSON | same JSON, byte-identical |
| no `Accept` header | JSON | JSON |

27/27 tests green, 4 new on the template — including that every field is escaped, so the page stays safe if someone later reflects a `client_id` or `redirect_uri` into it.

---
## Summary by cubic
Stops active clients from dying after 30 days by refreshing the registration TTL on successful authorize, and shows an actionable HTML page for already-expired registrations so users know how to fix it.

- **Bug Fixes**
  - Refresh the client registration TTL on authorize, turning the fixed 30-day lifetime into an idle timeout. Failures are logged and ignored so login isn’t blocked.

- **New Features**
  - When an unknown/expired client_id hits authorize and the request prefers HTML, return a clear page with instructions (e.g., run `npx @insforge/install`). Programmatic callers still get the same JSON error. Added tests to ensure all template fields are safely escaped.

<sup>Written for commit a4ecce33c3f69a13e253ac4ff378096fc27013b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/74?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -23,6 +23,7 @@ import {
   validateConfig,
 } from './config.js';
 import { renderProjectSelectionPage } from './templates/project-selection.js';
+import { renderOAuthErrorPage } from './templates/oauth-error.js';
 import { getAnalyticsService, extractClientInfo } from './analytics.js';
 import { PACKAGE_VERSION } from '../shared/version.js';
 
@@ -77,6 +78,28 @@ function isInitializeRequest(body: unknown): boolean {
   }
 
   return false;
+}
+
+/**
+ * How long a client registration survives without being used.
+ *
+ * Refreshed on every successful authorize (see below), so this is an idle
+ * timeout rather than a hard lifetime. It used to be neither: the value was
+ * written once at registration and never touched again, so every client
+ * stopped working exactly 30 days after it registered no matter how heavily
+ * it was being used.
+ */
+const CLIENT_REGISTRATION_TTL = 30 * 24 * 60 * 60;
+
+/**
+ * Whether this request is a browser navigation rather than a program's fetch.
+ *
+ * The authorize endpoint is reached by opening a URL in the user's browser —
+ * the MCP client never reads the response itself — so an error here has to be
+ * written for a person. Everything else still gets the OAuth JSON body.
+ */
+function prefersHtml(req: Request): boolean {
+  return req.accepts(['json', 'html']) === 'html';
 }
 
 /**
@@ -217,7 +240,7 @@ app.post(OAUTH_ENDPOINTS.register, async (req: Request, res: Response) => {
 
   await redis.setex(
     `mcp:oauth:client:${clientId}`,
-    30 * 24 * 60 * 60,
+    CLIENT_REGISTRATION_TTL,
     JSON.stringify(clientData)
   );
 
@@ -267,8 +290,26 @@ app.get(OAUTH_ENDPOINTS.authorize, async (req: Request, res: Response) => {
 
   // Validate client_id and redirect_uri
   const redis = getRedisClient();
-  const clientDataStr = await redis.get(`mcp:oauth:client:${client_id}`);
+  const clientKey = `mcp:oauth:client:${client_id}`;
+  const clientDataStr = await redis.get(clientKey);
   if (!clientDataStr) {
+    // Nothing here can recover automatically. The MCP SDK only re-registers on
+    // an invalid_client from the token endpoint or from registration, and it
+    // never sees this response — the browser does. So the person holding the
+    // tab is the only one who can act, and they need to be told how.
+    console.log(`[OAuth] Unknown client_id at authorize: ${client_id}`);
+    if (prefersHtml(req)) {
+      return res.status(400).type('html').send(
+        renderOAuthErrorPage({
+          heading: 'This connection needs to be set up again',
+          message:
+            'The app you are connecting from is not registered with this server any more. ' +
+            'Nothing is wrong with your account and no data was lost — remove the InsForge MCP ' +
+            'server from your client and add it back, and this will complete normally.',
+          action: 'npx @insforge/install',
+        })
+      );
+    }
     return res.status(400).json({
       error: 'invalid_client',
       error_description: 'Unknown client_id. Register client first via /oauth/register.',
@@ -309,6 +350,16 @@ app.get(OAUTH_ENDPOINTS.authorize, async (req: Request, res: Response) => {
       error: 'invalid_request',
       error_description: 'redirect_uri does not match any registered redirect URIs for this client.',
     });
+  }
+
+  // The registration has been used successfully, so restart its idle clock.
+  // Without this the record expires on a fixed schedule from the moment it was
+  // written, taking working clients down with it. Failing to refresh must not
+  // fail the login — the worst case is the record expiring on its old schedule.
+  try {
+    await redis.expire(clientKey, CLIENT_REGISTRATION_TTL);
+  } catch (error) {
+    console.error(`[OAuth] Could not refresh registration TTL for ${client_id}:`, error);
   }
 
   try {
--- a/src/http/templates/oauth-error.test.ts
+++ b/src/http/templates/oauth-error.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect } from 'vitest';
+import { renderOAuthErrorPage } from './oauth-error.js';
+
+describe('renderOAuthErrorPage', () => {
+  it('renders the heading, message and action', () => {
+    const html = renderOAuthErrorPage({
+      heading: 'Set this up again',
+      message: 'The app is not registered any more.',
+      action: 'npx @insforge/install',
+    });
+
+    expect(html).toContain('<h1>Set this up again</h1>');
+    expect(html).toContain('The app is not registered any more.');
+    expect(html).toContain('<code>npx @insforge/install</code>');
+    expect(html.startsWith('<!DOCTYPE html>')).toBe(true);
+  });
+
+  it('omits the code block when there is no action', () => {
+    const html = renderOAuthErrorPage({ heading: 'Nope', message: 'No action here.' });
+    expect(html).not.toContain('<code>');
+  });
+
+  it('escapes every field, so a reflected value cannot become markup', () => {
+    // Nothing untrusted is passed today, but this page sits on an unauthenticated
+    // endpoint reached straight from a URL — the escaping is what keeps it safe
+    // if someone later renders the client_id or redirect_uri into it.
+    const html = renderOAuthErrorPage({
+      heading: '<script>alert(1)</script>',
+      message: '"quoted" & <b>bold</b>',
+      action: "';DROP--<img src=x onerror=alert(1)>",
+    });
+
+    expect(html).not.toContain('<script>');
+    expect(html).not.toContain('<img src=x');
+    expect(html).not.toContain('<b>bold</b>');
+    expect(html).toContain('&lt;script&gt;alert(1)&lt;/script&gt;');
+    expect(html).toContain('&quot;quoted&quot; &amp; &lt;b&gt;bold&lt;/b&gt;');
+  });
+
+  it('escapes the heading in the title as well as the body', () => {
+    const html = renderOAuthErrorPage({ heading: '</title><script>x</script>', message: 'm' });
+    expect(html).not.toContain('</title><script>');
+  });
+});
--- a/src/http/templates/oauth-error.ts
+++ b/src/http/templates/oauth-error.ts
@@ -1,0 +1,107 @@
+/**
+ * OAuth Error Page HTML Template
+ *
+ * The OAuth authorize endpoint is reached by a browser navigation, not by the
+ * MCP client's own fetch — so when it fails, the only party who can see the
+ * response is the person staring at the tab. A JSON error body tells them
+ * nothing they can act on. This page tells them what to do instead.
+ */
+
+export interface OAuthErrorPageOptions {
+  heading: string;
+  message: string;
+  /** Optional literal the user should run or check. Rendered as code. */
+  action?: string;
+}
+
+export function renderOAuthErrorPage(options: OAuthErrorPageOptions): string {
+  const { heading, message, action } = options;
+
+  return `<!DOCTYPE html>
+<html lang="en">
+<head>
+  <title>${escapeHtml(heading)} - InsForge MCP</title>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+  <link href="https://fonts.googleapis.com/css2?family=Manrope:wght@400;500;600;700&display=swap" rel="stylesheet">
+  <style>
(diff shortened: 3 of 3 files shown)


jwfing (commented):
## Summary

Turns the 30-day client-registration lifetime into an idle timeout by refreshing the Redis TTL on a successful `/oauth/authorize`, and replaces the raw `invalid_client` JSON with an actionable HTML page for the browser that actually sees it — a tightly scoped, well-reasoned fix.

## Requirements context

No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the linked bug scope (`@qa-bot` / `@manager`), and the surrounding code in `src/http/server.ts`. The PR's "deliberately not included" section is unusually clear about intent and scope, which made this easy to check.

The diff matches its stated scope exactly: TTL refresh + HTML error page, nothing else. Good scope discipline.

## Findings

### Critical

(none)

### Suggestion

**1. Functionality — the fix's "stops expiring at all" claim has a boundary hole because the registration TTL and the access-token lifetime are the same 30 days.** `src/http/server.ts:92,360,643`

The registration TTL (`CLIENT_REGISTRATION_TTL`, `server.ts:92`) and the issued access token (`expires_in: 30 * 24 * 60 * 60`, `server.ts:643`) are the identical `30d` constant, and `refresh_token` is unsupported (`server.ts:658-667`). The TTL is only refreshed on `/authorize` (`server.ts:360`).

So the two clocks are coupled: a client authorizes at `T0` (registration → `T0+30d`), receives a token a moment later that expires at `≈T0+30d+ε`, rides that single token to expiry, and only *then* re-authorizes — at which point the registration has already lapsed. That re-authorize lands on the dead-registration path, i.e. exactly the symptom the PR wants to eliminate for an active client.

This is strictly better than `master` (where every client died unconditionally at 30 days), and clients that re-run the flow more often than every 30 days (restarts, reconnections) are fully covered — so it's not blocking. But to robustly deliver the "actively-used client stops expiring at all" claim, the registration TTL should exceed the token lifetime by a margin (e.g. `60d`, or `tokenTTL + slack`) so a re-authorize triggered by token expiry always lands inside a live registration. Worth a line in the PR body noting this residual case if you'd rather not change the constant.

**2. Software engineering — the actual bug fix has no automated regression test.** `src/http/server.ts:295-317,355-363`

The new tests cover the pure template (`oauth-error.test.ts`) thoroughly, including escaping — nice. But the two *behavior* changes in `server.ts` are untested: (a) the TTL is refreshed on a successful authorize, and (b) `prefersHtml` routes browsers to HTML and programs/`Accept`-less requests to JSON. Per the repo's TDD convention, a bug fix should carry a regression test that fails on `master` and passes here; right now that's asserted only via the manual before/after measurement in the PR body.

I recognize `server.ts` has no existing unit-test harness (endpoints are exercised only by `src/integration/real-project.test.ts`), so this is more than a trivial add — hence Suggestion, not blocking. A focused test with a mocked Redis around the authorize handler (asserting `expire` is called with the key + TTL, and that content negotiation branches correctly) would lock in the core of the fix.

### Information

**3. `escapeHtml` is now duplicated.** `src/http/templates/oauth-error.ts:100` and `src/http/templates/project-selection.ts:341` define byte-identical helpers. A shared `templates/escape-html.ts` util would DRY this up — though given the PR's (correct) scope discipline, leaving it for a dedicated cleanup is reasonable.

**4. User-controlled `client_id` is echoed into logs.** `src/http/server.ts:300` — `console.log(...Unknown client_id at authorize: ${client_id})`. `client_id` is an unauthenticated query param, so a crafted value with newlines could inject lines into logs. This matches the pre-existing logging style elsewhere in the file and is low impact; noting only for awareness.

## Dimension coverage

- **Security** — the HTML page escapes every interpolated field (`&`, `<`, `>`, `"`, `'`), including inside `<title>`, and the tests prove reflected markup can't break out; it exposes no secrets and doesn't touch auth. The only note is the informational log-injection item above. No auth checks weakened; no new dependencies. Solid.
- **Performance** — one extra O(1) `redis.expire` per authorize; no N+1, no blocking work on the event loop, no per-request allocations of note. No findings.
- **Functionality** — see Suggestion 1; otherwise the content negotiation is correct (`req.accepts(['json','html'])` yields `json` for `*/*` and missing-`Accept` requests, `html` only for real browser navigations), the TTL refresh is correctly placed after full validation, is failure-tolerant, and can't fail the login.
- **Software engineering** — see Suggestion 2 and Information 3; code style, naming, comments, and error handling all match the surrounding file.

## Verdict

**approved** (informational — no Critical findings). The two Suggestions are worth considering before or shortly after merge, but neither blocks. A human still gives the explicit GitHub approval.

jwfing (commented):
## Review — Stop client registrations expiring while they are in use

**Summary:** A tight, well-scoped fix that converts the 30-day client-registration TTL from a hard lifetime into an idle timeout (refreshed on validated authorize) and replaces the browser-facing JSON error for a dead registration with an actionable, XSS-safe HTML page — I found no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and there is no matching spec/plan — assessing against the PR description, the referenced bug (`@qa-bot`/`@manager` scoping), and the surrounding code alone. The PR body's reasoning and its before/after measurement table were used as the statement of intent.

I verified the key claims against the actual code and libraries:
- The TTL refresh (`redis.expire`) is placed *after* `client_id` lookup, JSON parse, field validation, and `redirect_uri` match (`src/http/server.ts:355-363`) — so it only fires on a genuinely valid authorize, matching "read and validated." ✅
- The registration key `mcp:oauth:client:*` is only written at `/oauth/register` and read at `/oauth/authorize`; `/oauth/token` does **not** read it (uses `client_id` only for analytics logging, `server.ts:611+`). This confirms the PR's argument that the flow can only die at authorize and that refreshing there is the right hook. ✅
- `req.accepts(['json','html'])` returns `'json'` when the request has no `Accept` header (and for `*/*`), per the `accepts`/`negotiator` internals — so `prefersHtml` (`server.ts:100-102`) correctly serves JSON to programmatic callers and HTML only to browsers, exactly as the table claims. ✅
- `escapeHtml` escapes `&` first, then `< > " '` (`src/http/templates/oauth-error.ts:99-105`), and both interpolation sites (`<title>` and `<h1>`/`<p>`/`<code>`) route through it. Escaping order is correct and the tests exercise the `</title>`-breakout case. ✅

### Findings

#### Critical
(none)

#### Suggestion
- **[software engineering] The core behavioral fix has no automated regression test.** `src/http/templates/oauth-error.test.ts` thoroughly covers the template (rendering, action omission, full XSS escaping — nice), but the two server-side behaviors that *are* the bug fix — the `redis.expire` TTL refresh (`server.ts:355-363`) and the `prefersHtml` content-negotiation branch (`server.ts:295-317`) — are only verified by the manual real-Redis script described in the PR body, not by a checked-in test. There is currently no endpoint-level test harness for `server.ts` (no `server.test.ts`), so I'm raising this as non-blocking rather than a must-fix: adding one is a larger lift than this PR's scope. If a lightweight harness is feasible, a regression test asserting "TTL is `CLIENT_REGISTRATION_TTL` again after a successful authorize on an aged record" would lock in the exact fixed behavior and guard against a future refactor silently dropping the `expire` call.

#### Information
- **[functionality] "Idle" is measured from the last `/oauth/authorize`, not the last overall use.** Because the registration record is consumed only at authorize, a client that authorizes once and then lives on refresh tokens for more than 30 days will still have its registration record expire in the background. This is harmless until that client next needs to re-authorize (e.g. its refresh token is revoked/expires), at which point it hits the new friendly re-setup page — a graceful outcome, not a regression, and strictly better than today. Worth being aware of that the fix eliminates the "expires while in use" failure for authorize-driven flows specifically; noting it so the framing ("stops expiring at all") isn't read as covering refresh-token-only clients.
- **[functionality] The suggested command is the bare `npx @insforge/install`** (`server.ts:309`), whereas the README always documents it with `--client <name> --env ...` flags. Running it bare won't complete an install on its own — but the page also tells the user to "remove … and add it back," so the command reads as a pointer to the installer rather than a copy-paste cure. Minor; consider matching the README form if you want it to be directly runnable.

### Non-findings (dimensions with nothing to flag)
- **Security:** No new secrets/PII logged — `client_id` is a public registration identifier, and it's the only value logged (`server.ts:300`). The new HTML page escapes every field and currently interpolates only static strings, so it's XSS-safe today and defensively safe if a `client_id`/`redirect_uri` is later reflected. No auth/authorization check is weakened — the unknown-client path still returns `400` for programmatic callers with a byte-identical JSON body; only the browser presentation changed. No new dependencies.
- **Performance:** One extra O(1) Redis `EXPIRE` per authorize — negligible, no loop, no blocking I/O added to a hot path. No new N+1 or allocations of note.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the Suggestion and Information items above are non-blocking. Clean scope discipline: exactly the two changes described, no drive-by edits.

jwfing (approved):
LGTM - approved.

tonychang04:
## QA — measured against a live server + live Redis

Built the branch, ran `dist/http-server.js` against Redis 7.0.15 with OAuth credentials configured. `npm test` 27/27, `npm run lint` 0 errors (6 pre-existing warnings).

### What holds

**The idle clock works, and only on a successful authorize.**

```json
{ "A_successfulAuthorize":        { "status": 302, "ttlBefore": 120, "ttlAfter": 2592000 },
  "B_failedAuthorize_badRedirect":{ "status": 400, "ttlAfter": 120, "refreshed": false } }
```

The refresh sitting *after* `redirect_uri` validation is the right placement — a request that fails validation leaves the clock alone, so a wrong `redirect_uri` can't be used to keep a registration warm.

**The JSON contract is intact for every non-browser caller.** Same dead `client_id`, four callers:

| Accept | status | content-type | body still JSON | tells the user what to do |
|---|---|---|---|---|
| browser (`text/html,…`) | 400 | `text/html` | no | yes |
| `*/*` (curl default) | 400 | `application/json` | yes | — |
| `application/json` | 400 | `application/json` | yes | — |
| *(no Accept header)* | 400 | `application/json` | yes | — |

So `verify:deploy`, curl and anything scripted see exactly what they saw before, including the 400. Only a real browser gets the page.

**Escaping.** Nothing user-controlled reaches the template today — `heading`/`message`/`action` are all literals in the handler — and I probed it anyway with `client_id="><script>alert(1)</script>` and a `redirect_uri` carrying an `onerror` payload: neither appears raw in the response, and the `client_id` isn't reflected at all. Your escaping test also has teeth: dropping `escapeHtml` from the `message` interpolation fails exactly that test and no other.

### One thing to correct in the claim, not in the code

> an actively-used client stops expiring at all

That is true of a client that keeps *authorizing*, not of a client that is merely *in use* — and those are different populations. Token bindings slide on every authenticated request (`resolveProjectFromToken` → `touchBinding` → `setex` 30d, `oauth-manager.ts:367`). Registrations slide only at `/oauth/authorize`, which is the only other place `mcp:oauth:client:*` is written. Measured — one authenticated MCP request, both keys aged to 120s first:

```json
{ "mcpRequest": 200, "bindingTtlAfterUse": 2592000, "registrationTtlAfterUse": 120 }
```

So the ordinary user — signs in once, then just uses the tools daily — renews their token forever and never returns to authorize, and their registration still dies 30 days after it was written. It stays invisible until something forces a re-auth (a Redis restart or eviction, `/oauth/revoke`, a host move) — and on that day they land on your new page, which is exactly the recovery path this PR adds. So the PR strictly improves things; the claim is just wider than the change.

Closing the residual isn't a one-liner: `TokenBinding` (`oauth-manager.ts:65`) has no `clientId` field, so `touchBinding` has nothing to refresh the registration by. It's one field plus one `expire` — worth a follow-up issue rather than growing this PR, and it may evaporate entirely under the Redis-deprecation plan.

**Ship it from my side.**

*Not tested:* a real browser rendering the page, a complete OAuth round trip against the platform AS (no credentials here — the authorize leg was verified to the 302), and anything on prod.

— Quinn

PR #75, opened 2026-07-29
Slide the client registration whenever its token is used

tonychang04:
Follow-up to #74, from @qa-bot's correction. **Stacked on `fix/client-registration-expiry`** so #74 can ship on its own; I'll retarget to `master` once it merges.

## The gap

#74 refreshes a registration at `/oauth/authorize`. That covers a client that keeps *authorizing* — which is not the same population as a client that is merely *in use*.

- Token bindings renew on **every authenticated request** (`resolveProjectFromToken` → `touchBinding`).
- Registrations renewed **only at `/oauth/authorize`**.

So the ordinary user — signs in once, then only calls tools — renews their token forever while the registration behind it still dies on a fixed schedule from the day it was written. It stays invisible until something forces a re-auth, and then they hit the dead end.

## The change

`TokenBinding` had no way to name its registration, so `touchBinding` had nothing to refresh. Carry `clientId` on the binding where it's created (`authState.clientId` is already in scope there) and refresh both records on the same beat.

`clientId` is **optional on purpose**: `bindTokenToProject` serves the direct bind API, where no registered OAuth client is involved at all, and every binding written before this field existed lacks it too. A refresh failure is logged and swallowed — it's housekeeping, and losing it costs the registration its extension rather than the user their request.

Also moves the registration key prefix and TTL into `oauth-manager.ts` beside the other OAuth prefixes, so #74's authorize-path refresh and this one share a single definition instead of repeating the literal.

## Measured, real Redis

Both keys aged to 120s, then one `touchBinding` — which is what every authenticated MCP request runs:

| binding | binding TTL after use | registration TTL after use |
|---|---|---|
| carries `clientId` (written from this change on) | 2592000 | **2592000** |
| predates the field (everything in Redis today) | 2592000 | **120** |

### The second row is the honest limit of this change

Bindings already in Redis **cannot be repaired** — nothing records which registration issued them, and guessing at that mapping is not something to do on a security-adjacent key. So this stops the bleed from the next sign-in onward rather than rescuing the ~623 clients already connected. Those keep their original expiry, and if something forces them to re-auth they land on #74's recovery page, which is exactly the case #74 was built for.

The two changes are complementary, and neither is complete alone.

31/31 tests, 4 new on `touchBinding` — including that a binding with no `clientId` touches no registration, and that a failing refresh doesn't fail the request.

**Not tested:** a full round trip against the platform AS, or prod.

---
## Summary by cubic
Active OAuth clients no longer expire while their tokens keep working. We now slide the client registration TTL on every authenticated request by carrying `clientId` on token bindings.

- **Bug Fixes**
  - Store optional `clientId` on token bindings and refresh the registration TTL in `touchBinding`.
  - Old bindings without `clientId` keep working but do not refresh the registration.
  - Log and ignore registration TTL refresh failures to avoid breaking requests.
  - Centralize `CLIENT_REGISTRATION_PREFIX` and `CLIENT_REGISTRATION_TTL` in `oauth-manager` and use them in register/authorize; add tests for `touchBinding`.

<sup>Written for commit 9fa55191a628ee84024dad8dbfdcfe7d83a13b3d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/75?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/oauth-manager.test.ts
+++ b/src/http/oauth-manager.test.ts
@@ -1,0 +1,97 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const redisGet = vi.fn();
+const redisSetex = vi.fn();
+const redisExpire = vi.fn();
+
+vi.mock('./redis.js', () => ({
+  getRedisClient: () => ({ get: redisGet, setex: redisSetex, expire: redisExpire }),
+}));
+
+// The manager pulls in the whole platform API surface at import time; none of
+// it is on touchBinding's path.
+vi.mock('./insforge-api.js', () => ({
+  validateToken: vi.fn(),
+  getProjectAccess: vi.fn(),
+  getAllUserProjects: vi.fn(),
+  InsforgeApiError: class extends Error {},
+}));
+
+const { getOAuthManager, CLIENT_REGISTRATION_PREFIX, CLIENT_REGISTRATION_TTL } = await import(
+  './oauth-manager.js'
+);
+
+const TOKEN_HASH = 'a'.repeat(64);
+
+function binding(extra: Record<string, unknown> = {}) {
+  return JSON.stringify({
+    tokenHash: TOKEN_HASH,
+    userId: 'u1',
+    userEmail: 'user@example.com',
+    projectId: 'p1',
+    projectName: 'Project',
+    organizationId: 'o1',
+    accessHost: 'https://p1.example.com',
+    apiKey: 'ik_test',
+    createdAt: 1,
+    lastUsedAt: 1,
+    ...extra,
+  });
+}
+
+describe('OAuthManager.touchBinding', () => {
+  beforeEach(() => {
+    redisGet.mockReset();
+    redisSetex.mockReset().mockResolvedValue('OK');
+    redisExpire.mockReset().mockResolvedValue(1);
+  });
+
+  it('slides the client registration on the same beat as the binding', async () => {
+    // The whole point: a binding renews on every authenticated request, but the
+    // registration behind it only renewed at /oauth/authorize — which an
+    // ordinary user never returns to after signing in once.
+    redisGet.mockResolvedValue(binding({ clientId: 'mcp_abc' }));
+
+    await getOAuthManager().touchBinding(TOKEN_HASH);
+
+    expect(redisSetex).toHaveBeenCalled();
+    expect(redisExpire).toHaveBeenCalledWith(
+      CLIENT_REGISTRATION_PREFIX + 'mcp_abc',
+      CLIENT_REGISTRATION_TTL
+    );
+  });
+
+  it('leaves the registration alone for a binding with no clientId', async () => {
+    // Every binding written before this field existed, plus the direct bind
+    // API, where there is no registered OAuth client at all.
+    redisGet.mockResolvedValue(binding());
+
+    await getOAuthManager().touchBinding(TOKEN_HASH);
+
+    expect(redisSetex).toHaveBeenCalled();
+    expect(redisExpire).not.toHaveBeenCalled();
+  });
+
+  it('does not fail the request when the registration refresh throws', async () => {
+    // A TTL refresh is housekeeping. Losing it costs the registration its
+    // extension, not the user their request.
+    redisGet.mockResolvedValue(binding({ clientId: 'mcp_abc' }));
+    redisExpire.mockRejectedValue(new Error('READONLY'));
+    const logged = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+    await expect(getOAuthManager().touchBinding(TOKEN_HASH)).resolves.toBeUndefined();
+
+    expect(redisSetex).toHaveBeenCalled();
+    expect(logged).toHaveBeenCalled();
+    logged.mockRestore();
+  });
+
+  it('touches nothing at all when the binding is gone', async () => {
+    redisGet.mockResolvedValue(null);
+
+    await getOAuthManager().touchBinding(TOKEN_HASH);
+
+    expect(redisSetex).not.toHaveBeenCalled();
+    expect(redisExpire).not.toHaveBeenCalled();
+  });
+});
--- a/src/http/oauth-manager.ts
+++ b/src/http/oauth-manager.ts
@@ -73,17 +73,32 @@ interface TokenBinding {
   apiKey: string;
   createdAt: number;
   lastUsedAt: number;
+  /**
+   * The registration this binding was issued through, so ordinary use can keep
+   * that registration alive. Optional on purpose: bindTokenToProject serves the
+   * direct bind API, where there is no registered OAuth client at all. Also
+   * absent from every binding written before this field existed.
+   */
+  clientId?: string;
 }
 
 // Redis key prefixes
 const AUTH_STATE_PREFIX = 'mcp:auth:state:';
 const TOKEN_BINDING_PREFIX = 'mcp:auth:binding:';
 const AUTH_CODE_PREFIX = 'mcp:auth:code:';
+export const CLIENT_REGISTRATION_PREFIX = 'mcp:oauth:client:';
 
 // TTLs
 const AUTH_STATE_TTL = 10 * 60; // 10 minutes
 const AUTH_CODE_TTL = 5 * 60; // 5 minutes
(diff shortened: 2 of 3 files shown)


jwfing (commented):
## Review — Slide the client registration whenever its token is used

**Summary:** A small, well-scoped, well-tested follow-up to #74 that closes a real gap — active OAuth clients whose tokens keep renewing but whose registration was previously only refreshed at `/oauth/authorize` — by carrying `clientId` on the token binding and sliding the registration TTL on the same beat as the binding in `touchBinding`.

**Requirements context:** No `/docs/superpowers/` (or other spec/plan) directory exists in this repo, so there is no matching spec — assessing against the PR description, the linked predecessor #74, and the surrounding code. The PR body's stated intent (renew the registration on every authenticated request, not only on re-authorize) is a coherent and accurate description of the problem.

The diff is against the stacked base `fix/client-registration-expiry` (#74); the workspace base ref is `master`, and I reviewed this PR's own commit (`9fa5519`) in isolation.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / functionality — `src/http/oauth-manager.ts:339-347`, `src/http/server.ts:361-362`:** the correctness of this change hinges on the key chain `authorize client_id → authState.clientId → binding.clientId → CLIENT_REGISTRATION_PREFIX + clientId`. I traced it and it's correct today, but the new unit tests exercise `touchBinding` in isolation with a hardcoded `'mcp_abc'` — they don't assert the end-to-end derivation. A single integration-style test (`createAuthorizationCode` writes a binding whose `clientId` matches the registration key that `register` produced, then `touchBinding` slides *that* key) would guard against future drift in how the client id is derived on either side. Non-blocking.

#### Information
- **Performance — `src/http/oauth-manager.ts:322-347`:** `touchBinding` now issues three sequential Redis round-trips per authenticated request (`get` → `setex` → `expire`) instead of two. This runs on every authenticated MCP request via `resolveProjectFromToken`. The added latency is one RTT and negligible in practice; if it ever matters, `setex` + `expire` could be pipelined (`MULTI`) since they don't depend on each other. Noting, not requesting.
- **Security — no concerns.** `clientId` is a non-sensitive public identifier (`mcp_<uuid>`), and the value stored on the binding originates from an `authorize` request that already validated the registration exists (`server.ts:287-310`), so the value reaching `redis.expire` is a known, registered key — no key-injection surface, and the worst case of an arbitrary key would only extend a TTL. The failure log (`oauth-manager.ts:345`) prints only the `clientId`, no token/secret/PII. No auth or authorization checks are touched.
- **Software engineering — `src/http/oauth-manager.test.ts`:** good coverage — the four cases (slides registration when `clientId` present; leaves it alone when absent; swallows an `expire` failure without failing the request; no-ops when the binding is gone) map cleanly to the described behavior and its edge cases. `no-console` is off and there's no max-len rule in `eslint.config.js`, so the long log line and `console.error` are fine. The `swallow-and-log` contract for the housekeeping refresh matches the identical treatment already used on the authorize path (`server.ts:352-356`), so the two refresh sites are consistent.
- **Functionality — the acknowledged limit is honest and correct.** Bindings written before this field (and direct-bind bindings via `bindTokenToProject`, which deliberately omits `clientId`) will have `binding.clientId === undefined` and correctly skip the refresh, exactly as the PR's measured table states. Those clients fall back to #74's re-auth recovery page — the complementary-not-complete framing is accurate.

### Verdict
**approved** (informational — a human still applies the GitHub approval via the approve flow). No blocking issues; the two non-blocking items above are optional. The refactor is a pure string-identity centralization, behavior-preserving, and the new behavior is well covered by tests.

jwfing (commented):
## Review: Slide the client registration whenever its token is used

**Summary:** A tightly-scoped, correct follow-up to #74 that carries `clientId` onto `TokenBinding` and slides the client-registration TTL in `touchBinding`, closing the gap where an in-use-but-not-re-authorizing client's registration silently expired.

### Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the linked predecessor #74, and the surrounding code. The stated intent (registrations should renew on the same beat as their bindings, i.e. on every authenticated request, not only at `/oauth/authorize`) is met.

I verified the change against the PR head (`9fa5519`):
- `npx vitest run src/http/oauth-manager.test.ts` → 4/4 pass.
- `npx tsc --noEmit` → clean.
- `npx eslint` on the three changed files → 0 errors (2 warnings, both pre-existing unused-import warnings on `oauth-manager.ts` lines 7/10 that this PR does not touch).

### Findings

#### Critical
(none)

#### Suggestion
- **Performance — `src/http/oauth-manager.ts:322-349`.** `touchBinding` runs on every authenticated request (via `resolveProjectFromToken`). It now issues the registration `EXPIRE` as a second sequential round-trip after the binding `SETEX`. Combined with the caller's own `getTokenBinding` GET plus `touchBinding`'s repeat GET, an authenticated request now costs ~4 sequential Redis round-trips (was ~3). Low blast radius — Redis calls are cheap and this isn't a correctness issue — but the `SETEX` + `EXPIRE` pair against the same connection could be issued as a pipeline/`MULTI` to shave one RTT off the hot path. Non-blocking.

#### Information
- **Functionality — `src/http/oauth-manager.ts:341`.** `redis.expire` on a registration key that has already lapsed returns `0` rather than throwing, so a client whose registration already died is correctly *not* resurrected (the data is gone and can't be rebuilt from a binding) while a live binding still slides its own TTL. This matches the PR's honest-limit framing; noting it only to confirm the behavior is intended and safe.
- **Software engineering — `src/http/oauth-manager.ts:334-347` / `src/http/server.ts:348-353`.** The `try/expire/swallow-and-log` block is now duplicated in two places with the same rationale. Small enough that inlining is defensible, but a shared `slideRegistrationTtl(redis, clientId)` helper on `OAuthManager` would let both the authorize path and the touch path share one definition (the constants are already centralized). Author's call.
- **Test coverage — `src/http/oauth-manager.test.ts`.** Coverage of the changed behavior is good: registration slides with `clientId`, is left alone without one, a throwing refresh doesn't fail the request, and a missing binding touches nothing. The only untested-but-asserted-by-prose claim is that `bindTokenToProject` writes a binding *without* `clientId` (the direct-bind path); it's covered indirectly by the "no clientId" branch, so a dedicated assertion is optional.

#### Security
No security-relevant concerns. `clientId` is the public OAuth client identifier (not a secret/PII), it is only ever read from the binding written during the client's own authorized flow (no cross-client TTL manipulation surface), and the error log at line 345 exposes only that identifier. Auth/PKCE paths are untouched.

### Verdict
**approved** (informational — a human still gives the GitHub approval). Zero Critical findings; the suggestions above are non-blocking. Clean, well-reasoned, well-tested change that correctly complements #74.

*Note: this PR is stacked on `fix/client-registration-expiry` (#74), so its base must merge first; the diff reviewed here is #75's changes alone.*

jwfing (approved):
LGTM - approved.

tonychang04:
## QA — verified against a live server, live Redis and a stand-in platform

This closes the divergence I measured on #74 (binding slid on every request, registration only at `/oauth/authorize`). Both records aged to 120s, then one ordinary authenticated request:

| case | request | binding TTL | registration TTL |
|---|---|---|---|
| binding carries `clientId` (post-#75 sign-in) | 200 | 2592000 | **2592000** — was staying at 120 |
| binding has no `clientId` (every existing one) | 200 | 2592000 | untouched at 120 |
| registration already expired | 200 | 2592000 | no error, request unaffected |

So the fix works, the ~623 live bindings keep working without it (the 30-day tail Max accepted when he ruled on this), and a missing registration can't break a request — which matters because this now runs on the hot path rather than once at authorize.

31/31 tests, lint 0 errors.

**Ship it from my side.**

*Not tested:* a real OAuth round trip against the platform AS (no credentials here — the binding was seeded in the shape `resolveProjectFromToken` reads), Redis failover, and prod.

— Quinn

PR #76, opened 2026-07-29
Remote mode ignores a caller-supplied apiKey

tonychang04:
Came out of checking @infra-bot's item 5 ("wrong project is REFUSED") in source rather than leaving it as an assumption for QA. **The premise turned out to be wrong, and checking it found a different thing that is real.**

## `projectId` is not a tool argument

```
tools taking projectId as an input argument:  0
how it enters:  registerInsforgeTools(host, { projectId, ... })  — bound per session at sign-in
```

So "an agent names someone else's project" isn't reachable through a tool argument today. Good news, but it means the protection isn't where it was thought to be.

## `apiKey` *is* a tool argument, and it won

```js
const apiKey = toolApiKey?.trim() || GLOBAL_API_KEY;   // caller wins
```

14 tools across `database.ts`, `storage.ts`, `functions.ts`, `docs.ts` expose `apiKey: z.string().optional()`. That exists for the self-hosted stdio mode, where the operator running the binary legitimately supplies the credential. In remote mode the credential is bound to the session, so honouring it means **the session credential is not authoritative** — an agent can substitute any key it knows.

### Scope, stated precisely rather than alarmingly

`apiBaseUrl` is **not** a tool argument — it's fixed per session. So a key for another project sent to *this* project's host does not authenticate. **This is not cross-project access.** What it is:

- an override nobody needs in remote mode, and
- a way to swap the session's credential for a different one against the same project — an escalation if the session was bound with a lesser key than the caller supplies (`anon_` vs `ik_`).

## The change

Ignored in remote mode only. stdio is unchanged, because there the argument *is* the configuration mechanism. One line, no schema change, so no client breaks.

## Tests

They pin which key actually reaches the backend by watching the outgoing request per mode, rather than reasoning about the closure. Two things worth noting:

- They mock **`node-fetch`**, not the global — the tool modules `import fetch from 'node-fetch'`, so mocking `globalThis.fetch` would have silently tested nothing. My first attempt did exactly that and failed loudly, which is the only reason I noticed.
- **Confirmed they have teeth:** reverting the one-line change fails the remote case, and only the remote case.

26/26 green.

## For @qa-bot

This narrows what's worth your time on item 5: the projectId vector doesn't exist, so the test to run is *credential substitution within a session*, plus the original question of whether `getProjectApiKey` refuses a project the authenticated user doesn't own — which is platform-side and unaffected by this PR.

---
## Summary by cubic
Remote mode now ignores a caller-supplied `apiKey` and always uses the session credential. This blocks within-project credential swapping and does not allow cross-project access; stdio mode is unchanged.

- **Bug Fixes**
  - Change: In remote mode, `getApiKey` drops the per-call `apiKey` and uses the session/global key; stdio still honors the argument.
  - Tests: Added coverage that inspects outgoing requests (mocking `node-fetch`) to confirm the correct key is sent per mode.

<sup>Written for commit db782037ee2401161b80ef46732248cc42401c4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/76?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API key handling in remote mode by consistently using the session credential.
  * Preserved per-call API key overrides for self-hosted and local deployments.
  * Maintained existing behavior when no per-call API key is provided.

* **Tests**
  * Added coverage verifying API key behavior across remote and self-hosted runtime modes.

### DIFF
--- a/src/shared/tools/api-key-binding.test.ts
+++ b/src/shared/tools/api-key-binding.test.ts
@@ -1,0 +1,118 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
+import { registerInsforgeTools } from './index.js';
+import { sdkToolHost } from './host.js';
+
+/**
+ * In remote mode the project credential is bound to the session at sign-in.
+ * Several tools also expose an optional `apiKey` argument, which exists for the
+ * self-hosted stdio mode where the caller legitimately supplies its own
+ * credential. Honouring that argument in remote mode would mean the session
+ * credential is not authoritative — a caller could substitute any key it knows.
+ *
+ * These tests pin which key actually reaches the backend, per mode, by watching
+ * the outgoing request rather than by reasoning about the closure.
+ */
+
+const SESSION_KEY = 'ik_bound_to_the_session';
+const CALLER_KEY = 'ik_supplied_by_the_caller';
+
+const sentKeys: string[] = [];
+
+function keyOf(init?: { headers?: Record<string, string> }): string | undefined {
+  return init?.headers?.['x-api-key'];
+}
+
+function reply(body: unknown, status = 200) {
+  return {
+    ok: status < 400,
+    status,
+    headers: { get: () => 'application/json' },
+    json: async () => body,
+    text: async () => JSON.stringify(body),
+  };
+}
+
+// The tool modules do `import fetch from 'node-fetch'`, so the global is never
+// consulted — mocking it would silently test nothing.
+vi.mock('node-fetch', () => ({
+  default: vi.fn(async (input: unknown, init?: { headers?: Record<string, string> }) => {
+    const url = String(input);
+    if (url.endsWith('/api/health')) {
+      return reply({ version: '1.2.0' });
+    }
+    const key = keyOf(init);
+    if (key) sentKeys.push(key);
+    return reply({ ok: true });
+  }),
+}));
+
+beforeEach(() => {
+  sentKeys.length = 0;
+});
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+/** Fail loudly rather than silently testing a tool that has no apiKey argument. */
+function requireTool(tools: Map<string, (args: any) => any>, name: string) {
+  const tool = tools.get(name);
+  if (!tool) throw new Error(`no tool named ${name}; got: ${[...tools.keys()].join(', ')}`);
+  return tool;
+}
+
+/** Register against a real McpServer and return the callable tool map. */
+async function toolsFor(mode: 'remote' | undefined) {
+  const server = new McpServer({ name: 'test', version: '0.0.0' });
+  const captured = new Map<string, (args: any) => any>();
+  const host = sdkToolHost(server);
+  const spy = {
+    registerTool(name: string, description: string, schema: any, handler: any) {
+      captured.set(name, handler);
+      host.registerTool(name, description, schema, handler);
+    },
+  };
+
+  await registerInsforgeTools(spy, {
+    apiKey: SESSION_KEY,
+    apiBaseUrl: 'https://project.example.com',
+    mode,
+    projectId: 'proj_1',
+    accessToken: 'tok',
+  });
+
+  return captured;
+}
+
+describe('which api key reaches the backend', () => {
+  it('remote mode ignores a caller-supplied apiKey and uses the session credential', async () => {
+    const tools = await toolsFor('remote');
+    const tool = requireTool(tools, 'get-table-schema');
+
+    await tool({ apiKey: CALLER_KEY, tableName: 't' }).catch(() => {});
+
+    expect(sentKeys.length).toBeGreaterThan(0);
+    expect(sentKeys).not.toContain(CALLER_KEY);
+    expect(sentKeys.every((k) => k === SESSION_KEY)).toBe(true);
+  });
+
+  it('self-hosted mode still honours it, because that is how stdio is configured', async () => {
+    // Not a bug there: the operator running the binary supplies the key.
+    const tools = await toolsFor(undefined);
+    const tool = requireTool(tools, 'get-table-schema');
+
+    await tool({ apiKey: CALLER_KEY, tableName: 't' }).catch(() => {});
+
+    expect(sentKeys).toContain(CALLER_KEY);
+  });
+
+  it('remote mode with no argument at all is unchanged', async () => {
+    const tools = await toolsFor('remote');
+    const tool = requireTool(tools, 'get-table-schema');
+
+    await tool({ tableName: 't' }).catch(() => {});
+
+    expect(sentKeys.every((k) => k === SESSION_KEY)).toBe(true);
+  });
+});
--- a/src/shared/tools/index.ts
+++ b/src/shared/tools/index.ts
@@ -237,7 +237,14 @@ export async function registerInsforgeTools(host: ToolHost, config: ToolsConfig 
   }
 
   const getApiKey = (toolApiKey?: string): string => {
-    const apiKey = toolApiKey?.trim() || GLOBAL_API_KEY;
+    // The per-call apiKey argument exists for the self-hosted stdio mode, where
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review: Remote mode ignores a caller-supplied `apiKey`

**Summary:** A tight, well-scoped security-hardening change that makes the session-bound credential authoritative in remote mode while leaving stdio behavior byte-for-byte identical — verified sound and well-tested.

### Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` (or any `docs/` spec dir), consistent with its conventions. Assessed against the PR description, the threat model it lays out, and the surrounding code. The PR's own framing (credential *substitution within a session*, explicitly **not** cross-project access since `apiBaseUrl` is fixed per session) matches what the code actually does.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / defense-in-depth — `src/shared/tools/index.ts:247`.** The fix is correct precisely *because* every tool routes its `apiKey` arg through `getApiKey()` (verified: `database.ts`, `storage.ts`, `functions.ts`, `docs.ts`, and `deployment.ts` get-container-logs all call `getApiKey(args.apiKey)`; several inline it directly into `x-api-key` headers, e.g. `functions.ts:39` / `storage.ts:57`). That funnel is a convention, not a type-enforced invariant — a future tool that writes `headers: { 'x-api-key': args.apiKey }` directly would silently reintroduce the bypass in remote mode. Consider a guard that makes this structural: e.g. a lint/test asserting no handler reads `args.apiKey` outside `getApiKey`, or dropping the `apiKey` field from the schema entirely in remote mode so it can't reach a handler. Non-blocking; current tools are all clean.
- **Test coverage — `api-key-binding.test.ts`.** Coverage exercises `get-table-schema` as the single representative tool. Since all tools share the same `getApiKey` funnel this is reasonable, but adding one case against a tool that inlines the key into headers (e.g. `list-buckets`) would pin the behavior at a second, structurally-different call site and pair naturally with the guard above.

#### Information
- **Functionality — tool schemas unchanged (intentional).** The `apiKey` field remains advertised on all tools with the description *"optional if provided via --api_key"* (e.g. `database.ts:91`). In remote mode a caller-supplied value is now silently ignored, so the description is slightly misleading there. Keeping the schema stable to avoid breaking clients is the right call; just noting the doc-string could optionally clarify that the arg is honored in stdio only.
- **Software engineering — the fix preserves stdio exactly.** `(isRemote ? undefined : toolApiKey?.trim()) || GLOBAL_API_KEY` reduces to the original `toolApiKey?.trim() || GLOBAL_API_KEY` when `!isRemote`, so no self-hosted behavior changes — confirmed.
- **Test quality — strong.** The `vi.mock('node-fetch', …)` correctly targets the module the tools actually import (`import fetch from 'node-fetch'`, confirmed at `database.ts:2`), not `globalThis.fetch`; the three cases (remote ignores, stdio honors, remote no-arg) plus the author's stated revert-to-red check give the suite genuine teeth. `requireTool` failing loudly on a missing tool guards against silently testing nothing.

### Dimension coverage
- **Software engineering:** Clean, minimal, centralized; good TDD. Suggestions above are forward-looking, not current defects.
- **Functionality:** Correctly solves the stated problem (session credential now authoritative in remote); stdio path unchanged; no schema/API break.
- **Security:** This *is* a hardening — reduces attack surface (blocks within-session credential swap, e.g. `anon_`→`ik_` escalation). No new user input reaches SQL/shell; no secrets logged or returned. Behavior for the sensitive `run-raw-sql` tool now always uses the session key in remote.
- **Performance:** No impact — a single boolean branch in an already-hot helper.

### Verdict
**approved** (informational; a human still gives the explicit GitHub approval). Zero Critical findings — the two Suggestions and Information notes are non-blocking.

jwfing (commented):
## Review: Remote mode ignores a caller-supplied apiKey

**Summary:** A precise, well-scoped one-line hardening that ignores the per-call `apiKey` argument in remote mode (where the credential is session-bound) while preserving it for self-hosted stdio; the accompanying test has real teeth. No blocking issues.

### Requirements context
No matching spec/plan found — `docs/superpowers/` (and any `docs/specs/`) does not exist in this repo, consistent with prior knowledge of `insforge-mcp`. Assessed against the PR description, the linked `@infra-bot` item 5, and the surrounding code. The PR's own framing (this is a credential-substitution-within-a-session hardening, **not** cross-project access because `apiBaseUrl` is session-fixed) matches what the code actually does.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality / UX — silently-dropped argument** (`src/shared/tools/index.ts:247`): In remote mode a caller-supplied `apiKey` is discarded with no signal back to the caller. The schema still advertises `apiKey: z.string().optional()` on all 14 tools with the description *"optional if provided via --api_key"*, so an agent can pass a key, observe success, and reasonably believe it took effect. This is the right compatibility call (the PR correctly avoids a schema change so no client breaks), so I would not block on it — but consider a follow-up that either (a) updates the description to note the argument is ignored in remote mode, or (b) logs a debug line when a remote call supplies an `apiKey` that is dropped, to avoid silent surprise during debugging.

#### Information
- **Software engineering — test coverage is representative, not exhaustive** (`src/shared/tools/api-key-binding.test.ts:89-117`): The tests exercise a single tool (`get-table-schema`). Because the fix is centralized in the shared `getApiKey` closure and every one of the 14 tools routes through `getApiKey(apiKey)` (verified across `database.ts`, `storage.ts`, `functions.ts`, `docs.ts`, `deployment.ts`), one representative tool is sufficient to pin the behavior — no need to fan out. Worth stating explicitly so the single-tool scope reads as deliberate rather than a gap.
- **Software engineering — mock fidelity is correct and deliberate** (`api-key-binding.test.ts:36-47`): Mocking `node-fetch`'s `default` export rather than `globalThis.fetch` is the right call — the tool modules do `import fetch from 'node-fetch'` (confirmed at `index.ts:1`, `database.ts`), so a global mock would have tested nothing. The PR body already documents this; noting it here because it's the kind of subtlety that makes the test trustworthy.
- **Functionality — fix is reachable in every remote entry point**: Verified all three remote registrations in `src/http/session-manager.ts:83,202,242` pass `mode: 'remote'`, `src/stdio/index.ts:31` passes `mode: 'local'`, and `isRemote = config.mode === 'remote'` (`index.ts:164`). In remote mode `GLOBAL_API_KEY` is `sessionData.apiKey` (`session-manager.ts:81`), so falling back to it after ignoring the caller key correctly yields the session-bound credential.

### Dimension coverage
- **Software engineering:** Centralized fix at the single chokepoint rather than 14 edit sites — good. Regression test confirmed to fail only the remote case on revert (per PR body), and the mock targets the correct module boundary. Comment on `index.ts:240-246` clearly documents the reasoning and the scope limits.
- **Functionality:** Correctly narrows the intended vector (credential substitution within a session, escalation from a lesser `anon_` to a supplied `ik_`). stdio behavior is preserved by design. No behavior change for callers that omit `apiKey`. The projectId vector the PR set out to check is correctly identified as non-existent (not a tool argument) and left alone — good scope discipline.
- **Security:** Net improvement — removes a caller's ability to override the session credential in remote mode. No secrets/PII newly logged or returned; the ignored key is simply dropped. No auth check weakened. No new dependencies. The residual (a caller can still authenticate with the session-bound key, unchanged) is out of scope, and the platform-side `getProjectApiKey` ownership check is correctly flagged as unaffected and left to QA.
- **Performance:** No impact — one boolean branch in an existing per-call path; no new I/O, allocation, or loops.

### Verdict
**approved** (informational — the human still approves via the GitHub approve flow). Zero Critical findings; the Suggestion and Information items are non-blocking. Clean, minimal, in-scope, and well-tested.

jwfing (approved):
LGTM - approved.

PR #77, opened 2026-07-29
Pin the published tool names, in both modes

tonychang04:
@manager cited the #68 check as the reason a drive-by rename would be caught. I went and read it, because policy resting on a guarantee is only as good as the guarantee. **It doesn't check that.**

## What #68 actually asserts

`host.test.ts` checks that every registered tool has a non-empty, unique name, that the count matches, and — one name, via `toContain` — that `fetch-docs` exists. All of that stays true after `run-raw-sql` becomes `db-query`.

**Demonstrated rather than argued.** Renaming `run-raw-sql` → `db-query` in `database.ts`:

```
× the npm binary registers exactly its documented names        <- new
× the hosted server registers exactly its documented names     <- new
✓ every pre-existing host test                                 <- all still green
```

So the safety net was the intent, not the code. Now it's the code.

## The thing writing it turned up: there are TWO published surfaces

Not one. Registering in each mode and diffing:

```
only self-hosted:  bulk-upsert       (reads a local filePath)
only remote:       start-deployment  (triggers the build after the agent
                                      has uploaded from its own machine)
```

The divergence is **deliberate and correct** — a hosted server can't touch the user's filesystem, so file-touching tools are replaced by variants that hand the agent a command to run (`database.ts:179`, `deployment.ts:556`).

But it means someone writing a prompt against `mcp.insforge.dev` and someone writing one against the npm binary are writing against **different contracts, and both are contracts**. That's worth stating explicitly given we've been discussing "the invocation surface" as a single thing all evening.

So both lists are pinned, and **the delta between them is pinned separately** — a tool quietly appearing in one mode and not the other shows up as a test failure rather than a support thread.

## What a failure here means

Not a broken test. The surface moved, and the question is whether that was intended:

- **adding a tool** → add it to the list, minor version
- **renaming or removing** → breaking change, major version, alias + deprecation window, not a rename in a PR

The comment says exactly that, so whoever hits it at 2am doesn't just edit the list to make it green.

27/27. Independent of the OAuth work — it touches no runtime code, only adds a test file.

---
## Summary by cubic
Pins the published tool names for both self-hosted and remote modes, and locks the exact delta between them. Adds tests that fail on renames, additions, or removals, with no runtime changes.

<sup>Written for commit fd3b856b17c19ca899a40d21f96a6f4f6b5650d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/77?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/shared/tools/published-surface.test.ts
+++ b/src/shared/tools/published-surface.test.ts
@@ -1,0 +1,152 @@
+import { describe, it, expect, vi } from 'vitest';
+import { registerInsforgeTools } from './index.js';
+import type { ToolHost } from './host.js';
+
+/**
+ * Tool names are the published contract.
+ *
+ * They come back in `tools/list` and people write them into agent prompts,
+ * skills and documentation by name. Renaming one breaks every instruction
+ * written against this server silently — the tool simply stops existing, and
+ * nothing in the error points at the cause.
+ *
+ * The existing host tests check that every registered tool has a non-empty,
+ * unique name. That passes just as happily after `run-raw-sql` becomes
+ * `db-query`. This is the test that doesn't.
+ *
+ * A failure here is not a broken test. It means the published surface moved,
+ * and the question is whether that was intended:
+ *   - adding a tool          -> add it to the list, minor version
+ *   - renaming or removing   -> breaking change, major version, and it needs
+ *                               an alias and a deprecation window rather than
+ *                               a rename in a PR
+ */
+
+// Registration refuses to start unless it can read the backend version, and the
+// registered set is version-gated. A version above every gate is what makes
+// this the FULL surface rather than whatever a particular backend allows.
+vi.mock('node-fetch', async () => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const actual = (await vi.importActual('node-fetch')) as any;
+  return {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    default: vi.fn(async (url: string | URL, init?: any) => {
+      if (url.toString().endsWith('/api/health')) {
+        return {
+          ok: true,
+          status: 200,
+          json: async () => ({ status: 'ok', version: '99.99.99', service: 'insforge' }),
+        };
+      }
+      return actual.default(url, init);
+    }),
+  };
+});
+
+/**
+ * The published tool surface as of @insforge/mcp 1.2.x.
+ *
+ * Sorted so the diff on a change is readable. Do not edit this to make a test
+ * pass — edit it because you decided to change the contract.
+ */
+const SELF_HOSTED_TOOL_NAMES = [
+  'bulk-upsert',
+  'create-bucket',
+  'create-deployment',
+  'create-function',
+  'delete-bucket',
+  'delete-function',
+  'download-template',
+  'fetch-docs',
+  'fetch-sdk-docs',
+  'get-anon-key',
+  'get-backend-metadata',
+  'get-container-logs',
+  'get-function',
+  'get-table-schema',
+  'list-buckets',
+  'run-raw-sql',
+  'update-function',
+];
+
+/**
+ * Remote deliberately differs, and the difference is not arbitrary: a hosted
+ * server cannot touch the user's filesystem, so the tools that read or write
+ * local files are replaced by variants that hand the agent a command to run.
+ *
+ *   only self-hosted:  bulk-upsert      (reads a local filePath)
+ *   only remote:       start-deployment (triggers the build after the agent
+ *                                        has uploaded from its own machine)
+ *
+ * Both are published surfaces. Someone writing a prompt against
+ * mcp.insforge.dev and someone writing one against the npm binary are writing
+ * against different contracts, and both are contracts.
+ */
+const REMOTE_TOOL_NAMES = [
+  'create-bucket',
+  'create-deployment',
+  'create-function',
+  'delete-bucket',
+  'delete-function',
+  'download-template',
+  'fetch-docs',
+  'fetch-sdk-docs',
+  'get-anon-key',
+  'get-backend-metadata',
+  'get-container-logs',
+  'get-function',
+  'get-table-schema',
+  'list-buckets',
+  'run-raw-sql',
+  'start-deployment',
+  'update-function',
+];
+
+/** The extra config the hosted server passes. */
+const REMOTE = { mode: 'remote' as const, projectId: 'proj_1', accessToken: 'tok' };
+
+async function registeredNames(extra?: Record<string, unknown>): Promise<string[]> {
+  const names: string[] = [];
+  const host: ToolHost = {
+    registerTool(name) {
+      names.push(name);
+    },
+  };
+  await registerInsforgeTools(host, {
+    apiKey: 'test-key',
+    apiBaseUrl: 'http://localhost:7130',
+    ...extra,
+  });
+  return names;
+}
+
+describe('published tool surface', () => {
+  it('the npm binary registers exactly its documented names', async () => {
+    const names = (await registeredNames()).sort();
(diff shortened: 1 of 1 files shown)


jwfing (commented):
## Review: Pin the published tool names, in both modes

**Summary:** A test-only PR that snapshots the published tool-name surface for both the local (npm/stdio) and remote (hosted/http) modes and pins the exact delta between them; it adds one file and touches no runtime code, and I verified the pinned lists match the actual registration exactly.

### Requirements context
`insforge-mcp` has no `/docs/superpowers/` (or other spec/plan) directory — no matching spec/plan found, so I assessed against the PR description alone plus the actual registration code. The stated intent (the existing `host.test.ts` checks name non-emptiness/uniqueness/count but *not* the specific published names, so a rename slips through) is accurate: nothing in the pre-existing suite pins `run-raw-sql` et al. by name.

### Verification performed
I reproduced the PR's claims against the code at `fd3b856`:

- **Lists are exact.** Enumerating every `registerTool(...)` across `database.ts`, `deployment.ts`, `docs.ts`, `functions.ts`, `storage.ts` and applying the mode/version gating in `index.ts` yields exactly the two pinned 17-name lists.
- **The delta is real and correct.** `bulk-upsert` is local-only (filtered by `LOCAL_ONLY_TOOLS`, `index.ts`), and `start-deployment` is remote-only (registered only in the `if (isRemote)` branch, `deployment.ts:654-656`). Matches the third test.
- **Mode simulation is faithful.** The real stdio entrypoint passes `mode: 'local'` (`src/stdio/index.ts:31`) and the http session manager passes `mode: 'remote'` (`src/http/session-manager.ts:83/202/242`). The test omits `mode` for the npm case, which yields the same `isRemote === false`, so the registered surface is identical.
- **Full-surface framing holds today.** The `version: '99.99.99'` mock clears the only two active min-gates (`create-deployment` ≥1.4.7, `fetch-sdk-docs` ≥1.5.1, `index.ts`), so both gated tools are included.
- **Gates pass locally:** `vitest run` on the new file → 4/4 passing; `eslint` clean; `tsc --noEmit` clean.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / robustness — `published-surface.test.ts:32-47`:** The "full surface" is obtained by forcing `version: '99.99.99'`, which is above every *minVersion* gate. That is correct for the current gate map, but a future `maxVersion`-gated (deprecated) tool would be *excluded* at a very-high version and would silently never appear in the pinned lists — the opposite of what the comment ("A version above every gate is what makes this the FULL surface") implies. Not a defect today (no `maxVersion` entries are active in `index.ts`), but worth a one-line note in the test, or pinning at a version chosen to include max-gated tools, if deprecation gating is ever used.

#### Information
- **`published-surface.test.ts:114-116`:** The mock `ToolHost.registerTool` captures only the name and discards `description`/`inputSchema`/`handler`. That's the right scope for a name-pinning test — just flagging that "registers exactly its documented names" pins *names* only, not that each is wired with a valid schema/handler (already covered elsewhere).
- **`published-surface.test.ts` npm-case:** The real stdio entrypoint passes `mode: 'local'` explicitly while the test omits `mode`. Equivalent today (`isRemote` only checks `=== 'remote'`), but passing `mode: 'local'` in the local case would mirror the real call site 1:1 and be robust to any future `local`-vs-`undefined` distinction.
- The comment's "as of @insforge/mcp 1.2.x" is accurate — `package.json` is at `1.2.10`.

### Dimension coverage
- **Software engineering:** Strong. This *is* test coverage; it fills a real gap, mirrors the existing `host.test.ts` vitest style, and is thoroughly documented. Suggestions above are minor robustness notes.
- **Functionality:** Verified — the pinned lists and delta match the actual registration, and the suite passes.
- **Security:** No security-relevant changes — test-only, no runtime code, no user input paths, no secrets (only literal `'test-key'`/`'tok'` fixtures).
- **Performance:** No performance-relevant changes — registration is invoked a handful of times against a mocked fetch; negligible.

### Verdict
**approved** (informational; posted as a comment — the GitHub green-check approval remains a separate human action). Zero Critical findings. A clean, well-reasoned, in-scope test addition; the two Suggestion/Information notes are optional polish.

jwfing (commented):
## Review: Pin the published tool names, in both modes

**Summary:** A well-crafted, correctly-scoped test-only PR that pins the published tool-name surface for both self-hosted and remote modes and locks the delta between them; the pinned lists exactly match the actual registration code, and the suite passes.

### Requirements context
No `docs/superpowers/` (or `docs/specs/`) directory exists in this repo — no matching spec/plan found, so I assessed against the PR description alone plus the surrounding registration code. The PR's stated intent — replace the *implied* rename-safety of #68's `host.test.ts` with an *explicit* name-contract test, across both published surfaces — is met.

### Verification performed
I checked the pinned lists against the real registrars rather than trusting the description:
- Enumerated every `registerTool(...)` call across `docs.ts`, `database.ts`, `storage.ts`, `functions.ts`, `deployment.ts` and traced the `isRemote` branches. Shared surface = 16 tools; `bulk-upsert` is local-only (`index.ts:72-74` `LOCAL_ONLY_TOOLS`), `start-deployment` is remote-only (`deployment.ts:655`). That yields **17 self-hosted / 17 remote**, matching `SELF_HOSTED_TOOL_NAMES` and `REMOTE_TOOL_NAMES` exactly.
- Confirmed the version mock (`99.99.99`) clears every gate in `TOOL_VERSION_REQUIREMENTS` (`create-deployment` ≥1.4.7, `fetch-sdk-docs` ≥1.5.1), so the test legitimately captures the *full* surface as claimed.
- Confirmed the `node-fetch` default-import mock is the right seam: `fetchBackendVersion` (`index.ts:124`) is the only registration-time fetch; domain fetches happen only inside handlers.
- Confirmed the REMOTE config mirrors the real hosted entrypoint (`src/http/session-manager.ts:83,202,242` pass `mode:'remote'` + `projectId` + `accessToken`).
- Ran the gates locally: **4/4 tests pass**, `tsc --noEmit` clean, `eslint` clean on the new file.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / coverage — `src/shared/tools/published-surface.test.ts:75-100`:** The contract pinned is tool *names* only. A tool's input schema and description are arguably also part of the published contract (agents/skills bind to arg names), and a silently-narrowed `z` shape would sail past this test. That's consistent with the PR's explicit scope ("tool names are the published contract"), so I'd treat schema-pinning as a possible follow-up rather than a gap here — flagging so it's a conscious boundary, not an oversight.

#### Information
- **Fidelity nit — `published-surface.test.ts:110-119`:** the self-hosted case omits `mode` entirely and relies on `isRemote` defaulting to false. Behaviorally identical to the real stdio binary, which passes `mode: 'local'` explicitly (`src/stdio/index.ts:31`). Passing `mode: 'local'` would mirror the actual invocation exactly; purely cosmetic.
- **`published-surface.test.ts:69`:** the `@insforge/mcp 1.2.x` comment matches `package.json` (1.2.10) today. It's an illustrative label, not asserted by any test, so it will silently drift on version bumps — fine, just noting it's documentation not a guardrail.
- **Security / performance:** no security-relevant changes (test-only; the `'tok'`/`'test-key'` values are dummies, no real secrets, no new input paths). No performance impact — no runtime code touched; registration is mocked at the health check.

### Verdict
**approved** (informational — human still gives the GitHub approval). Zero Critical findings: the test is correct, verified green, and matches the real surface. The Suggestion and Information items are non-blocking.

jwfing (approved):
LGTM - approved.

PR #78, opened 2026-07-29
Make the container actually able to start

tonychang04:
@tony pasted a deployment log and it names the cause of every failed Manufact deploy. **It was not Redis.**

```
npm error Missing script: "start"
INFO Main child exited normally with code: 1
[   12.255664] reboot: Restarting system
```

The platform runs `npm start`. There is no `start` script, so npm exits 1 **before any of our code runs**, the container restarts, and it loops. That's why no runtime log ever reached us, why `/health` kept answering the platform's placeholder, and why 19 deploys produced no diagnosis. Merging the Redis fix would not have changed it — and I was one of the people saying it would.

Reproduces on master: `npm start` prints exactly the error above.

## Two more sat behind it

Each would have caused the *next* deploy to fail differently:

**Nothing read `PORT`.** Both `--port` and `--host` had hardcoded commander defaults, so `cliOptions.port` was never empty and an injected `PORT` was ignored — the process would bind 3000 whatever the platform assigned. (I initially "fixed" this with a fallback *after* `cliOptions.port`, which was dead code for exactly that reason. Caught by testing it, not by reading it.) Defaults now come from `PORT`/`HOST` when the flag is absent; an explicit flag still wins, since commander only applies a default when the flag isn't given.

**Default host is `127.0.0.1`.** Even once it started it would accept only loopback and the edge couldn't reach it. The start script sets `--host 0.0.0.0` explicitly rather than changing the default — a locally-run binary stays off the network, and the container's exposure is visible in the script rather than implied.

## Verified against the built artifact

| case | result |
|---|---|
| `npm start`, `PORT=8471` injected | `0.0.0.0:8471`, `/health` 200 |
| `--port 8472` with `PORT=9999` | `127.0.0.1:8472` — flag wins |
| `PORT` set, no flags | uses `PORT` |
| `PORT` unset, no flags | `127.0.0.1:3000` — unchanged |

23/23, tsc and eslint clean.

## Still unverified

Whether the platform runs a build step. There's no `prepare` script, so if it only installs and starts, `dist/` won't exist and the next error will be a missing module rather than a missing script. I've left that alone rather than guessing — the next deploy log will say. If it does turn out to need one, `"prepare": "npm run build"` is the fix.

---
## Summary by cubic
Fixes container boot loops by adding a proper start command and honoring `PORT`/`HOST` so the server starts and is reachable in deploys. Local runs stay on loopback; the container binds to `0.0.0.0`.

- **Bug Fixes**
  - Add `start` script: `node dist/http-server.js --host 0.0.0.0`.
  - Use `PORT` and `HOST` as defaults when flags aren’t provided; explicit flags still win.
  - Keep default host as `127.0.0.1` for local safety; container exposure is explicit via the script.

<sup>Written for commit c58d5d03f63ba13e9ec40c445b083d070960553b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/78?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "insforge-mcp-server": "./dist/http-server.js"
   },
   "scripts": {
+    "start": "node dist/http-server.js --host 0.0.0.0",
     "dev:stdio": "tsx watch src/stdio/index.ts",
     "dev:http": "tsx watch src/http/server.ts",
     "build": "tsup",
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -10,8 +10,11 @@ program
   .name('insforge-mcp-server')
   .description('HTTP MCP server for Insforge backend-as-a-service')
   .version(PACKAGE_VERSION, '-v, --version')
-  .option('--port <number>', 'Port to run HTTP server on', '3000')
-  .option('--host <string>', 'Host to bind to', '127.0.0.1');
+  // Defaults come from the environment so a container platform's injected
+  // PORT/HOST are honoured. An explicit flag still wins: commander only
+  // applies a default when the flag is absent.
+  .option('--port <number>', 'Port to run HTTP server on', process.env.PORT || '3000')
+  .option('--host <string>', 'Host to bind to', process.env.HOST || '127.0.0.1');
 program.parse(process.argv);
 
 const cliOptions = program.opts();
@@ -21,10 +24,23 @@ const cliOptions = program.opts();
 // ============================================================================
 
 export const SERVER_CONFIG = {
-  /** Port to run HTTP server on */
+  /**
+   * Port to run HTTP server on.
+   *
+   * Sourced from PORT when no flag is given — every container platform assigns
+   * one that way and nothing here was reading it, so a hosted deploy bound
+   * 3000 regardless of what it had been told to use.
+   */
   port: parseInt(cliOptions.port) || 3000,
 
-  /** Host to bind to */
+  /**
+   * Host to bind to.
+   *
+   * Stays loopback unless asked otherwise: a locally-run binary should not be
+   * reachable from the network. A container opts in with --host 0.0.0.0 or
+   * HOST, which the start script does explicitly so it is visible rather than
+   * implied.
+   */
   host: cliOptions.host || '127.0.0.1',
 
   /** Public URL of this MCP server */


jwfing (commented):
## Review: Make the container actually able to start

**Summary:** Adds the missing `start` script and makes `--port`/`--host` defaults read from the environment so a container platform's injected `PORT`/`HOST` are honoured — a small, correct, well-reasoned fix that directly addresses the diagnosed deploy loop.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or equivalent) spec directory. Assessed against the PR description, the linked deploy log, and the surrounding code alone.

### Findings

#### Critical
(none) — the change is correct. Verified the mechanism holds end-to-end:
- `dist/http-server.js` is the tsup build of `src/http/server.ts` (`tsup.config.ts:6`), whose `startServer()` binds `app.listen(SERVER_CONFIG.port, SERVER_CONFIG.host, …)` (`src/http/server.ts:1086`). So the new `start` script (`package.json:14`) launches the right entrypoint.
- The commander "flag wins, env is the default" logic in `src/http/config.ts:16-17` is right: commander only applies the default when the flag is absent, so `--port 8472` beats an injected `PORT`, and with no flag `PORT` is used. Matches the PR's 4-case verification table.

#### Suggestion
- **Software engineering / test coverage — `src/http/config.ts:16-44`:** No automated regression test was added for exactly the logic the author admits getting wrong on the first pass (the dead-code fallback "caught by testing it, not by reading it"). The repo already has vitest wired (`test: vitest run --passWithNoTests`, `package.json:19`) but no test dir, so a small unit test asserting config resolution — `PORT` set + no flag → that port; `--port` + `PORT` set → flag wins; neither → `3000`/`127.0.0.1` — would lock this behaviour against a future regression and cost little. Non-blocking, but this is a close-call bug that a test would have caught.
- **Functionality — build-step risk (acknowledged in PR body):** `dist/` is gitignored and there is no `prepare`/`postinstall`/`prestart` script, so `npm start` fails with "Cannot find module" if the platform only does `install` + `start` without a build. The author correctly scoped this out pending the next deploy log. One caveat on the proposed fix: `"prepare": "npm run build"` runs on **every** `npm install` — under `npm install --production`/`--omit=dev` (no devDeps) `tsup` is absent and `prepare` would itself fail. So the fix is less trivial than "just add prepare"; a `prestart`-guarded build or an explicit build step in the platform pipeline is safer. Non-blocking — just don't reach for the one-liner reflexively.

#### Information
- **Doc precision — `src/http/config.ts:39-42`:** The `host` comment says a container "opts in with `--host 0.0.0.0` or `HOST`", but the `start` script hardcodes `--host 0.0.0.0`, and a flag always overrides the env default — so on the `npm start` path the `HOST` env var is effectively dead. Harmless; the comment just slightly overstates the `HOST` path.
- **Security:** Binding `0.0.0.0` is intentional and necessary for the edge to reach the container; it's opted into explicitly via the script rather than by changing the default, keeping a locally-run binary on loopback. No new user input reaches SQL/shell/HTTP, no secrets newly logged, no auth changes, no new deps. No security-relevant concerns.
- **Performance:** No performance-relevant changes — env reads happen once at module load.

### Verdict
`approved` (informational — human approval still required via the approve flow). Zero Critical findings; the two Suggestions and the notes are non-blocking. This is a tight, evidence-driven fix that removes the actual cause of the deploy loop.

jwfing (commented):
## Review: Make the container actually able to start

**Summary:** Adds the missing `start` script and sources `--port`/`--host` defaults from `PORT`/`HOST`, which correctly addresses the root cause (`npm error Missing script: "start"`) and the two latent failures behind it.

### Requirements context
No spec/plan directory exists in this repo (`insforge-mcp` has no `docs/superpowers/` or equivalent) — assessing against the PR description and linked deployment log alone. The diagnosis in the body is well-supported and the change is scoped tightly to it (2 files, +21/-4).

I verified the fix end-to-end against the source:
- `tsup.config.ts` emits `http-server` from `src/http/server.ts`, so `dist/http-server.js` (the `start` target and existing `insforge-mcp-server` bin) is the correct built artifact.
- `src/http/server.ts:1086` calls `app.listen(SERVER_CONFIG.port, SERVER_CONFIG.host, …)`, so the resolved port/host flow into the actual bind.
- The commander default-vs-flag semantics the fix relies on (default applied only when the flag is absent; explicit flag wins) are stable and match the empirical table in the PR body.

### Findings

#### Critical
(none) — the fix is correct, the bind path is confirmed, and no existing behavior is broken (`PORT` unset + no flags still yields `127.0.0.1:3000`).

#### Suggestion
- **Software engineering — no automated test for the new resolution logic** (`src/http/config.ts:16-17,34,44`). Repo convention leans TDD, and this is exactly the kind of "reads correctly but behaves wrong" precedence bug the body describes catching by hand (the earlier dead-code fallback). It's untested against regression. The obstacle is real — `config.ts` runs `program.parse(process.argv)` at import (line 18), so it can't be exercised without a module refactor — but extracting a small pure `resolvePort(env, opts)` / `resolveHost(...)` helper would make the precedence table testable and lock in the four verified cases. Non-blocking given the manual 23/23 verification.
- **Functionality — build-step dependency is unverified and unguarded** (`package.json:14`). `start` runs `node dist/http-server.js`; there is no `prepare` script and no Dockerfile/deploy config in the repo. If the platform deploys from source and only runs `npm install && npm start` (no `build`), `dist/` won't exist and the *next* deploy fails with a missing-module error rather than a missing-script one. You've flagged this in "Still unverified" and deferred it deliberately, which is reasonable — noting it here so it's tracked: the known fix is `"prepare": "npm run build"`, and it's only unnecessary if the platform installs the published npm artifact (which ships `dist/` via `files`) or runs build itself.

#### Information
- **Functionality — `PORT=0` falls back to 3000** (`src/http/config.ts:34`). `parseInt("0") || 3000` → 3000, so the "OS-assigns-a-free-port" convention wouldn't work. Almost certainly irrelevant for a platform that injects a concrete port; noting only for completeness.
- **Documentation nuance — the `HOST` env branch is unreachable via `npm start`** (`src/http/config.ts:39-43`). Because `start` always passes `--host 0.0.0.0` explicitly (correct, per your rationale), the `process.env.HOST` default only applies to a bare `node dist/http-server.js` invocation. The docstring "A container opts in with `--host 0.0.0.0` or `HOST`" reads as if both paths are live under `npm start`; minor.
- **Pre-existing, out of scope — `publicUrl` is independent of the injected port** (`src/http/config.ts:47`). `SERVER_CONFIG.publicUrl` still defaults to `http://localhost:3000` regardless of `PORT`. A hosted deploy already must set `MCP_SERVER_URL` for OAuth/discovery URLs to be correct, so this PR doesn't regress anything — flagging only because it's adjacent to the hosted-deploy correctness this PR is about.

#### Security
No security-relevant regressions. Binding `0.0.0.0` is intentional and confined to the `start` script (containers); the default stays loopback, which is the safer choice for local binaries. No new user input reaches SQL/shell, no secrets are logged, and `publicUrl` (used to avoid host-header spoofing) is unchanged.

#### Performance
No concerns — module-load-time config resolution only; no hot-path, loop, or allocation impact.

### Verdict
**approved** (informational; a human still gives the explicit GitHub approval). Zero Critical findings — the change is correct and well-scoped. The two Suggestions (a regression test for the precedence logic; confirming/guarding the build step) are worth addressing but not blocking.

jwfing (approved):
LGTM - approved.

tonychang04:
## QA — verified, including the two ways this kind of change usually breaks

Built `c58d5d0` and ran it against a live Redis. 23/23 tests, lint 0 errors.

| case | result |
|---|---|
| `npm start` with `PORT=8099` injected | binds **8099** — the broken case, fixed |
| `npm start` with `PORT` unset | binds **3000** — fallback intact |
| `--port 3001` **and** `PORT=8099` set | binds **3001** — the explicit flag still wins |
| default, no `HOST`/`--host` | `127.0.0.1` only — stays loopback |
| `HOST=0.0.0.0` alone, no flag | `0.0.0.0` |
| stdio bin with `PORT`/`HOST` set | unaffected, `--version` fine |

The third row is the one worth calling out: sourcing a commander default from the environment is exactly where flag precedence usually gets inverted, and it doesn't here — the comment's claim holds under test.

The fifth and sixth matter for the same reason the comment gives: the local binary staying loopback-by-default is a real property, and this keeps it while letting a container opt in.

Confirms what I measured on `master` earlier — `PORT=8099` with no flag bound 3000 there, so a platform routing to its injected port got a healthy, unreachable container. That's the failure this removes, and it's the one that looks like nothing is wrong.

**Ship it.** Given the reboot loop this is currently causing on the Manufact trial, I'd merge this ahead of the rest of the queue — it touches nothing the other PRs touch.

*Not tested:* Manufact's own edge reaching the container (no access), and whether their platform injects `PORT` at all — if it doesn't, this is harmless either way.

— Quinn

PR #79, opened 2026-07-31
Client registrations as signed ids, not Redis rows

tonychang04:
First piece of the no-Redis rewrite, and the one that stands alone. **No behaviour change** — the module isn't wired into `server.ts` yet; that lands with the `/oauth/register` and `/oauth/authorize` changes.

## Why this one first

`mcp:oauth:client:*` is one of the five things keeping Redis on the critical path, and it's pure derived data: the only thing we ever learn from that row is what the client told us at registration. So carry it in the id — `mcp_<payload>.<signature>`.

A client id **is not a secret** (it travels in every authorize URL), so the property that matters isn't confidentiality but unforgeability. Anyone can read one; only we can mint one. An attacker can't invent an id claiming a `redirect_uri` we never approved.

**This is the mcp-use `oauthProxy` shape with the check that implementation omits.** `oauthProxy` never validates `redirect_uri` against the registration — that's what makes authorization-code theft possible there. Here the registration *is* the id, so there's nothing else to compare against and the check can't be skipped by accident. Per @manager's framing: when platform DCR opens, this gets deleted and replaced by a library call, not rewritten.

## Deliberate choices

- **No expiry.** The stored version expired after 30 days and silently broke every client installed that long — a bug we shipped and then fixed twice (#74, #75). A `redirect_uri` doesn't go stale. Rotating the signing secret invalidates every registration at once if we ever need that.
- **`readClientId` throws rather than returning null**, so a caller can't treat an unverified registration as valid by forgetting a check.
- **Length-compared before `timingSafeEqual`**, which throws on a length mismatch — that throw would itself be a crude oracle.
- **Exact-match `redirect_uris`** per RFC 6749 §3.1.2.3. No prefix or origin matching: those relaxations are what turn a redirect_uri check into an open redirect, and an open redirect on an authorize endpoint is code theft.

## Tests

13, including six redirect_uri near-misses — a path traversal, a suffix domain (`app.example.attacker.test`), a query-string append, a scheme-case variant.

**Confirmed they have teeth:** stubbing the signature comparison to always pass fails exactly the three forgery tests and nothing else.

```
× rejects a payload edited to claim another redirect_uri
× rejects an id signed with a different secret
× ids minted under one generated secret do not verify under another
```

58/58 across the suite, lint clean.

## Still to come on this branch

Wiring into `/oauth/register` (mint, store nothing) and `/oauth/authorize` (read the id, validate `redirect_uri` from it), plus the signing-secret config. Kept separate so this piece can be reviewed on its own security merits.

---
## Summary by cubic
Adds support for stateless OAuth client registrations using signed client IDs. No behavior change yet; this module will be wired into `/oauth/register` and `/oauth/authorize` next.

- **New Features**
  - Signed client ID format: `mcp_<payload>.<signature>` (HMAC-SHA256). Payload includes `redirect_uris`, optional `client_name`, and `iat`.
  - APIs: `mintClientId`, `readClientId` (throws on invalid), `isRegisteredRedirectUri` (exact match per RFC 6749 §3.1.2.3).
  - Security: exact `redirect_uri` matching, no expiry by design (rotate secret to revoke), constant-time verification with a length check, and requires a configured signing secret (no fallback).

- **Bug Fixes**
  - Removed the per-boot secret generator to avoid accidental mass invalidation on restart; `mintClientId` and `readClientId` now error without a supplied secret.

<sup>Written for commit 8cf53508678edc26bcedbbed45c7c1fe3170a784. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/79?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added self-contained OAuth client IDs with signed registration details.
  * Added support for validating client IDs, redirect URIs, expiration data, and signatures.
  * Added secure signing-secret generation.
* **Bug Fixes**
  * Improved handling of malformed, forged, or improperly configured client IDs.
* **Tests**
  * Added comprehensive coverage for client ID creation, validation, redirect matching, expiration, and signing-secret security.

### DIFF
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -3,7 +3,6 @@ import {
   mintClientId,
   readClientId,
   isRegisteredRedirectUri,
-  generateSigningSecret,
   InvalidClientIdError,
 } from './client-id.js';
 
@@ -107,16 +106,21 @@ describe('redirect_uri matching', () => {
   });
 });
 
-describe('generateSigningSecret', () => {
-  it('produces distinct high-entropy secrets', () => {
-    const a = generateSigningSecret();
-    const b = generateSigningSecret();
-    expect(a).not.toBe(b);
-    expect(a.length).toBeGreaterThanOrEqual(32);
+describe('a missing secret is never silently substituted', () => {
+  it('refuses to mint or read without one', () => {
+    // Quinn's catch: a per-boot generated fallback would invalidate every
+    // client id on every restart, handing every user the re-add page at once
+    // — a mass logout dressed up as a convenience. There is deliberately no
+    // way to get a working signer without supplying the secret, so wiring
+    // cannot reach for one by accident.
+    for (const missing of ['', undefined as unknown as string, null as unknown as string]) {
+      expect(() => mintClientId({ redirect_uris: [CALLBACK] }, missing)).toThrow();
+      expect(() => readClientId('mcp_a.b', missing)).toThrow();
+    }
   });
 
-  it('ids minted under one generated secret do not verify under another', () => {
-    const id = mintClientId({ redirect_uris: [CALLBACK] }, generateSigningSecret());
-    expect(() => readClientId(id, generateSigningSecret())).toThrow(InvalidClientIdError);
+  it('exports no secret generator', async () => {
+    const mod = await import('./client-id.js');
+    expect(Object.keys(mod)).not.toContain('generateSigningSecret');
   });
 });
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -1,0 +1,122 @@
+import { describe, it, expect } from 'vitest';
+import {
+  mintClientId,
+  readClientId,
+  isRegisteredRedirectUri,
+  generateSigningSecret,
+  InvalidClientIdError,
+} from './client-id.js';
+
+const SECRET = 'test-signing-secret';
+const OTHER_SECRET = 'a-different-signing-secret';
+const CALLBACK = 'http://127.0.0.1:8765/callback';
+
+describe('client ids carry their own registration', () => {
+  it('round-trips what the client registered', () => {
+    const id = mintClientId({ redirect_uris: [CALLBACK], client_name: 'Claude Code' }, SECRET);
+    const reg = readClientId(id, SECRET);
+
+    expect(reg.redirect_uris).toEqual([CALLBACK]);
+    expect(reg.client_name).toBe('Claude Code');
+    expect(typeof reg.iat).toBe('number');
+  });
+
+  it('is recognisable as ours', () => {
+    expect(mintClientId({ redirect_uris: [CALLBACK] }, SECRET).startsWith('mcp_')).toBe(true);
+  });
+
+  it('does not expire', () => {
+    // The stored version expired after 30 days and silently broke every client
+    // installed that long. A redirect_uri does not go stale.
+    const tenYearsAgo = Math.floor(Date.now() / 1000) - 10 * 365 * 24 * 60 * 60;
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, SECRET, tenYearsAgo);
+
+    expect(readClientId(id, SECRET).redirect_uris).toEqual([CALLBACK]);
+  });
+});
+
+describe('a client id cannot be forged', () => {
+  it('rejects a payload edited to claim another redirect_uri', () => {
+    // The attack this whole module exists to stop: change where the code goes.
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, SECRET);
+    const [payload, signature] = id.slice('mcp_'.length).split('.');
+    const tampered = Buffer.from(
+      JSON.stringify({ redirect_uris: ['https://attacker.example/steal'], iat: 1 })
+    ).toString('base64url');
+
+    expect(() => readClientId(`mcp_${tampered}.${signature}`, SECRET)).toThrow(InvalidClientIdError);
+    // and the original still verifies, so the test is about the edit, not the shape
+    expect(readClientId(`mcp_${payload}.${signature}`, SECRET).redirect_uris).toEqual([CALLBACK]);
+  });
+
+  it('rejects an id signed with a different secret', () => {
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, OTHER_SECRET);
+    expect(() => readClientId(id, SECRET)).toThrow(InvalidClientIdError);
+  });
+
+  it('rejects ids that are not ours at all', () => {
+    for (const bad of ['', 'mcp_', 'mcp_nodot', 'mcp_.sig', 'mcp_payload.', 'random-string']) {
+      expect(() => readClientId(bad, SECRET)).toThrow(InvalidClientIdError);
+    }
+  });
+
+  it('rejects a signed payload that is not a registration', () => {
+    // Signed by us, but a shape we never intended to mint.
+    const payload = Buffer.from(JSON.stringify({ redirect_uris: [] })).toString('base64url');
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, SECRET);
+    const signature = id.slice(id.lastIndexOf('.') + 1);
+    expect(() => readClientId(`mcp_${payload}.${signature}`, SECRET)).toThrow(InvalidClientIdError);
+  });
+
+  it('refuses to work without a secret', () => {
+    expect(() => mintClientId({ redirect_uris: [CALLBACK] }, '')).toThrow();
+    expect(() => readClientId('mcp_a.b', '')).toThrow();
+  });
+});
+
+describe('redirect_uri matching', () => {
+  const reg = readClientId(
+    mintClientId({ redirect_uris: [CALLBACK, 'https://app.example/cb'] }, SECRET),
+    SECRET
+  );
+
+  it('accepts a registered uri exactly', () => {
+    expect(isRegisteredRedirectUri(reg, CALLBACK)).toBe(true);
+    expect(isRegisteredRedirectUri(reg, 'https://app.example/cb')).toBe(true);
+  });
+
+  it('rejects prefix and origin near-misses', () => {
+    // These are the relaxations that turn a redirect_uri check into an open
+    // redirect, which on an authorize endpoint is code theft.
+    for (const near of [
+      'http://127.0.0.1:8765/callback/../evil',
+      'http://127.0.0.1:8765/callback2',
+      'http://127.0.0.1:8765',
+      'https://app.example/cb?next=https://attacker.example',
+      'https://app.example.attacker.test/cb',
+      'HTTP://127.0.0.1:8765/callback',
+    ]) {
+      expect(isRegisteredRedirectUri(reg, near)).toBe(false);
+    }
+  });
+
+  it('rejects non-strings', () => {
+    for (const bad of [undefined, null, 42, ['a'], {}]) {
+      expect(isRegisteredRedirectUri(reg, bad)).toBe(false);
+    }
+  });
+});
+
+describe('generateSigningSecret', () => {
+  it('produces distinct high-entropy secrets', () => {
+    const a = generateSigningSecret();
+    const b = generateSigningSecret();
+    expect(a).not.toBe(b);
+    expect(a.length).toBeGreaterThanOrEqual(32);
+  });
+
+  it('ids minted under one generated secret do not verify under another', () => {
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, generateSigningSecret());
+    expect(() => readClientId(id, generateSigningSecret())).toThrow(InvalidClientIdError);
+  });
+});
--- a/src/http/client-id.ts
+++ b/src/http/client-id.ts
@@ -1,4 +1,4 @@
-import { createHmac, randomBytes, timingSafeEqual } from 'crypto';
+import { createHmac, timingSafeEqual } from 'crypto';
 
 /**
  * Client registrations as signed values rather than stored rows.
@@ -126,9 +126,4 @@ export function isRegisteredRedirectUri(
   redirectUri: unknown
 ): boolean {
   return typeof redirectUri === 'string' && registration.redirect_uris.includes(redirectUri);
-}
-
-/** A signing secret for local runs, so the server starts without configuration. */
-export function generateSigningSecret(): string {
-  return randomBytes(32).toString('base64url');
 }
--- a/src/http/client-id.ts
+++ b/src/http/client-id.ts
@@ -1,0 +1,134 @@
+import { createHmac, randomBytes, timingSafeEqual } from 'crypto';
+
+/**
+ * Client registrations as signed values rather than stored rows.
+ *
+ * `/oauth/register` currently writes a Redis row and hands back a random id;
+ * `/oauth/authorize` reads that row back to learn the client's redirect_uris.
+ * That row is one of the five things keeping Redis on the critical path, and
+ * it is pure derived data — the only thing we learn from it is what the client
+ * already told us at registration.
+ *
+ * So carry it in the id. The id is `<payload>.<signature>`: the payload is the
+ * registration, the signature is ours. Anyone can read a client id (it travels
+ * in every authorize URL and is not a secret), but only we can mint one, which
+ * is the property that matters — an attacker cannot invent a client id that
+ * claims a redirect_uri we never approved.
+ *
+ * This is the mcp-use `oauthProxy` shape with the check that implementation
+ * omits: it never validates redirect_uri against the registration, which is
+ * what makes authorization-code theft possible there. Here the registration
+ * IS the id, so the check cannot be skipped by accident — there is nothing
+ * else to compare against.
+ */
+
+/** What a client told us about itself at registration. */
+export interface ClientRegistration {
+  redirect_uris: string[];
+  client_name?: string;
+  /** Seconds since epoch, for diagnostics. Not used for expiry — see below. */
+  iat: number;
+}
+
+export class InvalidClientIdError extends Error {}
+
+const PREFIX = 'mcp_';
+
+function sign(payload: string, secret: string): string {
+  return createHmac('sha256', secret).update(payload).digest('base64url');
+}
+
+/**
+ * Mint a client id that carries its own registration.
+ *
+ * Deliberately not time-limited. The stored version expired after 30 days and
+ * silently broke every client that had been installed that long — a bug we
+ * shipped and had to fix twice. A registration describes a redirect_uri, and a
+ * redirect_uri does not go stale; if we ever need to invalidate one, rotating
+ * the signing secret does it for all of them at once.
+ */
+export function mintClientId(
+  registration: Omit<ClientRegistration, 'iat'>,
+  secret: string,
+  now: number = Math.floor(Date.now() / 1000)
+): string {
+  if (!secret) {
+    throw new Error('a signing secret is required to mint client ids');
+  }
+  const full: ClientRegistration = { ...registration, iat: now };
+  const payload = Buffer.from(JSON.stringify(full)).toString('base64url');
+  return `${PREFIX}${payload}.${sign(payload, secret)}`;
+}
+
+/**
+ * Recover a registration from a client id, or throw.
+ *
+ * Throws rather than returning null so a caller cannot accidentally treat an
+ * unverified registration as valid by forgetting a null check — the failure
+ * mode this whole module exists to prevent.
+ */
+export function readClientId(clientId: string, secret: string): ClientRegistration {
+  if (!secret) {
+    throw new Error('a signing secret is required to read client ids');
+  }
+  if (typeof clientId !== 'string' || !clientId.startsWith(PREFIX)) {
+    throw new InvalidClientIdError('not a client id issued by this server');
+  }
+
+  const body = clientId.slice(PREFIX.length);
+  const dot = body.lastIndexOf('.');
+  if (dot <= 0 || dot === body.length - 1) {
+    throw new InvalidClientIdError('malformed client id');
+  }
+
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review — Client registrations as signed ids

**Summary:** A clean, security-focused standalone module (`src/http/client-id.ts` + tests) that encodes OAuth client registrations as HMAC-signed, self-verifying ids; it is not wired into `server.ts` yet, so there is no behavior change and no Critical findings.

### Requirements context
This repo has no `docs/superpowers/` (or any `docs/`) spec directory — assessing against the PR description alone plus the existing `/oauth/register` + `/oauth/authorize` implementation in `src/http/server.ts:210-352`. The PR body's stated intent (unforgeable, exact-match, non-expiring, throws-not-null) is implemented faithfully.

---

### Critical
(none)

### Suggestion

**Software engineering — mint/read validation asymmetry** (`src/http/client-id.ts:57-70` vs `95-110`)
`readClientId` rejects a payload whose `redirect_uris` is empty or contains a non-string (line 103-107), but `mintClientId` performs no such validation and will happily sign `{ redirect_uris: [] }` or `{ redirect_uris: [123] }`. The result is an id we minted ourselves that then fails to read back with `InvalidClientIdError`. The current `/oauth/register` guards `redirect_uris` upstream (`server.ts:219`), so this isn't exploitable today, but the invariant "anything mint produces, read accepts" is worth enforcing at the mint boundary (or asserting in a test) so a future caller can't create a permanently-unreadable id. Note the existing test `'rejects a signed payload that is not a registration'` only covers a hand-crafted empty array, not a mint-time one.

**Functionality — dropped RFC 7591 fields for the follow-up wiring** (`client-id.ts:25-31`, cf. `server.ts:233-258`)
The registration carried in the id is only `{ redirect_uris, client_name, iat }`. Today's `/oauth/register` also stores and echoes back `grant_types`, `response_types`, `token_endpoint_auth_method`, and `scope` in its RFC 7591 registration response. When this module replaces the Redis row, the register response must still return those fields (defaulted, since they're constant here) or DCR clients that read them will see a regression. Out of scope for this standalone PR, but flag it for the `/oauth/register` wiring commit so it isn't lost.

**Security — `client_name` is unvalidated and unbounded** (`client-id.ts:96`, `128`)
Only `redirect_uris` is validated on read; `client_name` is whatever was signed. Since `/oauth/register` will embed the client-supplied `client_name` verbatim, a client can inflate the id length arbitrarily. It's signed data so it's not a forgery/injection risk, but a length cap (and a type check on read) would keep ids bounded and avoid surprising downstream consumers.

### Information

- **Security (positive):** verification ordering is correct — the signature is checked *before* the base64url/JSON decode (`client-id.ts:82-91`), so untrusted input never reaches `JSON.parse` without a valid HMAC. Length-check-before-`timingSafeEqual` (line 88) correctly avoids the throw-as-oracle. Comparison is on the base64url-*encoded* signature strings, so there's no decode-canonicalization ambiguity. Exact-match `redirect_uri` (`includes`, line 128) with the six near-miss tests (path traversal, suffix domain, query append, scheme-case) is exactly right per RFC 6749 §3.1.2.3.
- **Security (minor DoS):** `readClientId` computes the HMAC over the full attacker-supplied payload before it can reject (`sign(payload, secret)` at line 84), so a large `client_id` forces an HMAC over its whole length. Bounded in practice by Express query/body size limits; worth a sanity length guard on the raw id if you want defense-in-depth when wiring `/oauth/authorize`.
- **Performance:** nothing of concern — synchronous crypto over tiny inputs, no loops/allocations of note, no I/O. Removing the Redis round-trip is a net win.
- **Conventions:** matches the repo — ESM `.js` import specifiers, `src/http/` placement alongside `redis.ts`/`config.ts`, `strict: true` compatible (null/primitive `parsed` is handled by the `!reg` guard at line 102). No new dependencies (uses built-in `crypto`).
- **Design (agreed):** no-expiry and throw-not-null are deliberate, documented, and well-justified given the #74/#75 history.

---

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings: the module is well-scoped, the security-critical paths (unforgeability, exact redirect_uri matching, constant-time verification, verify-before-parse) are correct and well-tested, and it introduces no behavior change on its own. The Suggestions are best addressed either here or explicitly carried into the `/oauth/register` + `/oauth/authorize` wiring commits.

jwfing (commented):
## Review: Client registrations as signed ids, not Redis rows

**Summary:** A well-designed, self-contained module that carries OAuth client registrations in an HMAC-signed client id instead of a Redis row; the security posture is sound and no blocking issues were found.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessed against the PR description alone. The PR body documents intent thoroughly (unforgeability over confidentiality, exact `redirect_uri` match per RFC 6749 §3.1.2.3, no expiry by design). I also cross-checked the current wiring: `/oauth/register` and `/oauth/authorize` in `src/http/server.ts` already do exact-match `redirect_uris.includes(redirect_uri)` (server.ts:348) and mint `mcp_`-prefixed ids (server.ts:227), so this module preserves existing observable behavior. As the PR states, it is **not yet wired into `server.ts`** — this review judges the module on its own merits.

I ran the gates against the head commit: `npm ci` + `vitest run src/http/client-id.test.ts` → **13/13 pass**, `tsc --noEmit` → no client-id errors.

---

### Findings

#### Critical
(none)

#### Suggestion

**1. Software engineering — a forgery test doesn't reach the branch it documents (`src/http/client-id.test.ts:76-82`).**
The test *"rejects a signed payload that is not a registration"* pairs an empty-`redirect_uris` payload with a signature taken from a *different* (CALLBACK) payload. `readClientId` therefore rejects it at the **signature** check, not the structural one — I verified it throws `"client id signature does not verify"`, not `"...is not a registration"`. As a result the empty-`redirect_uris` rejection branch (`client-id.ts:110-118`) has no real coverage, and the comment "Signed by us, but a shape we never intended to mint" is inaccurate (it isn't correctly signed). The branch itself is correct — a properly-signed empty registration *does* throw `"...is not a registration"` (verified). To actually exercise it, sign the empty payload with the real secret, e.g. `mintClientId({ redirect_uris: [] }, SECRET)` and assert the throw.

**2. Functionality — `mintClientId` doesn't validate its own input (`src/http/client-id.ts:44-58`).**
`mintClientId({ redirect_uris: [] }, SECRET)` mints successfully, but the resulting id can *never* be read back (`readClientId` rejects the empty array). This mint/read asymmetry is a latent foot-gun for the upcoming `/oauth/register` wiring. Additionally there is no bound on the count/length of `redirect_uris` or on `client_name` — because the full registration is base64url-encoded into a client id that then travels in **every authorize URL** (server.ts:369-379), an oversized `client_name` or many long `redirect_uris` yields an arbitrarily long `client_id` that can hit URL/header/proxy limits. Consider validating non-empty + bounding sizes at mint time (mirroring the `redirect_uris` non-empty check the register endpoint already does at server.ts:220). Low blast radius today since mint isn't wired in yet.

#### Information

- **Functionality — partial payload validation (`src/http/client-id.ts:100-118`).** Only `redirect_uris` is type-checked after parse; `client_name` and `iat` are trusted as-is. This is acceptable because the HMAC gates the payload to values we minted, but if the future `/oauth/register` passes an unvalidated `client_name` from client input, it will round-trip unchecked into the id. Worth a type/length guard when wiring.
- **Security (positive note).** The security-critical ordering is correct and subtle: signature is verified over the raw payload string **before** `JSON.parse` (so malformed/oversized JSON never reaches the parser unauthenticated), and the length comparison precedes `timingSafeEqual` (avoiding the throw-on-length-mismatch oracle called out in the PR body). Exact-match `redirect_uri` with the documented near-miss tests (path traversal, suffix-domain, query append, scheme-case) is the right call. `throw`-not-`null` on `readClientId` is a good fail-safe default. No secrets or PII are logged by this module.
- **Performance.** No concerns — a single HMAC per call, no I/O, no loops; this removes a Redis round-trip once wired. `generateSigningSecret` uses 32 bytes of `randomBytes`, consistent with `oauth-manager.ts:21`.

---

No Critical findings — the two Suggestions are non-blocking (a test that doesn't cover its named branch, and input-validation symmetry/bounds for the mint path). Posted as a COMMENT; explicit GitHub approval remains a separate human action.

jwfing (dismissed):
LGTM - approved.

tonychang04:
`8cf5350` — @qa-bot, thanks for the pass, and **I've taken your #2 as a defect rather than a wiring-time note.**

You were generous calling it a decision for later. I wrote `generateSigningSecret()` with the comment *"so the server starts without configuration"* — which is exactly the shape of footgun that gets reached for: a signer that works with no config, whose failure mode stays invisible until production restarts.

Shipped as a fallback, every restart would mint a new secret and invalidate every client id at once — every user hitting the re-add page simultaneously. A mass logout dressed up as a convenience, and structurally the same failure #74/#75 already fixed once when registrations expired on a timer.

**So it's gone rather than documented.** `mintClientId` and `readClientId` already throw without a secret; there's now no way to obtain a working signer without supplying one, so the wiring can't reach for a default by accident. Two tests pin it:

```
refuses to mint or read without one   ('' / undefined / null)
exports no secret generator
```

The local-run convenience belongs in configuration, where a missing secret can be a loud startup failure — not in the crypto module, which has no business deciding policy.

On your #1: agreed and stated in the original PR body — this is the primitive, not the switch-over. Redis is still on the path until `/oauth/register` and `/oauth/authorize` are rewired, which is the next commit.

58 tests, lint clean.

jwfing (commented):
## Review — Client registrations as signed ids

**Summary:** A tight, defensively-written stateless-client-id module (HMAC-SHA256 signed `mcp_<payload>.<signature>`) that faithfully reproduces the existing `/oauth/register` → `/oauth/authorize` redirect_uri contract without Redis; no Critical issues, a few non-blocking suggestions.

### Requirements context
No `docs/superpowers/` (or other spec dir) exists in `insforge-mcp` — assessing against the PR description and the existing OAuth code alone. I verified the intent against the code it will replace: current `/oauth/register` (`src/http/server.ts:210-253`) mints `mcp_<uuid>` and writes a `mcp:oauth:client:*` Redis row, and `/oauth/authorize` (`src/http/server.ts:293-351`) reads it back and validates `redirect_uri` via `clientData.redirect_uris.includes(redirect_uri)`. The new `isRegisteredRedirectUri` (`src/http/client-id.ts:124-129`) is the exact same exact-match semantics, so the redirect_uri contract is preserved.

**Verification performed:** `npm ci` + `npx vitest run src/http/client-id.test.ts` → **13/13 pass**; `tsc --noEmit` → no errors in this module; `eslint` → clean. Confirmed the module is imported **nowhere** outside its test, so the "no behaviour change / not wired into `server.ts` yet" claim holds.

---

### Critical
(none) — no correctness, security, or data-loss issues, and no existing behaviour is touched.

### Suggestion

**Functionality / security — mint/read validation asymmetry** (`src/http/client-id.ts:50-61` vs `:101-111`)
`readClientId` rejects a payload with empty/non-string `redirect_uris`, but `mintClientId` does no input validation — it spreads `registration` straight into the payload. So `mintClientId({ redirect_uris: [] }, secret)` (or non-string entries / a non-string `client_name`) happily produces an id that `readClientId` will then throw on. Not exploitable — the read path is the security gate and it holds — but the upcoming wiring will pass untrusted `req.body` into `mintClientId`. The existing endpoint validates `redirect_uris` first (`server.ts:220-224`), so blast radius is low today; mirroring `readClientId`'s guards at mint time would preserve the module's own "can't accidentally produce an invalid registration" property rather than relying on every caller to pre-validate. Non-blocking.

**Software engineering / security — no minimum signing-secret strength** (`src/http/client-id.ts:55`, `:71`)
Both functions reject only a falsy secret; a 1-character or low-entropy secret is accepted, and HMAC unforgeability is the entire security property here. The secret provisioning lands in a later PR per the description — worth enforcing a minimum length/entropy there (and failing fast at boot) so a weak secret can't silently undermine the scheme. Flagging now so it isn't lost between PRs.

**Software engineering — small test-coverage gaps** (`src/http/client-id.test.ts`)
Coverage is genuinely strong (round-trip, three forgery cases, six redirect_uri near-misses, missing-secret, no-generator). Two cheap additions would round it out: (1) a `client_name`-omitted value round-tripping through `readClientId` (currently only exercised through `mintClientId`), and (2) that `mintClientId` output for a normal registration is accepted end-to-end by `readClientId` for a non-loopback `https` uri.

### Information

**Security — recovered `iat`/`client_name` are not type-checked** (`src/http/client-id.ts:101-111`)
The post-verify structural check validates `redirect_uris` only; `iat` and `client_name` are trusted as-is via the `as ClientRegistration` cast. Safe today because only this server can sign a payload. Callout for the wiring PR: `client_name` originates from the client and, if later rendered on a consent page, must be escaped at render time — this module passes it through verbatim.

**Auto-generated PR summary is stale** — the CodeRabbit release-notes block claims "validating … expiration data" and "secure signing-secret generation", but the code deliberately does **neither** (no expiry by design; the per-boot secret generator was removed, with a test asserting `generateSigningSecret` is not exported). The cubic summary is accurate. Just make sure the inaccurate bot notes don't flow into real release notes.

**Performance (positive)** — no concerns. Replacing a Redis round-trip on `/authorize` with an in-process HMAC is a net win; the length-check-before-`timingSafeEqual` (`:88-90`), constant-time compare, and exact redirect_uri match are all correct, and verifying the signature *before* `JSON.parse` is the right order.

---

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings; the module is well-designed, well-tested, and passes all gates. The suggestions above are non-blocking and several naturally belong to the follow-up wiring/config PRs.

jwfing (commented):
## Review — Client registrations as signed ids (#79)

**Summary:** A clean, self-contained `client-id.ts` module that replaces the Redis-stored OAuth client registration with an unforgeable HMAC-signed id; the design is sound, the crypto ordering is correct, and coverage is strong — no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or equivalent) spec directory, so I assessed against the PR description, the no-Redis-rewrite framing in the body, and the current implementation being replaced (`src/http/server.ts:210-360`, `CLIENT_REGISTRATION_TTL` at `src/http/server.ts:89-92`). I verified the module is **not** wired into `server.ts` yet, matching the PR's "no behavior change" claim.

**Verification run locally at head `8cf5350`:** `vitest` 13/13 pass, `tsc --noEmit` clean, `eslint` clean. I also confirmed the new `isRegisteredRedirectUri` exact-match semantics match the current `clientData.redirect_uris.includes(redirect_uri)` check at `src/http/server.ts:348`, and the `mcp_` prefix is preserved from the current `randomUUID`-based id — so this is a faithful drop-in for the wiring PR, and it directly fixes the documented 30-day expiry bug.

---

### Critical
(none) — no correctness, security, or data-loss defects. The security-critical ordering is right: signature is verified **before** the payload is `JSON.parse`d (`src/http/client-id.ts:90` then `:97`), so untrusted bytes are never parsed; the constant-time compare is length-guarded (`src/http/client-id.ts:88-90`); `readClientId` throws rather than returning null; and there is no fallback signing secret. `lastIndexOf('.')` splitting is safe because base64url never contains `.`, and the empty-payload / empty-signature / no-dot cases are all rejected and tested.

### Suggestion

- **Software engineering — mint/read invariant asymmetry** (`src/http/client-id.ts:50-59`): `mintClientId` does no validation of its input, while `readClientId` enforces `redirect_uris` is a non-empty array of non-empty strings (`src/http/client-id.ts:100-108`). Minting `{ redirect_uris: [] }` therefore produces an id that `readClientId` will reject — an id that can't be read back. Consider failing fast at mint time so the invariant is enforced on both ends rather than only surfacing on read.
- **Functionality/security — redirect_uri scheme validation belongs at the mint boundary** (forward-looking, for the `/oauth/register` wiring): neither `mintClientId` nor the current `/oauth/register` (`src/http/server.ts:220`) validates that each `redirect_uri` is an absolute `http(s)` URL — a client could register `javascript:`/`data:`/relative values, which become a valid signed id. This is not a regression (current code has the same gap) and per OAuth DCR a client legitimately registers its own redirect, but the authorize endpoint redirects to whatever it reads back, so scheme/absoluteness validation should land with the wiring. Flagging here since the mint function is the natural choke point.

### Information

- **Security — no secret strength check** (`src/http/client-id.ts:55`, `:71`): `!secret` rejects only the empty string; a weak secret (`"x"`) passes. HMAC unforgeability rests entirely on the secret's entropy. The secret config lands in a later PR per the description — worth enforcing a minimum length there.
- **Security — `client_name` is carried unvalidated/unsanitized** (`src/http/client-id.ts:28`): fine as opaque data here, but if it is ever rendered on a consent/error screen it must be escaped at the render site to avoid stored-XSS. Note for the wiring/UI PR.
- **Performance — registration size is unbounded**: the entire payload becomes the `client_id` that travels in every authorize URL. When `/oauth/register` is wired, consider capping the count/length of `redirect_uris` to avoid oversized ids / mild amplification. No performance concern in this module itself — removing the Redis row actually drops a round-trip from the authorize path.

---

### Verdict
**approved** (informational — human approval is still a separate action). No Critical findings; the Suggestion and Information items are non-blocking, and most are notes for the follow-up wiring PRs rather than defects in this diff. Nicely scoped, well-tested, and the security rationale in the doc comments is a pleasure to review.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review: Client registrations as signed ids

**Summary:** A clean, well-tested, self-contained module that turns OAuth client registrations into HMAC-signed stateless client ids (`mcp_<payload>.<signature>`); no behaviour change since it isn't wired into `server.ts` yet. No blocking issues found.

### Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` (nor `docs/specs/`) directory, so I assessed against the PR description, the referenced regressions (#74/#75), and the surrounding OAuth code (`src/http/oauth-manager.ts`, `src/http/server.ts`). The design intent (unforgeable-but-public ids, exact `redirect_uri` match per RFC 6749 §3.1.2.3, no expiry, no per-boot secret fallback) is clearly stated in the PR body and the module's own doc comments, and the implementation matches it.

The module is deliberately **not** wired into `/oauth/register` or `/oauth/authorize` — that's called out as follow-up work on the same branch — so I did not treat the missing wiring or signing-secret config as gaps.

---

### Critical
(none)

The security-sensitive paths are correct:
- **Signature verified before `JSON.parse`** (`client-id.ts:97-104` then `:107`) — an unsigned/forged payload never reaches the parser.
- **Length check short-circuits before `timingSafeEqual`** (`client-id.ts:102`), avoiding the length-mismatch throw that would be a crude oracle; the leaked length is a public constant (43-char base64url of a 32-byte HMAC), so no secret leaks.
- **Signature binds the exact base64url payload string** (mint signs `payload`, read re-signs the same extracted `payload`) — canonical, no re-serialization/malleability gap.
- **Exact-match `redirect_uri`** via `Array.includes` (`client-id.ts:128`), no prefix/origin relaxation; the six near-miss tests (path traversal, suffix domain, query append, scheme-case) confirm it.
- **Secret is required with no fallback** on both mint and read; nothing is logged.

Prototype-pollution via a `__proto__` key is not reachable: only the secret-holder can mint, and `JSON.parse` assigns `__proto__` as an own property rather than mutating the prototype.

### Suggestion
- **Software engineering / test coverage — the structural-validation branch is not actually exercised.** `client-id.test.ts:63-69` ("rejects a signed payload that is not a registration") reuses a signature minted over the `[CALLBACK]` payload but pairs it with a *different* `{redirect_uris: []}` payload. `readClientId` therefore rejects at the signature check (`client-id.ts:102`, since `sign(empty) !== signature`) and never reaches the structural guard at `client-id.ts:108-116`. So the `redirect_uris` empty/non-array/non-string branch — the "signed by us but a shape we never intended" case the test claims to cover — has no test that reaches it. To genuinely exercise it, sign the malformed payload with `SECRET` directly so the signature passes and the structural check is what throws, e.g. `const payload = Buffer.from(JSON.stringify({ redirect_uris: [] })).toString('base64url'); const id = ` + "`mcp_${payload}.${sign(payload, SECRET)}`" + ` (using the internal signer). The code is correct; this is purely a coverage gap.

### Information
- **Forward-looking (functionality), for the wiring PR:** when `readClientId` replaces the Redis lookup in `/oauth/authorize`, its `InvalidClientIdError` throw must be mapped to the existing friendly "This connection needs to be set up again" page (`server.ts:295-317`), not a bare 500. This matters most right after a signing-secret rotation, when every previously-issued id stops verifying at once — exactly the mass re-add scenario the current handler is written to soften.
- **Defensive completeness (software engineering):** `readClientId` validates `redirect_uris` but not that `iat` is a number or `client_name` is a string (`client-id.ts:107-116`). Harmless because only secret-holders can mint, so the parsed shape is always what we serialized — noting only for symmetry with the interface contract.
- **Minor asymmetry (functionality):** `mintClientId` (`client-id.ts:53-63`) doesn't validate that `redirect_uris` is non-empty at mint time, while `readClientId` rejects empty on the way back. The caller (`/oauth/register`, `server.ts:220`) already enforces non-empty, so this is harmless; a symmetric guard would just fail faster.
- **Conventions:** matches repo style — `.js` ESM import extensions, native `crypto`, `vitest`, `throw`-not-`null`. Gates are `vitest` + `tsc` + `eslint` + `tsup`; PR reports 58/58 and lint clean, consistent with the code as read.

---

No Critical findings. One test-coverage Suggestion and a few Information notes, none blocking. Posted as a COMMENT — the explicit GitHub approval remains a separate human action.

PR #80, opened 2026-07-31
fix(http): preserve upstream client error statuses

tonychang04:
## Summary
- preserve InsForge API 4xx statuses in all seven HTTP catch paths
- keep upstream 5xx, Redis, and unknown failures mapped to 500
- add focused coverage for representative client errors and fallback failures

## Why
An upstream 404 was being rewritten as 500, causing MCP clients to retry non-retryable requests indefinitely.

## Verification
- `npm test` (69 passed)
- `npm run build`
- `npm run lint` (0 errors; 12 pre-existing warnings)

---
## Summary by cubic
Preserve upstream 4xx statuses from the InsForge API across all HTTP error paths so clients don’t retry non‑retryable requests. Session failures, upstream 5xx, and non-HTTP errors stay 500.

- **Bug Fixes**
  - Added `statusForHttpError` to pass through 4xx from `InsforgeApiError`; default 500 otherwise.
  - Applied helper to OAuth flows, project endpoints, and MCP session; explicitly keep session failures as server errors.
  - Added tests for representative 4xx and fallback-to-500 cases.

<sup>Written for commit 1f83511209aa833773492b207b4550520094e97e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/80?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved API error handling to return appropriate client error status codes instead of always reporting an internal server error.
  - Upstream server failures and unexpected errors continue to return HTTP 500 consistently.
  - Updated error responses across authentication, project, and streaming session workflows for more accurate client behavior.

- **Tests**
  - Added coverage for known client errors, upstream failures, and unexpected error conditions.

### DIFF
--- a/src/http/error-status.test.ts
+++ b/src/http/error-status.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from 'vitest';
+import { InsforgeApiError } from './insforge-api.js';
+import { statusForHttpError } from './error-status.js';
+
+describe('statusForHttpError', () => {
+  it.each([400, 401, 403, 404, 409, 429])(
+    'preserves an upstream %i client error',
+    (status) => {
+      expect(statusForHttpError(new InsforgeApiError('upstream rejected request', status)))
+        .toBe(status);
+    }
+  );
+
+  it.each([
+    new InsforgeApiError('upstream unavailable', 500),
+    new InsforgeApiError('upstream unavailable', 503),
+    new Error('Redis failed'),
+    'unknown failure',
+  ])('maps non-client failures to 500', (error) => {
+    expect(statusForHttpError(error)).toBe(500);
+  });
+});
--- a/src/http/error-status.ts
+++ b/src/http/error-status.ts
@@ -1,0 +1,19 @@
+import { InsforgeApiError } from './insforge-api.js';
+
+/**
+ * Preserve client-error statuses returned by the InsForge API.
+ *
+ * Treat upstream 5xx responses and non-HTTP failures as server errors: those
+ * are failures on our side of the MCP client's request and may be retryable.
+ */
+export function statusForHttpError(error: unknown): number {
+  if (
+    error instanceof InsforgeApiError &&
+    error.statusCode >= 400 &&
+    error.statusCode < 500
+  ) {
+    return error.statusCode;
+  }
+
+  return 500;
+}
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -26,6 +26,7 @@ import { renderProjectSelectionPage } from './templates/project-selection.js';
 import { renderOAuthErrorPage } from './templates/oauth-error.js';
 import { getAnalyticsService, extractClientInfo } from './analytics.js';
 import { sendUnauthorized, protectedResourceMetadata } from './auth-challenge.js';
+import { statusForHttpError } from './error-status.js';
 import { PACKAGE_VERSION } from '../shared/version.js';
 
 // ============================================================================
@@ -402,7 +403,7 @@ app.get(OAUTH_ENDPOINTS.authorize, async (req: Request, res: Response) => {
     res.redirect(authUrl.toString());
   } catch (error) {
     console.error('OAuth authorize error:', error);
-    res.status(500).json({
+    res.status(statusForHttpError(error)).json({
       error: 'server_error',
       error_description: 'Failed to initiate authorization',
     });
@@ -521,7 +522,7 @@ app.get(OAUTH_ENDPOINTS.callback, async (req: Request, res: Response) => {
       errorDescription: 'Failed to process callback',
       endpoint: '/oauth/callback',
     });
-    res.status(500).json({
+    res.status(statusForHttpError(error)).json({
       error: 'server_error',
       error_description: error instanceof Error ? error.message : 'Failed to process callback',
     });
@@ -561,7 +562,7 @@ app.get(OAUTH_ENDPOINTS.selectProject, async (req: Request, res: Response) => {
     }));
   } catch (error) {
     console.error('Project selection page error:', error);
-    res.status(500).send('Failed to load projects. Please try again.');
+    res.status(statusForHttpError(error)).send('Failed to load projects. Please try again.');
   }
 });
 
@@ -613,7 +614,7 @@ app.post(OAUTH_ENDPOINTS.selectProject, async (req: Request, res: Response) => {
     res.redirect(redirectUrl.toString());
   } catch (error) {
     console.error('Project selection error:', error);
-    res.status(500).json({
+    res.status(statusForHttpError(error)).json({
       error: 'server_error',
       error_description: error instanceof Error ? error.message : 'Failed to process project selection',
     });
@@ -736,7 +737,7 @@ app.get(API_ENDPOINTS.projects, async (req: Request, res: Response) => {
     res.json({ organizations: projects });
   } catch (error) {
     console.error('Get projects error:', error);
-    res.status(500).json({
+    res.status(statusForHttpError(error)).json({
       error: 'Failed to get projects',
       details: error instanceof Error ? error.message : 'Unknown error',
     });
@@ -770,7 +771,7 @@ app.post(API_ENDPOINTS.bindProject, async (req: Request, res: Response) => {
     });
   } catch (error) {
     console.error('Bind project error:', error);
-    res.status(500).json({
+    res.status(statusForHttpError(error)).json({
       error: 'Failed to bind project',
       details: error instanceof Error ? error.message : 'Unknown error',
     });
@@ -898,7 +899,7 @@ app.post(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
       });
     } catch (error) {
       console.error('[Streamable HTTP] Failed to create session:', error);
-      return res.status(500).json({
+      return res.status(statusForHttpError(error)).json({
         error: 'Failed to create session',
         details: error instanceof Error ? error.message : 'Unknown error',
       });
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -899,7 +899,7 @@ app.post(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
       });
     } catch (error) {
       console.error('[Streamable HTTP] Failed to create session:', error);
-      return res.status(statusForHttpError(error)).json({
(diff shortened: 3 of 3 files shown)


jwfing (commented):
## Review: fix(http): preserve upstream client error statuses

**Summary:** A tight, well-tested bugfix that routes seven HTTP catch paths through a new `statusForHttpError` helper so upstream InsForge API 4xx statuses pass through instead of being flattened to 500 — resolving the "clients retry non-retryable requests forever" problem.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the `InsforgeApiError` contract, and surrounding handler code.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — OAuth error-code / status-code mismatch** (`src/http/server.ts:406`, `:525`). The `/oauth/authorize` and `/oauth/callback` catch blocks return a body of `{ error: 'server_error', ... }` but will now emit a 4xx status when the underlying `InsforgeApiError` is a client error (e.g. a failed token exchange surfacing upstream 401). Per RFC 6749, `server_error` semantically pairs with 5xx. The change is still a net improvement over the old blanket-500, but consider mapping the OAuth `error` field to match (e.g. `invalid_request`/`access_denied`) when passing through a 4xx, or scoping the helper to the non-OAuth JSON/API handlers where a raw upstream status is unambiguous. Non-blocking.

- **Software engineering — no handler-level assertion that the helper is wired in** (`src/http/error-status.test.ts`). The new test exercises `statusForHttpError` directly and thoroughly (4xx pass-through incl. 429, 5xx→500, non-HTTP `Error`→500, string→500 — good coverage). But nothing guards against a regression where a handler silently reverts to a hardcoded `res.status(500)`. A single lightweight integration/route test (e.g. force `getAvailableProjects` to throw `InsforgeApiError(…, 404)` and assert `GET /api/projects` responds 404) would lock in the behavior the PR actually ships. Optional given the helper is well unit-tested.

#### Information

- **Functionality — remaining hardcoded 500s are correct, not misses.** I verified `src/http/server.ts:284/340/349/357` (OAuth misconfig, corrupt/unparseable Redis client data), `:819` (Redis session-restore returned null), and `:1104` (SSE `createSSESession` failure) are genuine server-side faults, not upstream-status paths, so leaving them at 500 is right. `resolveProjectFromToken` (`oauth-manager.ts:341`) only reads Redis bindings and returns `null` on miss — it never throws `InsforgeApiError` — so the pre-`try` calls at `:825`/`:1005` aren't upstream-status paths either. The seven converted catches cover every handler that actually calls the InsForge API.

- **Security — no new exposure.** Passing the upstream status code through leaks nothing sensitive. The `error.message` echoing in several bodies (which can include upstream response text) is pre-existing and unchanged by this PR. No auth/authorization check is weakened, no new dependency, no new secret logging.

- **Performance — none.** `statusForHttpError` is a pure branch on an already-materialized error; zero added I/O or allocation in any hot path.

- **Conventions** all met: co-located `*.test.ts` (matches `auth-challenge.test.ts`/`redis.test.ts`), ESM `.js` import specifiers, named export, `instanceof InsforgeApiError` narrowing consistent with existing code.

---

### Verdict
**approved** (informational — human approval still required via the approve flow). Zero Critical findings; the two Suggestions are non-blocking polish. Clean, in-scope, well-tested fix.

jwfing (commented):
## Summary
A small, well-scoped fix that stops rewriting upstream InsForge 4xx responses as 500 across the HTTP error paths, so MCP clients no longer retry non-retryable requests — clean helper, good test, correct default-to-500 fallback.

## Requirements context
No matching spec/plan found — this repo has no `docs/superpowers/` (or `docs/specs/`) directory, so I assessed against the PR description and the surrounding code alone. Intent (preserve upstream 4xx, keep 5xx/Redis/unknown as 500) is clear and self-consistent.

## Findings

### Critical
(none)

**Software engineering** — no blocking issues. `statusForHttpError` is a pure, exported helper (`src/http/error-status.ts:9-18`) with a focused unit test covering both branches. `InsforgeApiError.statusCode` (`src/http/insforge-api.ts:60-69`) matches the field the helper reads, and ESM `./*.js` import extensions follow repo convention.

**Functionality** — the change is correct where it matters. The upstream 4xx sources sit inside the seven updated try-blocks (`getAllUserProjects`, `getProjectAccess`, `bindTokenToProject`, `validateToken` via the OAuth/project endpoints), so those now propagate the real status. Where the try-block has no upstream call, the helper safely defaults to 500, so the change is inert rather than wrong. I specifically checked that `resolveProjectFromToken` (`src/http/server.ts:825`, `:1005`) only reads Redis token bindings and never throws `InsforgeApiError`, so it is *not* a missed upstream path — the fact that it sits outside the updated catch is fine. 429 being preserved (test at `error-status.test.ts:7`) is the correct behavior since 429 is legitimately retryable and clients key off the status.

**Security** — no security-relevant changes. Only the HTTP status code is altered; response bodies (including the pre-existing `details: error.message` echoes) are untouched, so no new information is exposed. Auth/authorization checks are unchanged.

**Performance** — no performance-relevant changes. The helper is an O(1) `instanceof` + range check on the error path only; no new I/O, loops, or allocations.

### Suggestion
- **functionality — OAuth error-code semantics** (`src/http/server.ts:406`, `:525`, `:617`): the OAuth endpoints now return a 4xx status while the body still carries `error: 'server_error'`. Per RFC 6749, `server_error` connotes a 5xx-class condition, so a client that keys retry decisions off the OAuth *error code* (rather than HTTP status) could still misclassify these. Low blast radius, and it doesn't undermine the HTTP-status fix, but consider mapping the body's `error` field alongside the status in a follow-up.
- **software engineering — wiring coverage** (`src/http/error-status.test.ts`): the test exercises the helper in isolation but not that the seven call sites actually invoke it. The wiring is trivial (mechanical `res.status(500)` → `res.status(statusForHttpError(error))` substitution), so this is low value, but a single integration-style assertion that, e.g., `GET /api/projects` returns 404 when the upstream returns 404 would guard against future regressions on the hot path named in the bug.

### Information
- **doc comment wording** (`src/http/error-status.ts:5-7`): "failures on our side of the MCP client's request" reads a little ambiguously; "failures on the MCP server's side (not the client's fault), which may be retryable" would be clearer. Cosmetic.
- **deprecated SSE path** (`src/http/server.ts:1084`, `:1104`): the legacy SSE session-creation catch still hardcodes 500. This is consistent, not a miss — its upstream project resolution goes through the Redis-only `resolveProjectFromToken`, so no `InsforgeApiError` flows here. Noting only so it isn't mistaken for an oversight.

## Verdict
**approved** (informational — human approval via the approve flow still required). No Critical findings: the change is correct, appropriately scoped, follows repo conventions, and is covered by a focused test. The Suggestions are non-blocking polish.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Summary
Small, well-scoped fix that stops the HTTP layer from rewriting upstream InsForge API 4xx responses as 500 by routing every catch path through a new `statusForHttpError` helper; the change is correct, tested, and I found no blocking issues.

## Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` or `/docs/specs/` directory. Assessed against the PR description/intent alone: preserve upstream 4xx statuses so MCP clients don't retry non-retryable requests, while keeping upstream 5xx, Redis, and unknown failures at 500.

## Findings

### Critical
(none)

### Suggestion
- **Software engineering / test coverage** — The helper is thoroughly unit-tested in isolation (`src/http/error-status.test.ts:1-22` covers 400/401/403/404/409/429 passthrough, 500/503, a generic `Error`, and a non-`Error` string), which is good. What's missing is an endpoint-level test asserting the fix actually propagates end-to-end — e.g. `GET /api/projects` returning 401 when `validateToken`/`getAllUserProjects` throws an `InsforgeApiError(…, 401)`. The whole point of the change is status propagation through the route handlers (`src/http/server.ts:740`, `:774`), so a thin integration test would guard against a future regression where someone re-wraps or swallows the error. Non-blocking given the isolated helper is covered.

### Information
- **Functionality / consistency** — Three of the six wrapped catch blocks emit a body labeled `{ error: 'server_error', … }` yet can now send an HTTP 4xx status: `authorize` (`src/http/server.ts:406`), `callback` (`:525`), and `selectProject` POST (`:617`). That's a mild semantic mismatch (4xx status + `server_error` code). In practice these paths operate on Redis / OAuth-state operations (`createAuthorizationState`, `getAuthorizationState`, `createAuthorizationCode`) and won't normally surface an `InsforgeApiError`, so the effective behavior there is unchanged (still 500). The two paths that genuinely benefit are `GET /api/projects` (`:740`) and `POST …/bind` (`:774`), whose try-blocks call the InsForge API via `getAvailableProjects`/`bindTokenToProject` and legitimately throw `InsforgeApiError` with the upstream status — verified in `src/http/oauth-manager.ts:383-427` and `src/http/insforge-api.ts:83-89`.
- **Functionality / edge** — `statusForHttpError` maps any non-4xx `InsforgeApiError` (including a hypothetical 3xx) to 500 (`src/http/error-status.ts:10-18`). Correct for this API (only 2xx/4xx/5xx are expected); noted only for completeness.
- **Description nit** — The PR body says "all seven HTTP catch paths"; the diff modifies six. The seventh appears to be a session-creation path deliberately kept at 500 (`src/http/server.ts:819`, `:902`, `:1104`), which are unchanged — consistent with the intent, just worth reconciling the wording.

### Security & Performance
- **Security** — No new user input reaches SQL/shell/HTTP; no new secrets or PII logged. The passthrough of upstream 4xx does not widen any information leak — the `details: error.message` / `error_description` fields already returned the upstream error text before this PR; only the numeric status changed. If anything, returning accurate 401/403 is more correct than masking as 500. No auth checks weakened, no new dependencies.
- **Performance** — Pure branching helper; no new queries, loops, allocations, or blocking I/O.

## Verdict
`approved` (informational — human still approves via the approve flow). Clean, correctly-scoped bug fix with adequate unit coverage; the items above are non-blocking polish.

PR #81, opened 2026-07-31
fix(http): normalise publicUrl at the source — missing from merged #69

tonychang04:
## Why this is a separate PR

These two commits were on `fix/oauth-www-authenticate`, but **PR #69 merged at
head `ae0e438` — two commits behind the branch**, so they are not on `master`:

```
origin/fix/oauth-www-authenticate  a7dc73e
#69 merged at                      ae0e438
missing:  d29b115  Normalise publicUrl at the source, not at two consumers
          a7dc73e  Assert on the advertised URLs, and forbid the bypass
```

Confirmed against `master`: `normaliseBaseUrl` is absent, and
`publicUrl: process.env.MCP_SERVER_URL || 'http://localhost:3000'` is unguarded.

## What breaks without it

A configured `MCP_SERVER_URL` ending in `/` gives a doubled slash to every URL
we advertise, and Express matches none of those routes. There are ten
consumers; four are functional (Quinn's narrowing) and each breaks the login at
a different step: the AS metadata document, the OAuth callback we register with
the platform, the project-selection redirect, and the authorize links in 401
bodies. It fails silently — the client discards the metadata per RFC 9728 §3.3
rather than reporting a mismatch.

This matters right now because `MCP_SERVER_URL` is being set on Manufact by
hand for `https://mcp.insforge.dev`, and a trailing slash there is one keystroke.

## The fix

Normalise once where `publicUrl` is defined, so every consumer is safe by
construction — not per-consumer patching, which had already missed two of the
four.

## Tests

- `normaliseBaseUrl` unit test — covers the config-level fix. **Verified by
  mutation:** neutering `normaliseBaseUrl` to `return url` fails it
  (`1 failed | 18 passed`), restoring it passes.
- Three assertions over the *advertised* URLs given a slashed public URL
  (protected-resource documents, derived metadata URL, `resource_metadata` in
  the challenge). Doubled slashes are checked after the scheme so `https://` is
  not a false hit.
- A guard forbidding any file outside `config.ts` from reading
  `process.env.MCP_SERVER_URL` — the only remaining way to reintroduce this.

**Honest coverage note:** the three advertised-URL tests pass a slashed base
explicitly, so they exercise `origin()` in `auth-challenge.ts`, not the config
normalisation. The config fix is covered by the `normaliseBaseUrl` test and the
bypass guard. Stating that rather than letting three green tests imply it.

74 tests pass on this branch.

---
## Summary by cubic
Normalizes the server `publicUrl` at the source by stripping trailing slashes from `MCP_SERVER_URL`, preventing `//` in advertised URLs and fixing silent login failures (AS metadata, OAuth callback, redirects, authorize links).

- **Bug Fixes**
  - Added `normaliseBaseUrl()` and applied it to `SERVER_CONFIG.publicUrl` in `src/http/config.ts`.
  - Tests assert no doubled slashes in advertised URLs (protected-resource docs, derived metadata URL, `WWW-Authenticate` challenge).
  - Strengthened guard: only `config.ts` may mention `MCP_SERVER_URL` (catches dot, bracket, and destructuring forms; anchored file walk; asserts it scans real files).

<sup>Written for commit e54c08d96a8f14d9946487ad41bbc1150af4eb9e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/81?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/auth-challenge.test.ts
+++ b/src/http/auth-challenge.test.ts
@@ -149,3 +149,67 @@ describe('normaliseBaseUrl', () => {
     expect(normaliseBaseUrl('https://host/base/')).toBe('https://host/base');
   });
 });
+
+describe('no advertised URL can contain a doubled slash', () => {
+  // Quinn's ask: assert on the URLs we advertise, not on the helper that
+  // builds them. A trailing slash on MCP_SERVER_URL used to break the login
+  // in at least three separate places, at three different steps — the AS
+  // metadata, the project-selection redirect, and the authorize links in 401
+  // bodies. Per-consumer patching missed two of them.
+  const SLASHED = 'https://mcp.insforge.dev/';
+
+  /** Everything after the scheme, so `https://` itself is not a false hit. */
+  const hasDoubledSlash = (url: string) => url.replace(/^[a-z]+:\/\//, '').includes('//');
+
+  it('not in the protected-resource documents', () => {
+    for (const path of ['', '/mcp', '/sse']) {
+      const doc = protectedResourceMetadata(path, SLASHED);
+      expect(hasDoubledSlash(String(doc.resource))).toBe(false);
+      for (const as of doc.authorization_servers as string[]) {
+        expect(hasDoubledSlash(as)).toBe(false);
+      }
+    }
+  });
+
+  it('not in the metadata URL a client derives', () => {
+    for (const path of ['', '/mcp', '/sse']) {
+      expect(hasDoubledSlash(protectedResourceMetadataUrl(path, SLASHED))).toBe(false);
+    }
+  });
+
+  it('not in the WWW-Authenticate challenge', () => {
+    for (const path of ['', '/mcp', '/sse']) {
+      const challenge = buildResourceChallenge(path, SLASHED);
+      const named = challenge.match(/resource_metadata="([^"]+)"/)?.[1] ?? '';
+      expect(named).not.toBe('');
+      expect(hasDoubledSlash(named)).toBe(false);
+    }
+  });
+});
+
+describe('nothing bypasses the normalisation', () => {
+  it('only config.ts reads MCP_SERVER_URL', async () => {
+    // publicUrl is normalised where it is defined, so every consumer is safe
+    // by construction. The one way to reintroduce the bug is to read the raw
+    // environment variable somewhere else, so that is what this forbids.
+    const { readdirSync, readFileSync, statSync } = await import('node:fs');
+    const { join } = await import('node:path');
+
+    const offenders: string[] = [];
+    const walk = (dir: string) => {
+      for (const entry of readdirSync(dir)) {
+        const full = join(dir, entry);
+        if (statSync(full).isDirectory()) {
+          walk(full);
+          continue;
+        }
+        if (!full.endsWith('.ts') || full.endsWith('.test.ts')) continue;
+        if (full.endsWith('http/config.ts')) continue;
+        if (/process\.env\.MCP_SERVER_URL/.test(readFileSync(full, 'utf8'))) offenders.push(full);
+      }
+    };
+    walk('src');
+
+    expect(offenders).toEqual([]);
+  });
+});
--- a/src/http/auth-challenge.test.ts
+++ b/src/http/auth-challenge.test.ts
@@ -187,29 +187,74 @@ describe('no advertised URL can contain a doubled slash', () => {
   });
 });
 
-describe('nothing bypasses the normalisation', () => {
-  it('only config.ts reads MCP_SERVER_URL', async () => {
-    // publicUrl is normalised where it is defined, so every consumer is safe
-    // by construction. The one way to reintroduce the bug is to read the raw
-    // environment variable somewhere else, so that is what this forbids.
-    const { readdirSync, readFileSync, statSync } = await import('node:fs');
-    const { join } = await import('node:path');
+/** Every non-test .ts under src, by absolute path. */
+async function sourceFiles(): Promise<string[]> {
+  const { readdirSync, statSync } = await import('node:fs');
+  const { join, dirname, resolve } = await import('node:path');
+  const { fileURLToPath } = await import('node:url');
 
-    const offenders: string[] = [];
-    const walk = (dir: string) => {
-      for (const entry of readdirSync(dir)) {
-        const full = join(dir, entry);
-        if (statSync(full).isDirectory()) {
-          walk(full);
-          continue;
-        }
-        if (!full.endsWith('.ts') || full.endsWith('.test.ts')) continue;
-        if (full.endsWith('http/config.ts')) continue;
-        if (/process\.env\.MCP_SERVER_URL/.test(readFileSync(full, 'utf8'))) offenders.push(full);
+  // Anchored to this file, not to process.cwd(). It happens to be the project
+  // root under vitest today, which is a property of how the tests are launched
+  // rather than of the tree being checked — and a guard that silently walks an
+  // empty directory passes.
+  const root = resolve(dirname(fileURLToPath(import.meta.url)), '..');
+
+  const found: string[] = [];
+  const walk = (dir: string) => {
+    for (const entry of readdirSync(dir)) {
+      const full = join(dir, entry);
+      if (statSync(full).isDirectory()) {
+        walk(full);
+        continue;
       }
-    };
-    walk('src');
+      if (full.endsWith('.ts') && !full.endsWith('.test.ts')) found.push(full);
+    }
+  };
+  walk(root);
+  return found;
+}
+
+describe('nothing bypasses the normalisation', () => {
+  it('only config.ts mentions MCP_SERVER_URL at all', async () => {
+    // publicUrl is normalised where it is defined, so every consumer is safe by
+    // construction, and the one way to reintroduce the bug is to read the raw
+    // environment variable somewhere else.
+    //
+    // This forbids the NAME, not one syntax for reading it. The previous
+    // version matched /process\.env\.MCP_SERVER_URL/, so
+    // process.env['MCP_SERVER_URL'] and const { MCP_SERVER_URL } = process.env
+    // both walked straight past — a guard that does not guard, which is worse
+    // than none because it is counted as coverage. Nothing outside config.ts
+    // has any business naming the variable, so the name is the right thing to
+    // forbid and it has no syntax blind spots.
+    const { readFileSync } = await import('node:fs');
+
+    const offenders = (await sourceFiles()).filter(
+      (file) => !file.endsWith('http/config.ts') && readFileSync(file, 'utf8').includes('MCP_SERVER_URL')
+    );
 
     expect(offenders).toEqual([]);
+  });
+
+  it('is actually looking at files', async () => {
+    // The guard above passes vacuously if the walk finds nothing — which is
+    // exactly how a gate scores green without reaching the code it grades.
+    const files = await sourceFiles();
+    expect(files.length).toBeGreaterThan(5);
+    expect(files.some((f) => f.endsWith('http/config.ts'))).toBe(true);
+    expect(files.some((f) => f.endsWith('http/server.ts'))).toBe(true);
+  });
+
+  it('catches every way of naming the variable', async () => {
+    // The three forms that must all be caught, checked against the same
+    // predicate the guard uses rather than described in a comment.
+    const bypasses = [
+      "const u = process.env.MCP_SERVER_URL;",
+      "const u = process.env['MCP_SERVER_URL'];",
+      'const { MCP_SERVER_URL } = process.env;',
+      'const k = "MCP_SERVER" + "_URL"; process.env[k];', // the one it cannot catch
+    ];
+    const caught = bypasses.map((src) => src.includes('MCP_SERVER_URL'));
+    expect(caught).toEqual([true, true, true, false]);
   });
 });
--- a/src/http/auth-challenge.test.ts
+++ b/src/http/auth-challenge.test.ts
@@ -132,3 +132,20 @@ describe('sendUnauthorized', () => {
     expect(order).toEqual(['header', 'body']);
   });
 });
+
+describe('normaliseBaseUrl', () => {
+  it('strips trailing slashes so no consumer has to', async () => {
+    // Guarding the two consumers I first noticed left eight others — the AS
+    // metadata, the OAuth callback registered with the platform, the
+    // project-selection redirect, two authorize URLs in error bodies, the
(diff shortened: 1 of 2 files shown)


jwfing (commented):
## Summary
Normalises `publicUrl` at its definition in `config.ts` by stripping trailing slashes, so every one of the ~10 consumers of `SERVER_CONFIG.publicUrl` is safe-by-construction against a doubled-slash `MCP_SERVER_URL` — a clean, well-tested fix that closes the gap left when #69 merged two commits behind the branch.

## Requirements context
`insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) spec directory — confirmed empty. Assessing against the PR description, the referenced #69 history, and the surrounding code alone. The PR's stated intent (normalise once at the source, forbid re-introduction) is coherent and matches the implementation.

## Findings

### Critical
(none)

### Suggestion
- **[Software engineering — DRY] `src/http/auth-challenge.ts:10-12`** — After this fix, `origin()` and `normaliseBaseUrl()` are two independent copies of the same "strip trailing slash" logic (`/\/+$/`). The defense-in-depth in `auth-challenge.ts` is worth keeping (it's what makes the advertised-URL tests exercise a slashed base directly, per your honest coverage note), but `origin()` could `import { normaliseBaseUrl } from './config.js'` and delegate to it, keeping a single definition while preserving the pure, independently-testable signature. Non-blocking.

### Information
- **[Software engineering] `src/http/auth-challenge.test.ts:203-208`** — The bypass guard is a nice touch and correctly excludes itself (`.test.ts`) and `config.ts`, so no false positives. Two blind spots worth noting for the record: the regex `/process\.env\.MCP_SERVER_URL/` won't catch a destructured read (`const { MCP_SERVER_URL } = process.env`) or bracket access (`process.env['MCP_SERVER_URL']`). It's a best-effort tripwire, not an airtight guarantee — fine as-is, just don't over-trust it.
- **[Software engineering] `src/http/auth-challenge.test.ts:200`** — `walk('src')` resolves relative to the test process CWD. That's the project root under vitest so it works, but it's a slight coupling to how tests are launched; an absolute path derived from the test file location would be more robust.
- **[Functionality] `src/http/config.ts:37-39`** — `normaliseBaseUrl` strips only trailing `/`, not surrounding whitespace or a trailing `/` before a query/fragment. That's the right scope for a base-URL env value and I wouldn't expand it — noting only that the contract is "trailing slashes," nothing more.

### Dimension coverage
- **Security** — No security-relevant regressions. `publicUrl` is operator-supplied env, never request-derived, so no new untrusted input reaches URL construction; continuing to use `publicUrl` as the canonical base (vs. `Host` header) is the anti-spoofing-correct choice. No secrets logged, no auth checks touched, no new dependencies.
- **Performance** — One regex evaluated once at module load. No hot-path, allocation, or query impact.

## Verdict
`approved` (informational — posted as a COMMENT; a human still gives the GitHub approval). No blocking issues; the Suggestion and Information notes are all optional polish. The fix is correct, complete across all consumers, and backed by a mutation-verified test plus a re-introduction guard.

jwfing (approved):
LGTM - approved.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review — fix(http): normalise publicUrl at the source

**Summary:** A tightly-scoped, correct bugfix that strips trailing slashes from `MCP_SERVER_URL` once at the point `SERVER_CONFIG.publicUrl` is defined, so all ~10 consumers become slash-safe by construction rather than per-consumer patching.

### Requirements context
`InsForge/insforge-mcp` has **no `/docs/superpowers/` (or any `docs/`) spec directory** — assessed against the PR description and linked context (#69). Gates are `vitest` / `tsc` / `eslint` / `tsup` build. The PR body's central claim is verified against the checked-out head: `master` had `publicUrl: process.env.MCP_SERVER_URL || 'http://localhost:3000'` unguarded, and this PR wraps it in `normaliseBaseUrl(...)`.

### Verification performed
- Confirmed every consumer concatenates `SERVER_CONFIG.publicUrl` with a leading-`/` path — `config.ts:92` (`callbackUrl`), `server.ts:517` (select-project redirect), `server.ts:832/840` (`oauth_authorize_url`), `server.ts:841` (`projects_url`), `server.ts:1165-1202` (startup banner). A trailing slash on the base would have doubled the slash in all of them; normalising at source fixes each. ✅
- Confirmed `grep MCP_SERVER_URL src/` finds the raw env read only in `config.ts:62`, so the "only config.ts reads it" bypass-guard test genuinely passes on this branch. ✅
- Confirmed the three advertised-URL tests route through `origin()` in `auth-challenge.ts:10`, exactly as the PR's honest-coverage note states — the config-level fix is covered instead by the `normaliseBaseUrl` unit test + the bypass guard. ✅

### Findings

#### Critical
(none) — no correctness, security, data-loss, or requirements violations found.

#### Suggestion
- **[functionality] `normaliseBaseUrl` trims trailing slashes but not surrounding whitespace** (`src/http/config.ts:37-39`). The motivating case is a hand-set env var on Manufact; env values pasted or heredoc'd into a deploy commonly carry a trailing `\n`/space, which would survive as `https://host/mcp\n` and break route matching the same way the slash did. A `url.trim().replace(/\/+$/, '')` would close the adjacent failure mode at no cost. Non-blocking.

#### Information
- **[software engineering] Duplicated normalisation logic** — `normaliseBaseUrl` (`config.ts:37`) and `origin()` (`auth-challenge.ts:10-12`) are now byte-identical (`replace(/\/+$/, '')`). Now that `publicUrl` is normalised at source, `origin()` is redundant for the default path and only does work when a caller passes an explicitly un-normalised base (which the new tests do deliberately). Fine as defensive belt-and-suspenders, but consider having `auth-challenge.ts` import and reuse `normaliseBaseUrl` so the two can't drift.
- **[software engineering] Bypass-guard test is text/cwd-sensitive** (`src/http/auth-challenge.test.ts:191-213`). It `walk('src')` relative to cwd (fine under vitest at repo root) and regex-matches raw file text, so a future *comment* or string mentioning `process.env.MCP_SERVER_URL` in any non-`config.ts` file would fail it spuriously. Low blast radius and arguably a feature (it forces a conscious exception), just noting the sharp edge.

### Dimension coverage
- **Software engineering:** Good. New behavior has a mutation-verified unit test, advertised-URL assertions, and an architectural guard test; the PR is refreshingly honest about what each test does and does not cover. Follows repo conventions (`.js` import specifiers, exported pure helper, JSDoc-heavy style).
- **Functionality:** Solves the stated bug; source-level normalisation covers all consumers including the four the PR calls out. One adjacent edge (whitespace) noted as a Suggestion.
- **Security:** No security-relevant regressions. No secrets/PII logged or returned; no auth/authz weakened; no new dependencies. Arguably a mild correctness-of-discovery improvement for the OAuth flow.
- **Performance:** No concern — one regex at module load.

### Verdict
**approved** (informational; no Critical findings). Posted as a COMMENT — the GitHub green-check approval remains a separate human action.

jwfing (commented):
## Summary

Normalises `MCP_SERVER_URL` once at the config source (`SERVER_CONFIG.publicUrl`) so a trailing slash can no longer produce a doubled-slash in any advertised URL — a correct, well-scoped fix that recovers the two commits dropped when #69 merged two commits behind its branch.

## Requirements context

No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, linked #69, and the surrounding code (`src/http/config.ts`, `src/http/auth-challenge.ts`, `src/http/server.ts`).

## Findings

### Critical
(none)

### Suggestion

- **Software engineering — two sources of truth for the same normalisation** (`src/http/auth-challenge.ts:10-12` and `src/http/config.ts:37-39`). `origin()` and `normaliseBaseUrl()` are now the identical `url.replace(/\/+$/, '')`. With `publicUrl` normalised at the source, `origin()` is redundant on the production path and only still does work because the new tests pass a slashed base *explicitly*. Consider having `auth-challenge.ts` import `normaliseBaseUrl` (keeping the local `origin` name as a thin alias if you like) so the rule lives in exactly one place — the same "safe by construction" argument the PR makes for the config-level fix applies to the helper too. Non-blocking; the current duplication is behaviourally correct.

### Information

- **Functionality / honesty of coverage** — The PR body's own note is accurate and appreciated: the three advertised-URL assertions (`auth-challenge.test.ts:147-180`) pass `SLASHED` explicitly, so they exercise `origin()` in `auth-challenge.ts`, not the config normalisation. The config fix's real coverage is the `normaliseBaseUrl` unit test (`:135-150`) and the bypass guard (`:184-212`). That's a reasonable coverage story for a one-line source change; flagging only so the distinction stays visible.
- **Software engineering — filesystem-walk guard test** (`auth-challenge.test.ts:191-211`). The "only config.ts reads `MCP_SERVER_URL`" guard walks `src` via a cwd-relative path and excludes `http/config.ts` with `endsWith('http/config.ts')` (a POSIX-separator match). This is fine for the Linux CI runner and consistent with how the rest of the suite runs, but it is implicitly coupled to the test being invoked from the repo root; worth a mental note if the test runner's cwd ever changes.
- **Software engineering — regex robustness** — `normaliseBaseUrl` collapses any number of trailing slashes (verified by the `///` case at `:143`) and leaves path-bearing bases intact (`:145`). Good edge-case coverage. A scheme-only input like `https://` would degrade to `https:`, but `MCP_SERVER_URL` is never set that way in practice, so this is not worth guarding.

## Verdict

**approved** (informational — the GitHub green-check is a separate human action). No Critical findings: the fix is correct, minimal, in-scope, introduces no user input into SQL/shell/HTTP, adds no dependencies, and has no performance implications. The Suggestion (DRY the normalisation helper) and Information notes are all optional.

tonychang04:
## QA — verified against today's master

Re-ran rather than trusting my earlier pass on the orphaned branch: #79 rewrote `/oauth/register` and `/oauth/authorize` since then, so the old result no longer covered this code.

With `MCP_SERVER_URL="http://127.0.0.1:3160/"` (trailing slash), master + this PR:

```
AS metadata       issuer + registration_endpoint clean
PRM + challenge   clean
POST /oauth/register    201
GET  /oauth/authorize   302, redirect_uri=…%2Foauth%2Fcallback   (callback not doubled)
doubled slashes across every advertised URL: 0
74 tests, lint 0 errors
```

**Control on master without this PR, identical config — the bug is live:**

```
advertised registration_endpoint:  http://127.0.0.1:3161//oauth/register
POST that advertised URL        ->  404
GET  //oauth/authorize          ->  404
doubled slashes advertised: 4
```

The reason this hides: posting to `/oauth/register` directly still works. It only breaks when a client does the correct thing and follows the URL we advertise. Health, discovery documents and manual probes all look fine; only real clients fail.

Worth landing before the Manufact deploy — `MCP_SERVER_URL` there is hand-entered, and a pasted trailing slash is a silent sign-in outage.

*Not tested:* a real browser sign-in, and prod.

— Quinn

jwfing (commented):
## Summary

A minimal, correct config-level fix: `normaliseBaseUrl` strips trailing slashes from `MCP_SERVER_URL` at the single point where `publicUrl` is defined, so every advertised URL is safe by construction. Well-tested, honestly scoped, and in-scope.

## Requirements context

No `docs/superpowers/` or `docs/specs/` directory exists in this repo (confirmed by listing) — **no matching spec/plan found; assessed against the PR description and the referenced #69 gap alone.** The PR body's claim is verifiable: on `master`, `publicUrl: process.env.MCP_SERVER_URL || 'http://localhost:3000'` was unguarded, and the two commits that add `normaliseBaseUrl` never landed because #69 merged two commits behind its branch.

## Findings

### Critical
(none) — the fix is correct and complete; nothing blocks merge.

### Suggestion
(none)

### Information

- **[functionality/maintainability] `src/http/auth-challenge.ts:10-12`** — `origin()` still strips trailing slashes itself, which is now redundant with the config-level normalisation on the real runtime path (`publicUrl` is already normalised before any consumer sees it). This is deliberate and acknowledged in the PR body's "honest coverage note" — the three advertised-URL tests pass a slashed base explicitly, so they exercise `origin()` rather than the config fix. Fine as defense-in-depth; noting only that the trailing-slash logic now lives in two places (`config.ts:37-39` and `auth-challenge.ts:10-12`).
- **[software-engineering] `src/http/auth-challenge.test.ts:218-236`** — the bypass guard matches the literal string `MCP_SERVER_URL` anywhere in a source file, including comments and unrelated strings, not only actual env reads. A future doc-comment naming the variable in another file would fail the test. This looks intentional ("nothing outside config.ts has any business naming the variable"), and the trade-off is sound given the alternative (regex on read-syntax) had the demonstrated blind spots the test itself documents — just flagging the strictness.

## Dimension coverage

- **Software engineering** — Excellent. Unit test for `normaliseBaseUrl` is mutation-verified; advertised-URL assertions check protected-resource docs, the derived metadata URL, and the `WWW-Authenticate` challenge, with the doubled-slash check correctly anchored after the scheme so `https://` isn't a false hit. The bypass guard is backed by two meta-tests (one asserting the walk actually reaches real files including `config.ts`/`server.ts`, one pinning the match predicate against all three naming forms). Conventions followed: `.js` import extensions, exported helper, dynamic `import()` in tests. The PR is unusually honest about what each test does and does not cover.
- **Functionality** — Solves the described bug. Every advertised URL flows through `SERVER_CONFIG.publicUrl` (verified: `server.ts` redirects/error bodies/banner, `OAUTH_CONFIG.callbackUrl` at `config.ts:92`, and all `auth-challenge.ts` helpers). Grep confirms no other file in the repo reads `MCP_SERVER_URL`, so the "safe by construction" claim holds. `SERVER_CONFIG` is defined before `OAUTH_CONFIG`, so the module-load reference is well-ordered.
- **Security** — No new attack surface. `MCP_SERVER_URL` is operator-configured, not user input; nothing reaches SQL/shell; no secrets logged or newly returned; no auth/authorization checks touched. The change is mildly security-positive (`publicUrl` is the canonical base used to avoid host-header spoofing per `server.ts:174`).
- **Performance** — No runtime impact (`String.replace` on one config value at startup). The guard test walks `src/` synchronously reading each file, but that cost is test-time only and negligible.

## Verdict

**approved** (informational — human still approves via the GitHub approve flow). Zero Critical findings; the two Information notes are optional and both reflect trade-offs the author already reasoned through.

jwfing (approved):
LGTM - approved.

PR #82, opened 2026-07-31
Wire signed client ids into /oauth/register and /oauth/authorize

tonychang04:
Wires the module from #79. This is the commit that actually removes the Redis
client-registration row — #79 built the mechanism, this makes it the one in use.

## What changes

| | before | after |
|---|---|---|
| `/oauth/register` | `SETEX mcp:oauth:client:<uuid>` | mints a signed id, stores nothing |
| `/oauth/authorize` | `GET` the row, then `EXPIRE` to refresh a 30-day idle TTL | verifies the signature, reads the registration out of the id |
| registration lifetime | 30 days idle | none — nothing is stored, so nothing expires |

The signing key is **derived, not configured**: `HMAC(INSFORGE_CLIENT_SECRET,
"mcp-client-id-v1")`. No fourth environment variable. The platform secret is
never itself the signing key — a client id is public and travels in every
authorize URL, so its signatures are public samples of whatever key produced
them; deriving means those samples are of a key that can do nothing else.

## The four review notes from #79, landed here

Each was non-blocking on #79 *because nothing called the module*. Wiring is what
makes `req.body` reach `mintClientId`, so they stop being hypothetical:

1. **`mintClientId` validated nothing, `readClientId` enforced a shape** — mint
   could issue an id that could never be read back. Now rejected at registration,
   where the client is listening, instead of at authorize, in a browser.
2. **No `redirect_uri` scheme or absoluteness check.** `/oauth/authorize`
   redirects to whatever the registration names, so the mint boundary is the
   choke point. Rejects `javascript:` `data:` `vbscript:` `file:` `blob:`,
   requires absolute, and allows plaintext `http` only for loopback (RFC 8252
   §8.3) — an authorization code over plaintext http to a remote host is the
   code read off the wire.
3. **Unbounded registration.** The payload *is* the client id and the id is in
   every authorize URL, so `redirect_uris` (≤10) and `client_name` (≤256) are
   capped, and each URI is ≤2048.
4. **Fields listed rather than spread**, so extra keys in a request body cannot
   ride along into the id.

### One judgment call worth disagreeing with

The scheme rule is a **denylist, not an allowlist**, deliberately. RFC 8252 §7.1
says a native client should use a reversed-domain private-use scheme
(`com.example.app:`); real MCP clients send `cursor://` and `vscode://`. An
allowlist of https + loopback http + dotted schemes is the more principled rule
**and it would reject clients that work today** — which is exactly the silent
breakage this module exists to stop. Reasoning is in the code next to the
constant, so it can be revisited rather than rediscovered.

## Existing clients

A client registered before this deploys takes the `invalid_client` path — its id
names a row nothing reads any more. In a browser that is the re-add page from
#74, which is the correct answer for them: re-register and it works. Called out
in the code so it reads as intended rather than as an oversight.

## Verification

Unit tests cover the validation (136 in the suite). More importantly, **the
success path was executed against a running server**, because that is the part
that has bitten me before — failure paths are easy to exercise and prove nothing
about the path being claimed.

```
1 POST /oauth/register                                       201 + signed id
2 GET  /oauth/authorize  minted id + registered redirect_uri 302 -> platform
3 GET  /oauth/authorize  redirect_uri NOT registered         400 invalid_request
4 GET  /oauth/authorize  forged signature                    400 invalid_client
5 GET  /oauth/authorize  old Redis-era mcp_<uuid> id         400 invalid_client
5b    ... same with Accept: text/html                        400 + re-add page
6 POST /oauth/register   javascript:alert(1)                 400 invalid_client_metadata
7 POST /oauth/register   http://attacker.test/cb             400 invalid_client_metadata
8 POST /oauth/register   no redirect_uris                    400 invalid_client_metadata
9 POST /oauth/register   cursor://anysphere.../callback      201 (accepted)
```

And the claim that matters, measured rather than asserted — the complete Redis
command log for a register *and* an authorize:

```
CLIENT SETINFO / CLIENT SETINFO / INFO      connection setup
PING                                        startup check
SETEX mcp:auth:state:vjAs191c00Xh...        authorize's auth state
```

**No `mcp:oauth:client:` key appears at all** — not a write at register, not a
read or an EXPIRE at authorize. The auth state is the next slice.

Run against a stand-in RESP server rather than real Redis, since this box has
none; that is enough for "which keys were touched", which is the claim.

## Not in this PR

The startup `redis.ping()` still blocks boot without Redis, and the auth state,
codes, bindings and sessions still use it. This is 1 of the remaining 5.

---
## Summary by cubic
Make OAuth client registration stateless with signed client IDs, removing Redis and the 30‑day idle TTL. Tightens redirect URI rules and scopes scheme normalization to loopback only to match RFC 8252 and the platform.

- **New Features**
  - `/oauth/register` mints a signed `client_id`; `/oauth/authorize` verifies and reads the registration from it — nothing stored.
  - Signing key is derived: HMAC(`INSFORGE_CLIENT_SECRET`, `mcp-client-id-v1`) — no new env var.
  - Strict redirect URI validation: absolute, denylist `javascript:`/`data:`/`vbscript:`/`file:`/`blob:`/`about:`, and `http` only for loopback (`127.0.0.1`/`::1`/`[::1]`).
  - Matching: exact byte match, with loopback IP literals ignoring the port (not `localhost`); scheme case-insensitivity applies on loopback only.
  - Bounds: `redirect_uris` ≤10, each URI ≤2048, `client_name` ≤256, and the minted `client_id` ≤4096 (checked at mint only). Only intended fields are signed.
  - Logging and errors: escape `client_name` in logs; invalid registration → 400 `invalid_client_metadata`; unverifiable id → 400 `invalid_client` (HTML shows re‑add page); missing signing key → 500 `server_error`.

- **Migration**
  - Clients registered before this will hit `invalid_client` and see the re‑add page; re‑register to continue.

<sup>Written for commit 38602d2ae1f06fac5e904f36b60975408d9f8691. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/82?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - OAuth client registrations now use signed client IDs, improving portability.
  - Added validation for redirect URIs, client names, registration size, and supported fields.
  - Authorization requires an exact match with a registered redirect URI.
  - Supports secure HTTPS, loopback HTTP, and approved private-use redirect schemes.

- **Bug Fixes**
  - Rejects invalid, unsafe, malformed, oversized, or unsupported redirect URIs.
  - Invalid client IDs and signing configuration errors now produce clearer failures.

### DIFF
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -3,7 +3,9 @@ import {
   mintClientId,
   readClientId,
   isRegisteredRedirectUri,
+  isAcceptableRedirectUri,
   InvalidClientIdError,
+  InvalidRegistrationError,
 } from './client-id.js';
 
 const SECRET = 'test-signing-secret';
@@ -122,5 +124,116 @@ describe('a missing secret is never silently substituted', () => {
   it('exports no secret generator', async () => {
     const mod = await import('./client-id.js');
     expect(Object.keys(mod)).not.toContain('generateSigningSecret');
+  });
+});
+
+describe('minting refuses what reading would reject', () => {
+  // john-bot's review of the unwired module: mintClientId validated nothing
+  // while readClientId enforced a shape, so mint could issue an id that could
+  // never be read back. Harmless while nothing called mint; the moment
+  // /oauth/register passes req.body straight in, it stops being harmless.
+  it('rejects an empty or absent redirect_uris', () => {
+    for (const bad of [[], undefined, null, 'https://app.example/cb', {}]) {
+      expect(() =>
+        mintClientId({ redirect_uris: bad as unknown as string[] }, SECRET)
+      ).toThrow(InvalidRegistrationError);
+    }
+  });
+
+  it('caps how many redirect_uris a registration may carry', () => {
+    // The payload is the client id and the client id is in every authorize
+    // URL, so an unbounded registration is an unbounded URL — and the failure
+    // would land on the redirect, not here where it can be explained.
+    const many = Array.from({ length: 11 }, (_, i) => `https://app.example/cb${i}`);
+    expect(() => mintClientId({ redirect_uris: many }, SECRET)).toThrow(InvalidRegistrationError);
+    expect(() => mintClientId({ redirect_uris: many.slice(0, 10) }, SECRET)).not.toThrow();
+  });
+
+  it('caps client_name', () => {
+    expect(() =>
+      mintClientId({ redirect_uris: [CALLBACK], client_name: 'x'.repeat(257) }, SECRET)
+    ).toThrow(InvalidRegistrationError);
+  });
+
+  it('carries only the fields we mean to sign', () => {
+    // req.body reaches this function. A spread would put every extra key the
+    // caller sent into the id.
+    const id = mintClientId(
+      {
+        redirect_uris: [CALLBACK],
+        client_name: 'Claude Code',
+        junk: 'x'.repeat(4096),
+      } as unknown as { redirect_uris: string[]; client_name: string },
+      SECRET
+    );
+    expect(readClientId(id, SECRET)).toEqual({
+      redirect_uris: [CALLBACK],
+      client_name: 'Claude Code',
+      iat: expect.any(Number),
+    });
+    expect(id).not.toContain('junk');
+  });
+
+  it('round-trips a registration with no client_name', () => {
+    const id = mintClientId({ redirect_uris: [CALLBACK] }, SECRET);
+    const reg = readClientId(id, SECRET);
+    expect(reg.client_name).toBeUndefined();
+    expect(reg.redirect_uris).toEqual([CALLBACK]);
+  });
+});
+
+describe('isAcceptableRedirectUri', () => {
+  // /oauth/authorize redirects to whatever the registration names, so the mint
+  // boundary is the place to keep dangerous values out entirely.
+  it('accepts https anywhere', () => {
+    expect(isAcceptableRedirectUri('https://app.example/cb')).toBe(true);
+  });
+
+  it('accepts plaintext http only on loopback', () => {
+    expect(isAcceptableRedirectUri('http://127.0.0.1:8765/callback')).toBe(true);
+    expect(isAcceptableRedirectUri('http://localhost:8765/callback')).toBe(true);
+    // An authorization code over plaintext http to a remote host is the code
+    // read off the wire.
+    expect(isAcceptableRedirectUri('http://app.example/cb')).toBe(false);
+    expect(isAcceptableRedirectUri('http://127.0.0.1.attacker.test/cb')).toBe(false);
+  });
+
+  it('accepts the private-use schemes real clients actually send', () => {
+    // RFC 8252 §7.1 asks for a reversed-domain scheme; shipped MCP clients do
+    // not comply, and an allowlist would reject them. Rejecting a working
+    // client to enforce a SHOULD is the wrong trade here.
+    for (const uri of [
+      'cursor://anysphere.cursor-retrieval/oauth/callback',
+      'vscode://insforge.mcp/callback',
+      'com.example.app:/oauth2redirect',
+    ]) {
+      expect(isAcceptableRedirectUri(uri)).toBe(true);
+    }
+  });
+
+  it('rejects script-capable and opaque schemes', () => {
+    for (const uri of [
+      'javascript:alert(1)',
+      'JavaScript:alert(1)',
+      'data:text/html,<script>alert(1)</script>',
+      'vbscript:msgbox(1)',
+      'file:///etc/passwd',
+      'blob:https://app.example/uuid',
+    ]) {
+      expect(isAcceptableRedirectUri(uri)).toBe(false);
+    }
+  });
+
+  it('rejects relative and unparseable values', () => {
+    // A relative URI would resolve against our own origin at redirect time.
+    for (const uri of ['/callback', 'callback', '', '://nope', 'https://' + 'a'.repeat(3000)]) {
+      expect(isAcceptableRedirectUri(uri)).toBe(false);
+    }
+  });
+
+  it('rejects non-strings', () => {
+    for (const bad of [undefined, null, 42, ['https://app.example/cb'], {}]) {
+      expect(isAcceptableRedirectUri(bad)).toBe(false);
+    }
   });
 });
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -5,6 +5,7 @@ import {
   readClientId,
   isRegisteredRedirectUri,
   isAcceptableRedirectUri,
+  FORBIDDEN_SCHEMES,
   InvalidClientIdError,
   InvalidRegistrationError,
 } from './client-id.js';
@@ -220,17 +221,55 @@ describe('isAcceptableRedirectUri', () => {
     }
   });
 
-  it('rejects script-capable and opaque schemes', () => {
+  it('rejects every scheme on the denylist, derived from the denylist', () => {
+    // This test used to be named for a class and assert five enumerated
+    // values, so adding a sixth scheme to the set left it uncovered while the
+    // name still promised the class. Derived from FORBIDDEN_SCHEMES now, so a
+    // scheme cannot be added to the set without being tested.
+    expect(FORBIDDEN_SCHEMES.size).toBeGreaterThan(0);
+    for (const scheme of FORBIDDEN_SCHEMES) {
+      expect(isAcceptableRedirectUri(`${scheme}whatever`), scheme).toBe(false);
+    }
+  });
+
+  it('rejects them however the scheme is cased', () => {
+    // RFC 3986 §3.1 — and URL normalisation is what makes the set lookup work.
+    for (const uri of ['JavaScript:alert(1)', 'DATA:text/html,x', 'File:///etc/passwd']) {
+      expect(isAcceptableRedirectUri(uri)).toBe(false);
+    }
+  });
+
+  it('rejects them in the shapes an attacker would actually write', () => {
+    // The derived test above proves the set is enforced; these pin the real
+    // payloads, which is a different claim.
     for (const uri of [
       'javascript:alert(1)',
-      'JavaScript:alert(1)',
       'data:text/html,<script>alert(1)</script>',
       'vbscript:msgbox(1)',
       'file:///etc/passwd',
       'blob:https://app.example/uuid',
+      'about:blank',
     ]) {
       expect(isAcceptableRedirectUri(uri)).toBe(false);
     }
+  });
+
+  it('matches the reference platform on the wrapper schemes, and diverges on one', () => {
+    // A reviewer proposed a structural rule for these six. Measured against the
+    // platform whose filter we match 12 for 12: it accepts five of them and
+    // rejects about:. Adopting the structural rule would have made us stricter
+    // than the thing we chose to copy — and this is the exact filter being
+    // ported to the platform, so a narrowing here propagates there.
+    for (const uri of [
+      'view-source:https://attacker.test/',
+      'jar:https://attacker.test/!/a',
+      'filesystem:https://attacker.test/temporary/a',
+      'intent://attacker.test/#Intent;scheme=https;end',
+      'chrome://settings',
+    ]) {
+      expect(isAcceptableRedirectUri(uri), uri).toBe(true);
+    }
+    expect(isAcceptableRedirectUri('about:blank')).toBe(false);
   });
 
   it('rejects relative and unparseable values', () => {
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -96,10 +96,18 @@ describe('redirect_uri matching', () => {
       'http://127.0.0.1:8765',
       'https://app.example/cb?next=https://attacker.example',
       'https://app.example.attacker.test/cb',
-      'HTTP://127.0.0.1:8765/callback',
     ]) {
       expect(isRegisteredRedirectUri(reg, near)).toBe(false);
     }
+  });
+
+  it('treats the scheme as case-insensitive, because it is', () => {
+    // This used to be asserted as a near-miss and rejected, which was an
+    // artifact of comparing strings rather than URLs. RFC 3986 §3.1 makes the
+    // scheme case-insensitive, the destination is byte-identical, and the
+    // platform's matchRedirectUri normalises the same way. Changed deliberately
+    // when loopback matching started parsing instead of comparing text.
+    expect(isRegisteredRedirectUri(reg, 'HTTP://127.0.0.1:8765/callback')).toBe(true);
   });
 
   it('rejects non-strings', () => {
@@ -298,5 +306,79 @@ describe('the id itself is bounded, not just its parts', () => {
     const longId = `mcp_${payload}.${sig}`;
     expect(longId.length).toBeGreaterThan(4096);
     expect(readClientId(longId, SECRET).redirect_uris).toHaveLength(1);
+  });
+});
+
+describe('loopback redirect_uris ignore the port (RFC 8252 §7.3)', () => {
+  // Max found the same bug in the platform's matchRedirectUri, where it had
+  // been flattened to includes() and would have broken `insforge login` — the
+  // CLI binds server.listen(0, '127.0.0.1'), so its port differs every run.
+  // The MCP had it too: a native MCP client registers on an ephemeral port,
+  // and the SDK re-registers only when it holds no client information at all,
+  // never in response to an authorize-time rejection. So an exact match hands
+  // the user the re-add page on every restart with nothing they can do.
+  const reg = (uri: string) => readClientId(mintClientId({ redirect_uris: [uri] }, SECRET), SECRET);
+
+  it('accepts a different port on the same loopback callback', () => {
+    const r = reg('http://127.0.0.1:54321/callback');
+    expect(isRegisteredRedirectUri(r, 'http://127.0.0.1:54321/callback')).toBe(true);
+    expect(isRegisteredRedirectUri(r, 'http://127.0.0.1:61999/callback')).toBe(true);
+    expect(isRegisteredRedirectUri(r, 'http://127.0.0.1/callback')).toBe(true);
+  });
+
+  it('still requires everything except the port to match', () => {
+    const r = reg('http://127.0.0.1:54321/callback');
+    for (const near of [
+      'http://127.0.0.1:61999/evil',            // different path
+      'http://127.0.0.1:61999/callback/../x',   // path traversal
+      'http://127.0.0.1:61999/callback?next=1', // query appears
+      'https://127.0.0.1:61999/callback',       // different scheme
+      'http://[::1]:61999/callback',            // different loopback family
+    ]) {
+      expect(isRegisteredRedirectUri(r, near)).toBe(false);
+    }
+  });
+
+  it('does not extend the relaxation to localhost', () => {
+    // RFC 8252 §8.3: localhost resolves through DNS and the hosts file, so it
+    // can be pointed off the loopback interface. The platform draws the line
+    // in the same place.
+    const named = reg('http://localhost:54321/callback');
+    expect(isRegisteredRedirectUri(named, 'http://localhost:54321/callback')).toBe(true);
+    expect(isRegisteredRedirectUri(named, 'http://localhost:61999/callback')).toBe(false);
+
+    const literal = reg('http://127.0.0.1:54321/callback');
+    expect(isRegisteredRedirectUri(literal, 'http://localhost:54321/callback')).toBe(false);
+  });
+
+  it('does not extend the relaxation to anything routable', () => {
+    const r = reg('https://app.example:8443/cb');
+    expect(isRegisteredRedirectUri(r, 'https://app.example:8443/cb')).toBe(true);
+    // A port-flexible match on a public host would let anyone with a service
+    // on another port of that host collect the code.
+    expect(isRegisteredRedirectUri(r, 'https://app.example:9999/cb')).toBe(false);
+    expect(isRegisteredRedirectUri(r, 'https://app.example/cb')).toBe(false);
+  });
+
+  it('preserves exact match for the custom schemes clients send', () => {
+    const r = reg('cursor://anysphere.cursor-retrieval/oauth/callback');
+    expect(isRegisteredRedirectUri(r, 'cursor://anysphere.cursor-retrieval/oauth/callback')).toBe(true);
+    expect(isRegisteredRedirectUri(r, 'cursor://anysphere.cursor-retrieval/oauth/evil')).toBe(false);
+  });
+
+  it('matches the platform on the same inputs', () => {
+    // Two servers, one dialect. Divergence here is a client that works against
+    // one and not the other, which is the hardest kind of bug to see.
+    const r = reg('http://127.0.0.1:8765/callback');
+    const platformSays: Array<[string, boolean]> = [
+      ['http://127.0.0.1:8765/callback', true],
+      ['http://127.0.0.1:1/callback', true],
+      ['http://127.0.0.1:8765/other', false],
+      ['http://localhost:8765/callback', false],
+      ['https://127.0.0.1:8765/callback', false],
+    ];
+    for (const [uri, expected] of platformSays) {
+      expect(isRegisteredRedirectUri(r, uri), uri).toBe(expected);
+    }
   });
 });
--- a/src/http/client-id.test.ts
+++ b/src/http/client-id.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect } from 'vitest';
+import { createHmac } from 'crypto';
 import {
   mintClientId,
   readClientId,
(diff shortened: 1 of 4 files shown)


tonychang04:
## QA — verified against real Redis, and the denylist argued with

**The central claim holds, measured with `redis-cli` against a live instance rather than a stand-in:**

```
POST /oauth/register              -> 201, signed id
keys matching mcp:oauth:client:*  -> 0
total keys in db after register   -> 0
after 4 authorize attempts        -> only mcp:auth:state:* exists
```

A Redis key is genuinely gone, not merely unused. The authorize matrix reproduces on real Redis too: registered uri `302`, unregistered `400`, forged signature `400`, old `mcp_<uuid>` `400`. 136 tests, lint 0 errors.

### The scheme denylist — you asked to be argued with

It leaks. Six schemes accepted in about a minute:

```
intent://x#Intent;scheme=javascript;end   201   <- documented Android bypass
filesystem:http://attacker/x              201
jar:http://attacker/x!/y                  201
view-source:http://attacker/x             201
chrome://settings                         201
about:blank                               201
```

It correctly rejects `javascript:` in every casing tried, plus `data:`, `vbscript:`, `blob:`, `file:` and plain `http://`. The list is well-built — the *approach* is what can't win. A denylist must enumerate every scheme a browser will ever interpret, and `intent://` is precisely the class it can't anticipate, since its purpose is to carry another scheme inside itself.

Your objection to an allowlist is also correct: `cursor://anysphere.cursor-retrieval/oauth/callback` and `vscode://ms-vscode.mcp/callback` both register fine today and are real clients.

**A structural rule satisfies both:**

```
allow  https:
allow  http:  only when the host is loopback (127.0.0.1 / ::1 / localhost)
allow  any private-use scheme that parses with a non-empty authority (scheme://host…)
reject everything opaque (scheme:payload with no //), and anything nesting a URL
```

That admits both real clients and rejects all six above — `intent:`/`jar:`/`filesystem:`/`view-source:` nest a URL, `about:`/`javascript:`/`data:` are opaque.

Whatever rule lands, it should be tested against the same sixteen inputs; I'll re-run them.

*Not tested:* a real browser following any of these, and prod.

— Quinn

jwfing (commented):
## Review — Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** A clean, well-reasoned change that makes the signed-client-id module from #79 the live mechanism, removes the `mcp:oauth:client:<uuid>` Redis row from the register/authorize path, and lands the four review notes from #79 as enforced validation — no Critical issues.

### Requirements context
This repo has no `/docs/superpowers/` (or `docs/specs/`) directory — **no matching spec/plan found, so this was assessed against the PR description, the linked predecessor #79, and the surrounding code.** The PR body is unusually explicit about intent (before/after table, the four #79 notes, the denylist-vs-allowlist judgment call, and the deliberate breaking change for pre-existing clients), which made intent easy to verify against the diff.

Cross-checks performed in the workspace:
- Confirmed no remaining reader of `mcp:oauth:client:` anywhere in `src/` (grep clean) — the `/oauth/token` handler (`src/http/server.ts:632`) and `/oauth/callback` (`src/http/server.ts:422`) both rely on the auth-state record, not the removed client row. The removal is complete, not partial.
- Confirmed the HMAC verify path (`readClientId`, `src/http/client-id.ts:179-223`) length-checks before `timingSafeEqual`, splits on `lastIndexOf('.')` (safe: base64url payload/signature never contain `.`), and re-validates payload shape after signature — so a legacy `mcp_<32hex>` id has no `.` → `InvalidClientIdError` → correct `invalid_client` path.

### Findings

#### Critical
(none)

#### Suggestion

**1. [software engineering] The wiring itself — the actual subject of this PR — has no automated test.** `src/http/client-id.ts` is thoroughly unit-tested (`client-id.test.ts`, +113 lines), but the route glue in `src/http/server.ts:247-295` (register) and `src/http/server.ts:301-416` (authorize) is only exercised manually (the RESP-log verification in the PR body). There is no `server.test.ts` in the repo, so this is *consistent with existing convention* rather than a regression — but given the repo's TDD posture, the branches most likely to rot are the ones added here and not covered: the `clientName` default (`server.ts:257`), the `InvalidRegistrationError → 400` vs `Error → 500` split (`server.ts:270-282`), and the `InvalidClientIdError → re-add HTML page` branch (`server.ts:353-364`). A lightweight supertest-style test over these two routes would guard the part that actually changed.

**2. [functionality] `/oauth/register` does not gate on `isOAuthConfigured()` the way `/oauth/authorize` does.** `authorize` returns a specific 500 up front (`server.ts:304-309`); `register` instead reaches `clientIdSigningKey()` (`server.ts:267`), which throws on a missing `INSFORGE_CLIENT_SECRET` and is caught into a generic `server_error` 500 (`server.ts:276-281`). The behavior is safe and handled — just inconsistent in message/shape between the two endpoints. Consider an early `isOAuthConfigured()` check in `register` for symmetry.

#### Information

**3. [security] Denylist-not-allowlist for redirect_uri schemes (`src/http/client-id.ts:73`, `95-97`) is a sound call.** The PR explicitly invited disagreement here. I agree with the author: an allowlist of https + loopback-http + dotted schemes would reject `cursor://` / `vscode://` clients that work today, which is exactly the silent breakage this module exists to prevent. The forbidden set (`javascript: data: vbscript: file: blob:`) covers the script-capable/opaque schemes that matter for the eventual `res.redirect()` to the client URI at callback time (`server.ts:437`), and browsers refuse those in a `Location` header regardless — so this is defensible defense-in-depth. Worth a future glance if non-browser redirect consumers ever appear, but not a change to request now.

**4. [functionality] Loopback + IPv6 handling is correct.** `url.hostname === '[::1]'` matches the WHATWG-normalized bracketed form, and `http://127.0.0.1.attacker.test/` is correctly rejected (test at `client-id.test.ts:150`). No bypass found in the http-loopback exception.

**5. [software engineering] `client_name` is always non-empty at the route** (`server.ts:257` defaults to `'MCP Client'`), so the "no `client_name`" round-trip (`client-id.test.ts:177-182`) is only reachable via the unit test, never via the live route. Harmless — noting only that the signed payload will always carry a `client_name` in practice.

### Verdict
**approved** (informational — zero Critical findings; human approval remains a separate action). The security core is solid: forgery-resistant HMAC ids, exact-match `redirect_uri` per RFC 6749 §3.1.2.3 that *cannot* be skipped because the registration is the id, derived (not shared) signing key, no secrets logged, and a net performance win from dropping two Redis round-trips per authorize. The pre-existing-client breakage is deliberate, documented, and routed to the re-add page. The two Suggestions are non-blocking.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review — Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** Wiring #79's stateless signed-client-id module into the register/authorize endpoints (removing the last Redis client-registration row) with the four #79 review notes landed as real validation — a clean, well-reasoned, well-tested change with no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessed against the PR description, the linked #79/#74 intent, and the module's own doc comments. Verified locally on head `b24d6c9`: `npx tsc --noEmit` clean, `eslint` clean on all three changed source files, and `vitest run src/http/client-id.test.ts` → 24 passed.

### Findings

#### Critical
(none)

#### Suggestion
- **[software-engineering]** `src/http/server.ts:247-295, 301-380` — the endpoint *wiring* itself has no automated test: `client_name` defaulting to `'MCP Client'` on a non-string/empty value, the `InvalidRegistrationError → 400 invalid_client_metadata` mapping, and the `server_error → 500` fallback on a missing signing key are only exercised by the manual run in the PR body. The underlying `client-id.ts` module is thoroughly covered (24 tests, including the regression cases for the #79 notes), and the repo has no HTTP-endpoint test harness today, so this is non-blocking — but a small supertest-style test over `/oauth/register` and `/oauth/authorize` would lock in the branch that "bit before" (the success path) rather than relying on a one-time manual check.
- **[security]** `src/http/client-id.ts:73` — the scheme rule is a denylist by deliberate design, and the reasoning (real MCP clients send `cursor://`/`vscode://`, an allowlist would silently break them) is sound and documented next to the constant. Flagging only so it's an explicit reviewer decision rather than an omission: the true guard against code theft is the signature + exact `redirect_uri` match, and the denylist is defence-in-depth for the browser `Location` case — so I agree with the call. No change requested.

#### Information
- **[functionality]** `src/http/server.ts:344-368` — clients registered before this deploys take the `invalid_client` / re-add-page path. Intended and clearly documented in code; correct behavior for them.
- **[functionality]** Scope is correctly limited to register + authorize; auth-state/codes/sessions and the startup `redis.ping()` remain Redis-backed (1 of the remaining 5, per the PR). No dangling reads of the old `mcp:oauth:client:` key remain anywhere in `src/` — confirmed by grep.
- **[security]** Good hygiene throughout: signing key is HMAC-derived from `INSFORGE_CLIENT_SECRET` (never the secret itself), verification is timing-safe with a length pre-check to avoid the `timingSafeEqual` length-throw oracle (`client-id.ts:197-201`), `redirect_uri` is an exact match per RFC 6749 §3.1.2.3, and no secret/PII is logged (the client id is public by construction). Loopback-only plaintext-http logic verified against Node's `URL` parsing (`[::1]`, `127.1` normalization, and `127.0.0.1.attacker.test` all behave as the code assumes).
- **[performance]** Net improvement — removes the Redis `GET` + `EXPIRE` at authorize and the `SETEX` at register; the per-request HMAC is negligible. The registration caps (`redirect_uris` ≤10, URI ≤2048, `client_name` ≤256) mean a maximal registration yields a large client id in the authorize query string, but that's an unrealistic input and the tradeoff is documented at `client-id.ts:45-56`.

### Verdict
`approved` — zero Critical findings. Posted as a COMMENT (the bot does not click GitHub approve; a human does that via the approve flow). The two Suggestions are optional; the change is safe to merge as-is.

jwfing (commented):
## Review — Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** A clean, well-tested wiring of the #79 signed-client-id module that removes the Redis client-registration row from the OAuth hot path; I found no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, so there is no matching spec/plan — assessing against the PR description, the four review notes it carries forward from #79, and the linked issues (#74 re-add flow, #79 mechanism). Intent is unusually well-documented in the PR body and in code comments.

Verified locally at head `8bf6c4e`: `tsc --noEmit` clean, `eslint` clean on the three changed source files, **full `vitest` suite 140/140 pass** (28 of them in `client-id.test.ts`).

### Findings

#### Critical
(none)

#### Suggestion
- **[software-engineering] The server-side error mapping in `src/http/server.ts:263-282` and `:331-369` has no automated test.** The new branching — `InvalidRegistrationError` → 400 `invalid_client_metadata`, any other mint error → 500 `server_error`, `InvalidClientIdError` → HTML re-add page vs. JSON `invalid_client`, everything else → 500 — is verified only by the manual server run described in the PR body. The `client-id.ts` unit coverage is excellent, but the wiring that translates those errors into HTTP responses is exactly the layer a future refactor could silently break (e.g. reordering the `instanceof` check so a config error is reported to a client as `invalid_client`). This matches the repo's existing pattern (no route has a supertest-level test), so it's non-blocking — but given the security relevance of these exact status/error-code choices, a route-level test asserting each branch would be worth adding.

#### Information
- **[functionality] Intended breaking change for pre-deploy clients — well handled.** Any client registered before this deploys will fail signature verification and take the `invalid_client` path (`server.ts:352-368`); in a browser that renders the #74 re-add page. This is called out in both the PR body and the code comment and is the correct resolution, but flagging it here since it is a real, user-visible behavior change on rollout. Verified it does **not** break token exchange: `/oauth/token` (`server.ts:632-657`) never looks up `client_id` against a store (it's used only for analytics and the code/`redirect_uri`/PKCE are validated against the stored auth state in `oauth-manager.ts:248-252`), so the longer self-describing id flows through the callback→token legs unchanged.
- **[security] `readClientId` computes the HMAC over the full caller-supplied payload before the signature check (`client-id.ts:232-240`).** In principle an unverified caller can make the server hash arbitrary bytes, but the `client_id` arrives in the authorize query string, whose length is already bounded well below any concern by Node's `max-http-header-size` (~16KB, request line included) and typical nginx `large_client_header_buffers` (8KB). No meaningful DoS surface; noting only for completeness.
- **[security] Scheme validation is a denylist, not an allowlist (`client-id.ts:97,119-127`).** Called out as a deliberate judgment call in the PR and the code. I concur it's the right trade here: redirect_uris are exact-matched per-registration (`isRegisteredRedirectUri`, RFC 6749 §3.1.2.3) and scoped to the registering client, so a permissive-but-non-script scheme only ever affects that client's own callback. The dangerous vectors (script/opaque schemes, plaintext http to non-loopback) are closed.

### Notes on what was checked and is solid (no action needed)
- **Security (positive):** timing-safe signature compare with a length pre-check to avoid the `timingSafeEqual` length-mismatch throw becoming an oracle (`client-id.ts:236-238`); signing key **derived** via `HMAC(INSFORGE_CLIENT_SECRET, "mcp-client-id-v1")` rather than being the platform secret itself (`config.ts:116-122`); loopback-only plaintext `http` per RFC 8252 §8.3; fields listed rather than spread so extra request-body keys can't ride into the signed payload (`client-id.ts:187-191`); id-length bound enforced at mint only so tightening it later can't invalidate issued ids. No secrets are logged (the register log dropped the id and logs only the client name; error logs pass the `Error`, not the key).
- **Performance (positive):** net removal of Redis round-trips on the OAuth path — a `SETEX` at register and a `GET`+`EXPIRE` at authorize are gone; HMAC is cheap and synchronous-but-negligible. No new N+1, unbounded loop, or blocking I/O.

### Verdict
**approved** — zero Critical findings. The Suggestion (route-level test for the error mapping) and the Information notes are non-blocking; author may address or defer. (Posted as a comment; explicit GitHub approval remains a separate human action.)

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** A well-reasoned, thoroughly-tested security hardening that makes OAuth client registration stateless (removing the Redis `mcp:oauth:client:` row) by carrying the registration inside an HMAC-signed client id, and wires in the four validation gaps that were non-blocking on #79 only because nothing called the module — no Critical issues found.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the four review notes it lands from #79, the `#74` re-add-page migration path, and the surrounding code (`src/http/client-id.ts`, `config.ts`, `server.ts`). Verified locally on the head commit: `vitest` on `client-id.test.ts` → **35/35 pass**, `tsc --noEmit` clean, `eslint` clean, `npm ci` clean.

---

### Critical
(none)

### Suggestion

- **Software engineering / test coverage — the route wiring itself is only manually verified.** `src/http/client-id.ts` is excellently unit-tested (35 cases: bounds, schemes, loopback port relaxation, cross-server parity), but the new branches *in the routes* — `InvalidRegistrationError → 400 invalid_client_metadata` vs. plain `Error → 500 server_error` (`src/http/server.ts:269-282`), the `InvalidClientIdError → invalid_client` vs. `500` split at authorize (`src/http/server.ts:334-369`), and the `clientName` normalization (`src/http/server.ts:257`) — are exercised only by the manual run described in the PR body. The repo has no route-level (supertest) harness, so this is consistent with convention and low blast radius; a small supertest over `/oauth/register` and `/oauth/authorize` would lock that status/error-code mapping against future regressions. Non-blocking.

### Information

- **Functionality — echoed-but-unused registration fields.** The `/oauth/register` response still echoes `grant_types` / `response_types` / `token_endpoint_auth_method` / `scope` (`src/http/server.ts:290-293`), but only `redirect_uris` + `client_name` are carried in the signed id, and authorize enforces only `redirect_uri`. These fields are therefore cosmetic and unenforced — which is identical to the pre-PR behavior (the old Redis row echoed them and authorize never checked them either), so this is **not a regression**. Noting it so a future reader doesn't assume `scope`/`grant_types` are honored.

- **Performance — signing key derived per request.** `clientIdSigningKey()` (`src/http/config.ts:130-131`) recomputes `HMAC(INSFORGE_CLIENT_SECRET, "mcp-client-id-v1")` on every register/authorize call. Cost is negligible (one SHA-256 of a short label) and lazy derivation is the right call for the stdio/test import paths, so this is purely informational — a module-level memo would save an inconsequential hash.

- **Security — no relevant weaknesses; posture is a net improvement.** Walked the full chain end-to-end:
  - The core property holds: authorize recovers the registration from the signed id (`server.ts:333`) and only proceeds to a `redirect_uri` that exact-matches the signed set (`server.ts:375`), and the browser is ultimately redirected at `/oauth/callback` to `authState.redirectUri` (`server.ts:436-445`) — i.e. the *validated, registered* URI, not a request-supplied one. This closes the mcp-use `oauthProxy` open-redirect / code-theft gap the module targets.
  - The RFC 8252 §7.3 loopback port relaxation (`client-id.ts:302-332`) is correctly bounded: scheme/host/path/query must still match, it's restricted to IP literals (`localhost` excluded per §8.3), and the port-flexible path only widens targets on the victim's own loopback interface — the accepted §7.3 threat model. Path traversal (`/callback/../x`) is neutralized by `URL` normalization; array-valued query params fall through the `typeof` guards to a clean `invalid_client` rather than crashing.
  - Derived signing key (not the platform secret itself), `timingSafeEqual` with a length pre-check, `throw`-not-null on unreadable ids, and a documented denylist all check out. The denylist-vs-allowlist scheme choice is a defensible judgment call (documented at `client-id.ts:82-97`) — `res.redirect` Location headers make `javascript:`/`data:` inert in browsers anyway, so it is defense-in-depth.
  - No secrets/PII newly logged; the noisy `client_id` was dropped from the authorize log line (`server.ts:352`). No new dependencies.
  - The pre-existing-clients migration (`invalid_client` → re-add page) is intentional and clearly documented.

---

### Verdict: **approved** (informational — the human still approves via the approve flow)

Zero Critical findings. This is a clean, unusually well-documented change with strong test coverage on the extracted logic; the Suggestion and Information items are all non-blocking. Nice work carrying the `#79` review notes into the wiring commit where they actually bite.

jwfing (commented):
## Review — Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** Makes OAuth client registration stateless by carrying the registration inside an HMAC-signed `client_id`, removing the Redis client-registration row and the 30-day idle-TTL bug along with it; the implementation is careful, well-documented, and thoroughly tested, with no blocking issues.

### Requirements context
No spec/plan directory in this repo (`docs/superpowers/`, `docs/specs/`, and `docs/` are all absent), so this was assessed against the PR description, the linked predecessor #79, and the surrounding code. The PR is scoped exactly to "land the module from #79 as the code path in use" plus the four review notes from #79 — the diff matches that claim (4 files, no drive-by changes).

Verified locally against head `65f87b6`: `npx vitest run src/http/client-id.test.ts` → **38 passing**; `tsc --noEmit` clean; `eslint` on the three changed source files clean. `randomUUID`/`getRedisClient` remain used elsewhere in `server.ts`, so no dangling imports from the Redis removal.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality / consistency — case-insensitive scheme match is loopback-only.** `isRegisteredRedirectUri` (`src/http/client-id.ts:325-355`) does a byte-exact `includes()` first and only falls through to URL-parsing (which normalizes scheme case) when the *requested* host is a loopback IP literal. So a client that registers `https://app.example/cb` and later sends `HTTPS://app.example/cb` at authorize is rejected, while the same casing on `http://127.0.0.1/cb` is accepted. The test at `src/http/client-id.test.ts:104-107` and its comment frame scheme case-insensitivity as a general property ("because it is", RFC 3986 §3.1) and claim the platform's `matchRedirectUri` "normalises the same way" — but here that normalization only reaches the loopback branch. Not a regression (the old code was also byte-exact) and low blast radius in practice (clients echo the registered URI verbatim), but if the goal is "two servers, one dialect" this is a small divergence worth either closing or narrowing the comment/test claim to loopback.

#### Information

- **Security — the scheme rule is a denylist, and it accepts exotic schemes deliberately.** `FORBIDDEN_SCHEMES` (`src/http/client-id.ts:113-120`) blocks the script-capable/opaque set and lets everything else through, so `view-source:`, `intent://`, `chrome://`, `jar:`, `filesystem:` all mint. This is documented at length and matched "12 for 12" against the platform filter being ported, and it does not create a *new* code-theft vector: registration is open, so an attacker can already register `https://attacker.test/cb` directly — the exotic schemes add nothing an attacker couldn't do with plain https, and the inherent DCR-phishing risk (victim consenting to an attacker-controlled `client_id`) is unchanged by scheme choice. Flagging only so the accepted set is a conscious, revisitable line (the reasoning lives next to the constant, which is the right call). No action needed.

- **Functionality — registration echoes fields it does not sign.** `/oauth/register` (`src/http/server.ts:~254-296`) returns `grant_types`, `response_types`, `token_endpoint_auth_method`, and `scope` from the request/defaults, but only `redirect_uris` + `client_name` are signed into the id and recovered at authorize. This matches prior behavior (the old Redis row stored them but authorize never read them) and is spec-compatible, so it's fine — noting it so the "the registration IS the id" mental model isn't taken to mean *every* returned field round-trips.

- **Performance — net improvement.** The change removes a `SETEX` at register and a `GET` + `EXPIRE` at authorize from the OAuth hot path, replacing them with an in-process HMAC — no new N+1s, loops, or blocking I/O. `MAX_CLIENT_ID_LENGTH` (4096) is derived from the 8192 request-line limit rather than guessed, and `readClientId` deliberately does not re-enforce it so tightening the bound can't invalidate issued ids. Nicely reasoned.

- **Software engineering — strong.** Clear TDD signal: regression tests for all four #79 review notes, edge cases (non-strings, path traversal, loopback family mismatch, oversized product-of-bounds, hand-forged over-limit id proving read doesn't re-check), `FORBIDDEN_SCHEMES` exported so the denylist test derives from the set rather than a hand-copied list, and signature verification uses `timingSafeEqual` with a length pre-check to avoid the throw-as-oracle. Errors are correctly split (`InvalidRegistrationError` → 400 `invalid_client_metadata`; `InvalidClientIdError` → 400 `invalid_client`; missing signing key → 500 `server_error`, never mislabeled as an unknown client). Log-injection guard on `client_name` via `JSON.stringify` at `src/http/server.ts:~287` is a nice touch.

---

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings. The one Suggestion and the Information notes are non-blocking; the case-insensitive-scheme consistency item is the only one worth a quick look before merge, and even that is safe to defer.

jwfing (commented):
## Review — Wire signed client ids into `/oauth/register` and `/oauth/authorize`

**Summary:** Wiring PR that makes the signed-client-id module from #79 the live implementation — `/oauth/register` mints a signed, self-describing id and stores nothing, `/oauth/authorize` verifies the signature and reads the registration back out — and it lands the four review notes from #79 (mint-side validation, redirect_uri scheme/absoluteness checks, registration bounds, field-listing instead of spread). The implementation is careful, correct, and the reasoning is documented next to the code.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessed against the PR description, the linked #79 review notes, and the surrounding code. Verified the PR's central claim by grepping the tree: nothing outside these routes reads the removed `mcp:oauth:client:` key, `CLIENT_REGISTRATION_TTL` has no remaining references, and the only surviving `randomUUID` (`server.ts:886`) is for session ids — so removing the Redis registration write breaks nothing downstream. Downstream, the validated `redirect_uri` flows into `createAuthorizationState` and is the redirect choke point.

---

### Critical
(none)

The security-sensitive paths hold up under scrutiny:
- **Signature verification** (`client-id.ts:255-263`) length-checks before `timingSafeEqual` (avoiding the length-mismatch throw as an oracle) and runs before `JSON.parse`, so a forged id can't reach the parser.
- **`isRegisteredRedirectUri`** (`client-id.ts:325-355`) keeps exact match and relaxes *only* the port and *only* for loopback IP literals, with scheme/host/path/query still exact — the WHATWG `URL` parse normalizes path traversal (`/callback/../x` → `/x`) so the traversal test genuinely can't slip through, and a loopback-only relaxation can't be leveraged cross-host. Correctly excludes `localhost` (DNS-resolvable) per RFC 8252 §8.3.
- **Log forging** is closed by `JSON.stringify(clientName)` at `server.ts:288`.
- **Error→status mapping** in both routes correctly separates client fault (400) from server misconfiguration (500 on a missing/unusable signing key), and never reports a key problem as an unknown client.

### Suggestion

**Software engineering — no automated coverage of the wiring itself.** The module (`client-id.ts`) is exhaustively unit-tested (310 added lines), but the two wired routes in `server.ts:247-299` / `305-420` — specifically the `InvalidRegistrationError`→400 vs other→500 branch (`server.ts:269-282`), the `InvalidClientIdError`→400 vs other→500 branch (`server.ts:338-347`), and the HTML re-add-page path (`server.ts:357-368`) — are verified only manually per the PR body. That branching is exactly the kind of glue a future refactor can silently flip (e.g. a 500 turning into a 400 that tells a real client to re-register). This is a **Suggestion, not a blocker**, because the repo has no route-level test harness today (all suites are module-level), so it's consistent with existing convention; a lightweight `supertest` around the register/authorize error mapping would be a worthwhile follow-up.

### Information

- **Functionality — echoed registration fields.** The register response (`server.ts:290-298`) reflects `grant_types` / `response_types` / `token_endpoint_auth_method` / `scope` straight from `req.body` (or defaults). Only `redirect_uris` + `client_name` are actually validated and carried in the signed id, so those extra fields are unvalidated reflections back to the same caller — no blast radius (not stored, not used downstream), and largely pre-existing behavior. Noting for awareness only.
- **Functionality — `localhost` registrations still break on restart.** Because the port relaxation is deliberately loopback-IP-literal-only, a native client that registers `http://localhost:PORT/...` (rather than `127.0.0.1`) still gets the re-add page after a port change. This is intentional (mirrors the platform's `matchRedirectUri` and is safer), and the MCP/CLI clients bind `127.0.0.1`, so it's unlikely to bite — worth keeping in mind if a client is later found to register `localhost`.
- **Performance.** Net positive: this removes a Redis round-trip from both register (`SETEX`) and authorize (`GET` + `EXPIRE`). `clientIdSigningKey()` re-derives the HMAC per request (`config.ts:125-131`); the cost is negligible (HMAC of a 16-char label) but it could be memoized if ever hot. No N+1, no unbounded loops (`redirect_uris` capped at 10), no blocking I/O added.
- **Security — bounds.** `MAX_CLIENT_ID_LENGTH` enforced at mint only and deliberately *not* re-checked in `readClientId`, so tightening it later can't invalidate already-issued ids — a nice property, and the test at `client-id.test.ts` hand-builds an over-bound id to pin it.

---

No Critical findings. Strong, well-reasoned security work with thorough unit coverage; the Suggestion (route-level wiring tests) and Information notes are non-blocking. Posted as a comment — the explicit GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #83, opened 2026-07-31
Boot without Redis, and say so honestly

tonychang04:
Max's suggestion in-thread, and it is a strictly smaller change than the probe
branch it replaces: make the boot ping conditional so the real server boots on
Manufact today, and use it — rather than a throwaway probe — to answer whether
the edge routes `/oauth/*` and `/.well-known/*` or only `/mcp`.

## Why it never booted

`startServer()` pings Redis **before** `app.listen`. With no `REDIS_*` set the
promise rejects, the process exits 1, and nothing is reachable — not `/health`,
not discovery, not even a log line naming the cause once the container has
looped a few times. That is what every Manufact deploy since the start-script
fix has died on.

Redis is on its way out of this server anyway. Until the last use is gone, a
deploy without it should still come up: **a server that answers and refuses to
sign anyone in is far more diagnosable than one that is not there.**

## The predicate

```js
isRedisConfigured() === !!(REDIS_URL || REDIS_HOST)
```

Deliberately **not** `REDIS_PORT`, `REDIS_PASSWORD` or `REDIS_TLS`.
`getRedisConfig()` defaults the host to `localhost`, so counting any of those
would mean "configured" and send us straight back to a connection that can never
succeed. Tested explicitly.

## Three things that fell out, each found by running it

1. **`/health` needed Redis.** It calls `getStats()`, which `SCAN`s. So the one
   endpoint a deploy is graded on failed on a server that was up and serving
   every discovery document. `activeSessions` is now `null` when there is
   nothing to count.
2. **The banner printed `Redis: localhost:6379` whether or not anything was
   there** — the same shape of lie as the `REDIS_PASSWORD` that looked
   configured and never reached ioredis, which this repo already has a comment
   about.
3. **My first warning message claimed client registration works.** It does not
   on this branch — `/oauth/*` still 500s, because registration and auth state
   are the next slices. Corrected to say what is actually served. Reading the
   diff would not have caught that; running it did.

## verify-deploy gets the check that keeps this honest

Its bare `/oauth/authorize` probe stops at the missing-parameter check, which is
**before** the client lookup — so it scored **26/26 against a server that could
not log anyone in.** That is a false green on the exact gate we are about to
rely on.

A probe carrying parameters reaches the lookup. An unknown client there is a
`400 invalid_client`; a `500` means the lookup itself failed.

```
no Redis    FAIL  the client lookup on /oauth/authorize works (a login could start)
                  — status 500, the remaining Redis uses are not gone yet
            26/27, exit 1

with Redis  PASS  the client lookup on /oauth/authorize works
            28/28, exit 0
```

It discriminates, it fails today, it names why, and it turns green on its own
once the lookup no longer needs Redis. The session-ratio check is skipped rather
than defaulted to `0`, which would have turned "unknown" into "zero active" and
graded a real leak against it.

## Measured, with no Redis process anywhere

```
/health                                    200   sessions.activeSessions = null
/.well-known/oauth-authorization-server    200
/.well-known/oauth-protected-resource      200
/.well-known/oauth-protected-resource/mcp  200
/.well-known/oauth-protected-resource/sse  200
POST /mcp                                  401 + WWW-Authenticate resource_metadata
```

With Redis present nothing changes: ping runs, the sweep starts, the ratio is
checked, 28/28.

## What this does not do

Sign-in still will not complete — `/oauth/*` and every session path need the
remaining Redis uses removed (#82 is registration; auth state is next). This is
the change that makes Manufact **reachable and observable**, which is the step
that has been blocking everything else.

---
## Summary by cubic
Allow the server to boot without Redis so deployments are reachable and diagnosable. Health, discovery, and `POST /mcp` work; `/oauth/*` and session routes remain unavailable when Redis isn’t configured.

- **Bug Fixes**
  - Make Redis ping and session sweep conditional via `isRedisConfigured()` (`REDIS_URL` or `REDIS_HOST` only).
  - `/health` no longer depends on Redis; returns `activeSessions: null` when absent.
  - Startup banner now shows Redis as “not configured — `/oauth/*` and sessions unavailable” instead of a misleading `localhost:6379`.
  - `scripts/verify-deploy.mjs`: add a parameterized `/oauth/authorize` probe to verify client lookup; skip the session ratio check when no Redis.

<sup>Written for commit 5f9d4e1497beb8cdf5f21e0066f90713ea98b3b8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/83?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/verify-deploy.mjs
+++ b/scripts/verify-deploy.mjs
@@ -125,14 +125,20 @@ check(
 
 // Runtime sessions should not run away from the ones Redis still holds. A
 // small excess is normal in-flight; an order of magnitude is the leak.
+// `activeSessions` is null when the server is running without Redis, which is
+// not a failure — there is simply no second number to compare against, and a
+// default of 0 would silently turn "unknown" into "zero active" and grade a
+// real leak against it.
 const sessions = health.body?.sessions;
-if (sessions) {
-  const { activeSessions: active = 0, memorySessionCount: resident = 0 } = sessions;
+if (sessions && typeof sessions.activeSessions === 'number') {
+  const { activeSessions: active, memorySessionCount: resident = 0 } = sessions;
   check(
     'resident sessions are not far above active ones',
     resident <= Math.max(active * 2, active + 50),
     `active=${active} resident=${resident}`
   );
+} else if (sessions) {
+  console.log('  ~ session ratio not checked: no Redis, so there is no active count');
 }
 
 // --- the 401 has to tell a client where to look ---------------------------
@@ -220,6 +226,24 @@ check(
   'the authorize endpoint is reachable and validating',
   authorize.status === 400,
   `status ${authorize.status}`
+);
+
+// The bare probe above stops at the missing-parameter check, which is BEFORE
+// the client lookup — so it passes on a server that cannot log anyone in. A
+// probe carrying parameters reaches the lookup, and an unknown client there is
+// a 400 `invalid_client`. A 500 means the lookup itself failed: no Redis, or
+// no signing key. This is the difference between "the routes exist" and "a
+// login could start", and everything else in this file only proves the former.
+const authorizeWithParams = await getJson(
+  '/oauth/authorize?client_id=verify-deploy-probe' +
+    '&redirect_uri=http%3A%2F%2F127.0.0.1%3A1%2Fcb&response_type=code'
+);
+check(
+  'the client lookup on /oauth/authorize works (a login could start)',
+  authorizeWithParams.status === 400,
+  authorizeWithParams.status === 500
+    ? 'status 500 — the client lookup failed; the remaining Redis uses are not gone yet'
+    : `status ${authorizeWithParams.status}`
 );
 
 // --- follow the chain to whichever AS the resources name -------------------
--- a/src/http/redis.test.ts
+++ b/src/http/redis.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { resolveRedisClientSpec, type RedisConfig } from './redis.js';
+import { resolveRedisClientSpec, isRedisConfigured, type RedisConfig } from './redis.js';
 
 /**
  * These exist because of a bug that was invisible: REDIS_PASSWORD was read into
@@ -104,5 +104,30 @@ describe('resolveRedisClientSpec — cluster', () => {
     expect(spec).toMatchObject({ nodes: [{ host: 'redis.internal', port: 6379 }] });
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     expect((spec.options as any).redisOptions.tls).toEqual({});
+  });
+});
+
+describe('isRedisConfigured', () => {
+  // The boot ping used to be unconditional, so a container with no Redis
+  // exited 1 before app.listen and nothing was reachable — not /health, not
+  // discovery, not a log line naming the cause. This predicate is what lets
+  // the server come up and say so instead.
+  it('is true when a URL or a host is given', () => {
+    expect(isRedisConfigured({ REDIS_URL: 'redis://cache:6379' })).toBe(true);
+    expect(isRedisConfigured({ REDIS_HOST: 'cache' })).toBe(true);
+  });
+
+  it('is false when nothing points anywhere', () => {
+    expect(isRedisConfigured({})).toBe(false);
+    expect(isRedisConfigured({ REDIS_URL: '', REDIS_HOST: '' })).toBe(false);
+  });
+
+  it('does not count a port, a password or TLS as a destination', () => {
+    // getRedisConfig defaults the host to localhost, so counting any of these
+    // would mean "configured" and send us straight back to a connection that
+    // can never succeed — the failure this exists to prevent.
+    expect(isRedisConfigured({ REDIS_PORT: '6379' })).toBe(false);
+    expect(isRedisConfigured({ REDIS_PASSWORD: 'hunter2' })).toBe(false);
+    expect(isRedisConfigured({ REDIS_TLS: 'true', REDIS_CLUSTER: 'true' })).toBe(false);
   });
 });
--- a/src/http/redis.ts
+++ b/src/http/redis.ts
@@ -49,6 +49,22 @@ export function getRedisConfig(): RedisConfig {
 }
 
 /**
+ * Has anyone actually pointed us at a Redis?
+ *
+ * `getRedisConfig` defaults the host to localhost, which is a sensible default
+ * for a developer and a lie in a container — nothing is listening there, so the
+ * server retried a connection that could never succeed and exited 1 before
+ * `app.listen`. That is what 19 Manufact deploys failed on after the missing
+ * start script was fixed.
+ *
+ * REDIS_PORT is deliberately not counted: a port with no host is not a
+ * destination, and treating it as one would put us straight back to localhost.
+ */
+export function isRedisConfigured(env: NodeJS.ProcessEnv = process.env): boolean {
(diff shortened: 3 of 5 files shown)


jwfing (commented):
## Review — Boot without Redis, and say so honestly

**Summary:** Makes the boot Redis ping conditional on `isRedisConfigured()` so a deploy without `REDIS_URL`/`REDIS_HOST` comes up serving `/health`, discovery, and the 401 challenge instead of exiting 1 before `app.listen`; the change is well-scoped, honest about what it does *not* fix, and comes with tests. No blocking issues found.

**Requirements context:** No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the linked follow-ups (#82 registration), and surrounding code.

---

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering / test coverage** — `src/http/session-manager.ts:595-597`: the new `{ activeSessions: null }` early-return is the exact behavior `/health` (and therefore the deploy grade) now depends on, yet only the `isRedisConfigured` predicate has a direct unit test (`redis.test.ts`). The null contract is exercised only indirectly via `verify-deploy.mjs` against a live server. A focused `getStats()` test asserting `activeSessions === null` when Redis is unconfigured would lock the contract against a future regression (e.g. someone "tidying" the return type back to `number`). Non-blocking — the added predicate tests are good and the diff is otherwise well-covered.

#### Information

- **Functionality** — `src/http/session-manager.ts:588-597` / `src/http/server.ts:178-193`: `/health` now emits `sessions.activeSessions: null`. `verify-deploy.mjs` is correctly updated to treat it as "unknown, not zero" (`scripts/verify-deploy.mjs:128-141`), but any out-of-repo consumer (uptime dashboard, alerting rule doing `active || 0`) would coerce `null` to a falsy/zero active count. Worth a heads-up to whoever owns the deploy dashboards; nothing to change in this repo.
- **Style** — `isRedisConfigured()` is evaluated independently at three sites (boot gate `server.ts:1211`, banner `server.ts:1218-1223`, `getStats` `session-manager.ts:595`). `process.env` is immutable at runtime so the three reads are always consistent — noted only for awareness, no change needed.

#### Verified, no findings

- **Functionality** — the Redis client is a lazy singleton (`redis.ts:126-148`), so with no config `getRedisClient()` is never called at boot and no socket is opened; the predicate (`!!(REDIS_URL || REDIS_HOST)`) correctly ignores `REDIS_PORT`/`REDIS_PASSWORD`/`REDIS_TLS`, matching `getRedisConfig`'s `localhost` default and its tests. Shutdown is safe unconfigured: `closeRedisClient()` guards on `if (redisClient)` and `closeAllSessions()` is wrapped in try/catch (`server.ts:1318-1331`). The new `verify-deploy` probe reaches the client lookup and its `check(..., status === 400, ...)` is robust because `getJson` always returns `.status` regardless of body parseability (`verify-deploy.mjs:81-96`).
- **Security** — no security-relevant changes. The banner now prints *less* (`not configured — …` instead of a misleading `localhost:6379`); no secrets/tokens are logged or newly returned; the verify probe uses a hardcoded `client_id`/loopback `redirect_uri`; no new user input reaches SQL/shell/HTTP. No new dependencies.
- **Performance** — negligible. `isRedisConfigured()` is a couple of env reads per call; the SCAN path is unchanged and now skipped entirely when there is no Redis to scan.

---

Zero Critical findings. Clean, honest, tightly-scoped change with matching tests and a discriminating deploy gate that fails today for the right reason and turns green on its own once the client lookup no longer needs Redis. The one Suggestion (direct `getStats` null test) is optional. Posted as a comment — the green-checkmark approval is a separate human action.

jwfing (commented):
## Summary

Makes the boot Redis ping conditional so the server comes up and serves `/health` + discovery when no Redis is configured, tells the truth in the banner and warning, stops `/health` from depending on Redis, and adds a verify-deploy probe that distinguishes "routes exist" from "a login could start" — a tight, honest, well-tested change that matches its description.

## Requirements context

No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or any `docs/`) directory. Assessed against the PR description, the linked follow-up work (#82 registration, auth-state next), and the surrounding code. Verified the runtime behavior claims against the actual handlers.

## Findings

### Critical
(none)

### Suggestion

**Software engineering — degraded-boot paths lack direct unit coverage** (`src/http/server.ts:1211-1229`, `src/http/session-manager.ts:595-597`). The new predicate `isRedisConfigured` is thoroughly unit-tested (`src/http/redis.test.ts:109-133`, good TDD), but the branches it drives — the no-Redis boot warning, the `redisLine` banner, and `getStats()` returning `activeSessions: null` — are only exercised by `verify:deploy` against a live server. `getStats` is genuinely awkward to unit test (needs a SessionManager), so this is low priority, but a small test asserting `getStats()` returns `{ activeSessions: null, memorySessionCount: 0 }` when `REDIS_HOST`/`REDIS_URL` are unset would lock in the exact `/health` contract that verify-deploy now depends on.

**Functionality — idle sweep silently off in degraded mode** (`src/http/server.ts:1217`). `startIdleSweep` runs only in the Redis branch. Correct today, because no session can be established without Redis so `runtimeSessions` stays empty — but the coupling is implicit. If a future slice lets runtime sessions exist before all Redis uses are gone, in-memory reaping would be silently disabled. A one-line comment noting "no sweep because no session can form without Redis yet" would prevent that from becoming a latent leak.

### Information

**Functionality — verify-deploy is now red-by-design on Manufact until the Redis uses are removed** (`scripts/verify-deploy.mjs:237-247`). Confirmed correct: `express@^5.1.0` auto-forwards a rejected async handler to the error middleware, and `/oauth/authorize` (`src/http/server.ts:306-333`) checks `isOAuthConfigured()` → required params → `response_type=code` *before* the `redis.get` client lookup, so the parameter-carrying probe reaches the lookup and yields `400 invalid_client` with Redis / `500` without it — exactly what the check keys on. Worth flagging for operators: this gate intentionally exits 1 (26/27) against the current deploy, so a red verify-deploy here is expected, not a regression, until #82+ land.

**Security — no regressions; fails closed.** No new user input reaches SQL/shell/HTTP, no secrets are logged (the warning/banner reveal only *presence* of config, never values), and degraded mode weakens no auth check — a login attempt without Redis returns `500`, i.e. it fails closed rather than open. The parameter values used by the verify-deploy probe are fixed and inert.

**Performance — neutral-to-positive.** `getStats()` now returns early without the `SCAN` loop when Redis is unconfigured; no new queries, loops, or blocking work in a hot path. Minor note: once any request lazily constructs the client in degraded mode (e.g. a hit to `/oauth/*`), that ioredis singleton will reconnect to `localhost:6379` indefinitely (`retryStrategy` caps the delay at 3s but not the attempt count), logging `[Redis] Connection error` on a loop. Not a crash — the `'error'` handler at `src/http/redis.ts:135-137` prevents that — just log noise while the remaining Redis uses are being removed.

**Style** (`src/http/server.ts:1231`). Banner title trimmed to `(OAuth)`; cosmetic and consistent with the change.

## Verdict

`approved` (informational — the human still approves via the GitHub approve flow). Zero Critical findings; the change is scoped, correct, and its central claims are verified against the code. The two Suggestions are non-blocking hardening.

jwfing (approved):
LGTM - approved.

PR #84, opened 2026-07-31
Every browser-visible OAuth error says what to do

tonychang04:
Max verified on `master` that #74's actionable page is wired to exactly **one**
branch — the unknown `client_id` at `server.ts:340`. The `redirect_uri` mismatch
at `:387` returns `invalid_request` and never reaches it.

So the branch a real user is most likely to hit shows this, in a browser tab:

```json
{"error":"invalid_request","error_description":"redirect_uri does not match any registered redirect URIs for this client."}
```

No mention of re-adding the server. No `npx @insforge/install`. And the person
looking at that tab is the only one who can act — the MCP client is waiting on a
loopback callback that will never arrive.

## Why this is a helper and not a patched line

Patching `:387` alone would repeat the trailing-slash mistake exactly: guard the
two consumers you happened to notice, leave the other eight. Every
browser-reachable error now goes through one function.

```
/oauth/authorize   missing params · bad response_type · unknown client
                   corrupted registration · redirect_uri mismatch
/oauth/callback    platform error with nowhere to bounce it · missing code/state
                   expired state · token exchange failed
```

**The callback is always a browser navigation** — it is the platform redirecting
the person here — so JSON was never the right answer on any branch of it.

The JSON body is unchanged for anything that asks for JSON. A conforming client
that does read it still gets exactly what the spec says it should; the HTML is
additive.

## Wording is per error code, on purpose

The remedies genuinely differ, and a single generic page would be wrong for most
of them:

| code | says | names the command |
|---|---|---|
| `invalid_client` | re-register | yes |
| `invalid_request` (authorize) | "restarted on a different port" → re-register | yes |
| `server_error` | ours to fix, not yours | **no** |
| no params at all | "this page is the middle of a sign-in" | **no** |
| expired state | "sign-ins expire after a few minutes" | **no** |

`server_error` deliberately does not suggest reinstalling — re-adding the server
does not fix our outage, and telling someone to do it sends them round a loop
that cannot terminate. Opening `/oauth/authorize` by hand gets its own wording
rather than being told to reinstall something.

## Tests

`humanFormOf` is pure and exported, so the sentences are testable without
standing up express. That is deliberate: **the wording is the whole point of the
change**, and it is what a route-level test would be least likely to assert.

The test asserts the property over the error codes rather than one example —
"the page is reachable from one branch" is precisely the bug being fixed, so a
test that checks one branch would reproduce it.

## Verified against a running server, both audiences, all seven branches

```
authorize: no params            browser 400 <!DOCTYPE html>   client 400 {"error":...
authorize: bad response_type    browser 400 <!DOCTYPE html>   client 400 {"error":...
authorize: unknown client       browser 400 <!DOCTYPE html>   client 400 {"error":...
authorize: redirect mismatch    browser 400 <!DOCTYPE html>   client 400 {"error":...
callback:  no code/state        browser 400 <!DOCTYPE html>   client 400 {"error":...
callback:  expired state        browser 400 <!DOCTYPE html>   client 400 {"error":...
callback:  platform error       browser 400 <!DOCTYPE html>   client 400 {"error":...
```

What a person now reads on the live failure:

> **This connection needs to be set up again**
> The address your app asked us to send you back to is not the one it
> registered. That usually means the app was reinstalled or restarted on a
> different port. Remove the InsForge MCP server from your client and add it
> back, and this will complete normally.
> `npx @insforge/install`

Success path still 302.

## Relationship to the other PRs

#82 removes the *cause* for the loopback case (RFC 8252 §7.3 port-agnostic
matching). This fixes the *symptom* for every case, including the ones #82 does
not and should not make disappear — a genuinely wrong path is still an error,
it just needs to be a legible one. Branched off `master`, so it is independent
of #81/#82/#83 and can merge in any order.

---
## Summary by cubic
Show clear HTML guidance for all browser-visible OAuth errors while preserving JSON for API clients. Fix reconnect instructions to remove and add by URL so repairs work on any host.

- **Bug Fixes**
  - All errors from `/oauth/authorize`, `/oauth/callback`, and `/oauth/register` go through `sendOAuthError` (including catch blocks), returning HTML to browsers and JSON to clients; `/oauth/select-project` is unchanged.
  - `invalid_client` and `invalid_request` now instruct:
    - `npx add-mcp remove <publicUrl>/mcp -y`
    - `npx add-mcp <publicUrl>/mcp`
  - Removal uses the URL, not a hardcoded name; both commands are built from `SERVER_CONFIG.publicUrl` + `STREAMABLE_HTTP_ENDPOINTS.mcp` to avoid no-op cutovers and host header spoofing.
  - Callback-specific states (manual open, missing/expired params, token exchange failures) show clear messages; `server_error` shows no action.
  - `reconnectCommand` returns both commands; template `action` supports `string | string[]`.
  - Tests cover wording, content negotiation, host-exact commands, remove-before-add behavior, and removal-by-URL.

<sup>Written for commit dc648538a7abb6116eda32f8175cb4b789336332. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/84?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Enhanced OAuth error handling with clearer, human-readable guidance for browsers.
  * Browser responses now provide reconnect instructions, redirect guidance, and actionable error details.
  * Programmatic clients continue receiving unchanged JSON error responses.
  * Reconnect actions support separate remove and add commands for easier recovery.

* **Tests**
  * Added comprehensive coverage for browser and client responses, fallback errors, redirects, reconnect commands, and response ordering.

### DIFF
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -1,0 +1,125 @@
+import { describe, it, expect, vi } from 'vitest';
+import { humanFormOf, prefersHtml, reconnectCommand, sendOAuthError } from './oauth-error-response.js';
+
+const MCP_URL = 'https://mcp.insforge.dev/mcp';
+
+/** Just enough of express's Request for the Accept negotiation. */
+const asBrowser = { accepts: (types: string[]) => (types.includes('html') ? 'html' : false) };
+const asClient = { accepts: (types: string[]) => (types.includes('json') ? 'json' : false) };
+
+function mockResponse() {
+  const res = {
+    status: vi.fn(() => res),
+    type: vi.fn(() => res),
+    send: vi.fn(() => res),
+    json: vi.fn(() => res),
+  };
+  return res as unknown as Parameters<typeof sendOAuthError>[1] & typeof res;
+}
+
+describe('who is actually looking at this error', () => {
+  it('gives a browser HTML', () => {
+    const res = mockResponse();
+    sendOAuthError(asBrowser, res, 400, { error: 'invalid_request' });
+    expect(res.type).toHaveBeenCalledWith('html');
+    expect(res.json).not.toHaveBeenCalled();
+    expect(res.status).toHaveBeenCalledWith(400);
+  });
+
+  it('gives a program the unchanged OAuth JSON', () => {
+    // A conforming client that does read the body still gets exactly what the
+    // spec says it should — the HTML is additive, not a replacement.
+    const res = mockResponse();
+    const body = { error: 'invalid_request', error_description: 'nope' };
+    sendOAuthError(asClient, res, 400, body);
+    expect(res.json).toHaveBeenCalledWith(body);
+    expect(res.send).not.toHaveBeenCalled();
+  });
+});
+
+describe('every browser-visible failure says what to do', () => {
+  // #74 wired the page to exactly one branch, the unknown client_id. The one a
+  // user is most likely to hit — a redirect_uri that does not match — kept
+  // returning raw JSON. Patching that single branch would have repeated the
+  // trailing-slash mistake, so this asserts the property over the codes.
+  it.each([
+    'invalid_client',
+    'invalid_request',
+    'server_error',
+    'unsupported_response_type',
+    'token_exchange_failed',
+    'access_denied',
+  ])('%s produces a heading and a message a person can act on', (error) => {
+    const human = humanFormOf({ error }, MCP_URL);
+    expect(human.heading.length).toBeGreaterThan(10);
+    expect(human.message.length).toBeGreaterThan(40);
+    // No machine vocabulary leaking into the sentence a person reads.
+    expect(human.heading).not.toMatch(/_|redirect_uri|client_id/);
+  });
+
+  it('names the command that repairs a REMOTE client, not the local installer', () => {
+    // These users arrived over the remote URL — that is the only way to reach
+    // this page at all, since this server serves it. `npx @insforge/install`
+    // installs a local stdio server with an API key: a different product, not
+    // a repair. This page said that until it was traced.
+    for (const error of ['invalid_client', 'invalid_request']) {
+      const action = humanFormOf({ error }, MCP_URL).action as string[];
+      expect(action).toEqual([
+        'npx add-mcp remove insforge -y',
+        'npx add-mcp https://mcp.insforge.dev/mcp',
+      ]);
+      expect(action.join(' ')).not.toContain('@insforge/install');
+    }
+  });
+
+  it('removes before adding, so it is not a no-op on an unchanged URL', () => {
+    // add-mcp deep-merges under the server key and byte-preserves the rest, and
+    // the key is inferred from the hostname. So on a same-hostname cutover a
+    // bare `add-mcp <url>` rewrites an identical entry and changes nothing:
+    // the stuck user runs it, fails identically, and opens a ticket. Removing
+    // first is what makes the instruction do something — and it is correct on
+    // a new hostname too, where it clears the stale entry anyway.
+    const [first, second] = reconnectCommand(MCP_URL);
+    expect(first).toContain('remove');
+    expect(second).toContain(MCP_URL);
+  });
+
+  it('names the host the person is actually talking to', () => {
+    // On the Manufact slug it must say the slug, or someone follows the
+    // instruction and reconnects to the box we are migrating off.
+    expect(reconnectCommand('https://keen-pulse-fsjr9.run.mcp-use.com/mcp')).toEqual([
+      'npx add-mcp remove insforge -y',
+      'npx add-mcp https://keen-pulse-fsjr9.run.mcp-use.com/mcp',
+    ]);
+  });
+
+  it('does not tell someone to reinstall when that would not help', () => {
+    // A server-side failure is not fixed by re-adding the server, and saying so
+    // would send a person round a loop that cannot terminate.
+    expect(humanFormOf({ error: 'server_error' }, MCP_URL).action).toBeUndefined();
+  });
+
+  it('renders the redirect_uri case as the port-change it almost always is', () => {
+    const human = humanFormOf({ error: 'invalid_request' }, MCP_URL);
+    expect(human.message).toMatch(/different port/);
+  });
+
+  it('falls back to the description for a code it has never seen', () => {
+    expect(humanFormOf({ error: 'weird_new_code', error_description: 'the disk melted' }, MCP_URL).message)
+      .toBe('the disk melted');
+  });
+
+  it('still says something when there is no description either', () => {
+    expect(humanFormOf({ error: 'weird_new_code' }, MCP_URL).message.length).toBeGreaterThan(40);
+  });
+});
+
+describe('prefersHtml', () => {
+  it('is false for a client asking for json', () => {
+    expect(prefersHtml(asClient)).toBe(false);
+  });
+
+  it('is true for a browser', () => {
+    expect(prefersHtml(asBrowser)).toBe(true);
+  });
+});
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -65,7 +65,7 @@ describe('every browser-visible failure says what to do', () => {
     for (const error of ['invalid_client', 'invalid_request']) {
       const action = humanFormOf({ error }, MCP_URL).action as string[];
       expect(action).toEqual([
-        'npx add-mcp remove insforge -y',
+        `npx add-mcp remove ${MCP_URL} -y`,
         'npx add-mcp https://mcp.insforge.dev/mcp',
       ]);
       expect(action.join(' ')).not.toContain('@insforge/install');
@@ -84,13 +84,24 @@ describe('every browser-visible failure says what to do', () => {
     expect(second).toContain(MCP_URL);
   });
 
-  it('names the host the person is actually talking to', () => {
+  it('names the host the person is actually talking to, in BOTH commands', () => {
     // On the Manufact slug it must say the slug, or someone follows the
     // instruction and reconnects to the box we are migrating off.
-    expect(reconnectCommand('https://keen-pulse-fsjr9.run.mcp-use.com/mcp')).toEqual([
-      'npx add-mcp remove insforge -y',
-      'npx add-mcp https://keen-pulse-fsjr9.run.mcp-use.com/mcp',
+    //
+    // Both commands, because the removal used to hardcode "insforge" while the
+    // add step was host-derived: on the slug the removal then matched no
+    // server at all and the repair silently did nothing. add-mcp matches
+    // `server.identity === query` and identity is the URL for a remote entry,
+    // so passing the URL is exact on every host.
+    const slug = 'https://keen-pulse-fsjr9.run.mcp-use.com/mcp';
+    expect(reconnectCommand(slug)).toEqual([
+      `npx add-mcp remove ${slug} -y`,
+      `npx add-mcp ${slug}`,
     ]);
+    for (const command of reconnectCommand(slug)) {
+      expect(command).toContain('keen-pulse-fsjr9');
+      expect(command).not.toContain('insforge');
+    }
   });
 
   it('does not tell someone to reinstall when that would not help', () => {
--- a/src/http/oauth-error-response.ts
+++ b/src/http/oauth-error-response.ts
@@ -104,7 +104,23 @@ export interface HumanForm {
  * never reached. Measured, not inferred.
  */
 export function reconnectCommand(mcpUrl: string): string[] {
-  return ['npx add-mcp remove insforge -y', `npx add-mcp ${mcpUrl}`];
+  // Remove BY URL, not by name. The name is derived from the hostname, so a
+  // hardcoded "insforge" matches nothing on any host that infers a different
+  // key — including the Manufact slug we spent a day testing on:
+  //
+  //   mcp.insforge.dev                   -> key "insforge"
+  //   keen-pulse-fsjr9.run.mcp-use.com   -> key "keen-pulse-fsjr9"
+  //
+  // I made the ADD step host-derived and left the REMOVE step hardcoded, which
+  // is the same inconsistency in one function. Max caught it.
+  //
+  // The URL works because add-mcp's findMatchingServers matches
+  // `server.identity === query` as well as by name, and extractServerIdentity
+  // returns the server's url for a remote entry (add-mcp@2.0.0,
+  // chunk-2LJORNPV.js). So passing the URL is exact, host-derived by
+  // construction, and cannot drift the way a copy of their inference rule
+  // would.
+  return [`npx add-mcp remove ${mcpUrl} -y`, `npx add-mcp ${mcpUrl}`];
 }
 
 /**
--- a/src/http/oauth-error-response.ts
+++ b/src/http/oauth-error-response.ts
@@ -1,0 +1,196 @@
+import type { Response } from 'express';
+import { renderOAuthErrorPage } from './templates/oauth-error.js';
+import { SERVER_CONFIG, STREAMABLE_HTTP_ENDPOINTS } from './config.js';
+
+/**
+ * Sending an OAuth error to whoever is actually looking at it.
+ *
+ * `/oauth/authorize` and `/oauth/callback` are reached by a browser
+ * navigation. The MCP client never reads those responses — it is waiting on a
+ * loopback callback that will now never arrive — so the only party who sees an
+ * error body is the person staring at the tab.
+ *
+ * #74 added a page saying what to do, and wired it to exactly one branch: the
+ * unknown client_id. Every other way authorize can fail kept returning raw
+ * OAuth JSON to that same tab. The one a user is most likely to hit — a
+ * redirect_uri that does not match the registration — was among them, so the
+ * live failure mode on the deployed server shows `{"error":"invalid_request"}`
+ * and no instruction at all.
+ *
+ * That is the same shape of mistake as normalising a trailing slash at two of
+ * ten consumers. So this is a helper every browser-reachable error goes
+ * through, not a second patched branch.
+ */
+
+/**
+ * The one thing this needs from a request. Narrower than express's Request on
+ * purpose: the negotiation is the whole logic, and a two-line stand-in in a
+ * test is worth more than a full Request mock nobody reads.
+ */
+export interface AcceptNegotiable {
+  accepts(types: string[]): string | false;
+}
+
+export interface OAuthErrorBody {
+  error: string;
+  error_description?: string;
+}
+
+export interface HumanForm {
+  heading: string;
+  message: string;
+  /** Optional literal(s) the user should run, rendered as code. */
+  action?: string | string[];
+}
+
+/**
+ * The command that actually repairs one of these users.
+ *
+ * NOT `npx @insforge/install`, which is what this page said until Max traced
+ * where the connected clients came from. There are two install paths and they
+ * produce different products:
+ *
+ *   npx @insforge/install   installs @insforge/mcp as a LOCAL stdio server with
+ *                           an API key. Never contacts this server at all.
+ *   npx add-mcp <url>       adds the REMOTE server — server.json remotes[], and
+ *                           what the setup docs tell people to run.
+ *
+ * Everyone who can possibly SEE this page arrived over the remote URL, because
+ * this page is served by the remote server. So the old instruction was wrong
+ * one hundred percent of the time, not merely sometimes — and following it
+ * would have quietly moved someone onto a different product configuration
+ * rather than fixing the one they had.
+ *
+ * TWO commands, not one, and that is the part measured rather than assumed.
+ * Blair read add-mcp@2.0.0: writeJsonConfig deep-merges under the server key
+ * and byte-preserves everything else, and the key is inferred from the
+ * hostname. So on a SAME-hostname cutover `add-mcp <url>` writes an entry
+ * identical to the one already there and changes nothing — the stuck user runs
+ * it, fails identically, and opens a ticket. Removing first is what makes the
+ * instruction do something.
+ *
+ * Correct on both roads, which is why it does not wait on the hostname
+ * decision: on a new hostname the removal clears the stale entry you wanted
+ * gone anyway; on the same hostname it is the only thing that helps. I checked
+ * `remove <query>` exists with -y in that version rather than trusting the
+ * name, and that the package stores no tokens of its own (its single "oauth"
+ * reference is an oauthScopes CONFIG field).
+ *
+ * What this still cannot promise: whether removing the config entry also drops
+ * the client's saved OAuth registration is client-specific — add-mcp deletes
+ * its config key and nothing more. If some client keys its registration by
+ * server name rather than URL, even remove+add leaves it stuck. That is a real
+ * test on a real editor, not something to settle in a comment.
+ *
+ * Built from publicUrl, which is the CANONICAL host this deployment is
+ * configured to be — deliberately not the host the request arrived on. The Host
+ * header is attacker-controlled, and a page that echoed it would hand a
+ * stuck user a command pointing wherever the attacker liked. Same reason the
+ * discovery documents use publicUrl (server.ts, "to avoid host header
+ * spoofing").
+ *
+ * Do not "fix" this to use the requested host. Today one deployment serves one
+ * hostname so the two coincide and nothing distinguishes them; the moment a
+ * retired hostname is pointed at a live deployment they diverge, and the
+ * requested-host reading tells someone stuck on the dead host to re-add the
+ * dead host. That is a loop with no exit, aimed at exactly the users a cutover
+ * exists to rescue.
+ *
+ * Note for anyone planning that retired-hostname setup: it does not work by
+ * itself, and not because of this function. A client reaching an old host whose
+ * publicUrl is the NEW host gets a protected-resource document naming a
+ * different origin, and the SDK throws in selectResourceURL
+ * (client/auth.js:339) BEFORE it ever opens the browser — so this page is
+ * never reached. Measured, not inferred.
+ */
+export function reconnectCommand(mcpUrl: string): string[] {
+  return ['npx add-mcp remove insforge -y', `npx add-mcp ${mcpUrl}`];
+}
+
+/**
+ * What to tell a person, given the machine-readable error.
+ *
+ * Pure and exported so the wording is testable without standing up express —
+ * the part that matters here is the words, and they are the part a route test
+ * would be least likely to assert.
+ */
+export function humanFormOf(body: OAuthErrorBody, mcpUrl: string): HumanForm {
+  switch (body.error) {
+    case 'invalid_client':
+      return {
+        heading: 'This connection needs to be set up again',
+        message:
+          'The app you are connecting from is not registered with this server any more. ' +
+          'Nothing is wrong with your account and no data was lost — remove the InsForge MCP ' +
+          'server from your client and add it back with the command below, and this will ' +
+          'complete normally.',
+        action: reconnectCommand(mcpUrl),
+      };
+
+    case 'invalid_request':
+      // Overwhelmingly a redirect_uri that does not match the registration.
+      // The remedy is identical to invalid_client — re-register — and the
+      // cause is not something the person can inspect, so saying "invalid
+      // request" and stopping would be true and useless.
+      return {
+        heading: 'This connection needs to be set up again',
+        message:
+          'The address your app asked us to send you back to is not the one it registered. ' +
+          'That usually means the app was reinstalled or restarted on a different port. ' +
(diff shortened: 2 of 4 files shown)


jwfing (commented):
## Review: Every browser-visible OAuth error says what to do

**Summary:** A clean, well-motivated refactor that routes every browser-reachable OAuth error through one `sendOAuthError` helper (HTML for browsers, unchanged JSON for programs), fixing the #74 gap where only the `unknown client_id` branch got the actionable page.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`, or any `docs/`) directory exists in this repo — assessing against the PR description, the linked #74 behavior, and the surrounding code alone. No invented requirements.

**Verification performed (fresh workspace at head `520cadc`):**
- `vitest run src/http/oauth-error-response.test.ts` → **15/15 pass**
- `tsc --noEmit` → clean
- `eslint` on all three changed files → clean

---

### Findings

#### Critical
(none) — no correctness, security, data-loss, or behavior-breaking issues found.

#### Suggestion

**Functionality — callback platform-error branch reuses code-keyed wording out of context** (`src/http/server.ts:464-467`)
This is the one `sendOAuthError` call in the callback handler that passes **no** `human` override, so it falls back to `humanFormOf(body.error)`. Two consequences:
- For the common real case `error=access_denied` (the user clicked *Deny* on consent), `humanFormOf` hits the `default` branch → heading "Sign-in could not be completed", `action: 'npx @insforge/install'`, and the fallback message "removing the InsForge MCP server … is the usual remedy". Telling someone who *chose* to decline to reinstall the server is misleading.
- If the platform ever returns `error=invalid_request` here, the user sees the authorize-specific "the app was reinstalled or restarted on a **different port** / redirect_uri" wording (`oauth-error-response.ts:57-66`), which has nothing to do with a callback platform error.

`humanFormOf` maps purely on the OAuth *code*, but the `invalid_request` wording is authorize-redirect_uri-specific. Consider giving this branch its own `human` override (as the other three callback branches already do), or making the message code+context aware.

**Test coverage — the `action` intent for `default`-mapped codes is unasserted** (`src/http/oauth-error-response.test.ts:37-49`)
The PR's stated design ("`server_error` deliberately does not name the command") is well tested for `server_error`, but the codes that route through `default` (`unsupported_response_type`, `token_exchange_failed`, `access_denied`) all silently inherit `action: REINSTALL`, and the `it.each` only asserts heading/message length — not `action`. A test pinning which codes get the reinstall command would lock in the "don't send people round a loop" invariant that motivates the whole PR. Similarly, no test exercises `sendOAuthError`'s `human`-override merge — specifically that passing `{ action: undefined }` actually *suppresses* the default `action` via the `{ ...humanFormOf(body), ...human }` spread. That's a close-call behavior (relies on spread copying explicit-`undefined` own keys) and every server.ts call site depends on it; a one-line regression test would be cheap insurance.

#### Information

**Security (verified, positive):** attacker-controllable `error` / `error_description` from the `/oauth/callback` query string reach the HTML page via the `default` branch's `message: body.error_description`. This is safe — `renderOAuthErrorPage` escapes every interpolated value (`heading`, `message`, `action`) through `escapeHtml` (`src/http/templates/oauth-error.ts:23,92-94,100-107`), which covers `& < > " '`. No reflected XSS. Noting because it's the one place raw request input flows into rendered HTML. No secrets/tokens are newly logged or returned (only OAuth error codes/descriptions, already logged pre-PR).

**Performance:** No new queries, loops, allocations, or blocking I/O. Purely string-building on an already-terminal error path. No concerns.

**Software engineering (positive):** Extracting `prefersHtml`/`sendOAuthError`/`humanFormOf` into a pure, exported module and testing the wording directly is the right call — it makes the actual behavior of interest (the sentences a person reads) testable without express, and the property-over-codes test correctly targets the "reachable from one branch only" bug rather than one example. `prefersHtml` degrades safely: a client with no/`*/*` Accept header gets JSON (express returns the first listed type). Import style, error handling, and naming match the surrounding file.

---

Zero Critical findings; posted as a `COMMENT`. The two Suggestions (callback-branch wording context; asserting `action` per code + the override-merge behavior) are worth a look but non-blocking.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: Every browser-visible OAuth error says what to do

**Summary:** Routes browser-reachable `/oauth/authorize` and `/oauth/callback` errors through a single `sendOAuthError` helper that renders an actionable HTML page to browsers while leaving the OAuth JSON body unchanged for programmatic clients — a clean, well-motivated refactor that fixes the raw-JSON-in-a-browser-tab problem #74 only half-solved.

**Requirements context:** No `docs/superpowers/` (or any spec/plan directory) exists in this repo, so I assessed against the PR description alone. The description is unusually thorough — it states the intent ("every browser-reachable error now goes through one function"), the audience model, and a manual verification matrix across seven branches, which I used as the source of intent.

---

### Critical
(none)

### Suggestion

**Functionality — two browser-reachable `catch` blocks still emit raw JSON, contradicting the "every branch" thesis** — `src/http/server.ts:421-425` (authorize catch) and `src/http/server.ts:572-583` (callback catch). The PR's central argument is that patching individual branches "repeats the trailing-slash mistake — guard the two consumers you happened to notice, leave the other eight," and that *every* browser-reachable error now flows through `sendOAuthError`. But both handler-level `catch` blocks still do `res.status(...).json(...)`. These are genuinely browser-reachable: any exception thrown by `oauthManager.createAuthorizationState` / `getAuthorizationState`, a `fetch` network failure to the token endpoint, or a Redis `setex` failure lands here and shows raw JSON in the user's tab. Ironically, the callback catch is exactly the "platform outage during token exchange" path the `server_error` wording was written for, yet it never reaches that wording. Non-blocking because this is pre-existing behavior the PR strictly improves around rather than regresses — but routing these two through `sendOAuthError` (with the `server_error` human form) would fulfill the completeness claim the PR makes for itself.

**Functionality — callback platform-error falls through to the default reconnect wording** — `src/http/server.ts:464-467` passes no `human` override, so `humanFormOf` hits the `default` case and appends `action: reconnectCommand(...)` ("Removing the InsForge MCP server ... and adding it back is the usual remedy"). The platform's `error` here is arbitrary (`access_denied`, etc.). For a user who simply declined consent (`access_denied`), "reinstall the server" is misleading — retrying and approving is the fix. Consider a tailored override for this branch, or letting `humanFormOf` omit the command for codes where reinstall is not the remedy (mirroring the `server_error` treatment).

**Software engineering — no test asserts the route wiring, which is precisely where the original bug lived** — `src/http/oauth-error-response.test.ts`. The pure-function coverage of `humanFormOf` / `prefersHtml` / `reconnectCommand` is excellent and the property-over-codes assertion is the right instinct. But the bug #74 introduced was a *wiring* bug (page wired to one branch), and a pure-function test structurally cannot catch a future branch that forgets to call `sendOAuthError` — the two catch blocks above are live proof. A lightweight route-level test (supertest against the authorize/callback handlers asserting `Content-Type: text/html` for a browser `Accept`) would guard the exact regression class this PR is about. The author explicitly chose manual verification here; flagging as a durable-regression gap, not a blocker.

### Information

- **Stale example in the PR description** — the "What a person now reads on the live failure" block still shows `npx @insforge/install`, but the code (correctly) emits `npx add-mcp <url>` via `reconnectCommand`, and the test asserts it is *not* `@insforge/install` (`oauth-error-response.test.ts:52-58`). Description only; the code is right.
- **`prefersHtml` on absent/`*/*` Accept** — `req.accepts(['json','html'])` returns `json` when there is no `Accept` header or a bare `*/*`, so such requests get JSON. This is the correct default for curl/programmatic clients; noting only because a rare browser sending `Accept: */*` would then see JSON.

**Security:** No security-relevant regressions. Reflected `error`/`error_description` from platform query params reach the HTML page (`callback` platform-error branch → `humanFormOf` default → escaped and rendered), but `renderOAuthErrorPage` escapes `heading`/`message`/`action` via `escapeHtml` (`&`, `<`, `>`, `"`, `'`), so no reflected XSS. JSON bodies are unchanged, so no contract change for clients. No new secrets logged and no auth checks weakened.

**Performance:** No concerns — pure string formatting on an error path, no new I/O, queries, or hot-path work.

---

### Verdict: approved (informational)

No Critical findings. Solid, well-reasoned change that materially improves the failure UX; the suggestions above (especially the two unconverted `catch` blocks) would make the "every browser-reachable error" claim actually hold. Human approval still required via the separate approve flow.

jwfing (changes requested):
## Review — #84 Every browser-visible OAuth error says what to do

**Summary:** Solid, well-reasoned refactor that funnels OAuth error responses through a single `sendOAuthError` helper with per-code human wording, correct content negotiation, and proper HTML escaping — but one browser-reachable authorize branch was left returning raw JSON, which contradicts the PR's own completeness claim.

**Requirements context:** No `/docs/superpowers/` (or any `docs/` spec dir) exists in this repo — assessing against the PR description/title and the referenced #74/#82 alone. No spec/plan matched.

---

### Critical

**Functionality — `src/http/server.ts:421-427` — the authorize `catch` block still returns raw JSON to a browser.**
The entire `/oauth/authorize` GET is a browser navigation, so this branch is browser-reachable. If `oauthManager.createAuthorizationState` throws (e.g. Redis unavailable), the person's tab gets:
```json
{"error":"server_error","error_description":"Failed to initiate authorization"}
```
— exactly the failure this PR exists to eliminate. The description states *"Every browser-reachable error now goes through one function"* and enumerates the authorize branches, but this `server_error` path is not among the seven verified branches and was never converted to `sendOAuthError`. This reproduces the "guard the two consumers you noticed, leave the other eight" mistake the PR explicitly frames itself against. It is not a regression (it returned JSON before too), but it is a real gap against the PR's stated goal and title. Route it through the helper, e.g. `return sendOAuthError(req, res, statusForHttpError(error), { error: 'server_error', error_description: 'Failed to initiate authorization' });`.

### Suggestion

**Functionality — `src/http/oauth-error-response.ts:118-133` — the `default` branch attaches the reconnect command to every unhandled code.**
Via the callback platform-error branch (`server.ts:464-467`), `error` is whatever the platform put in the query string — e.g. `access_denied` (the user clicked "Deny") or any future code. Those fall to the `default` case, which returns `action: reconnectCommand(...)`. For `access_denied` the correct remedy is "start again and approve," not "remove and re-add the server," so the page would show a misleading `npx add-mcp` instruction. Consider defaulting `action` to `undefined` in the fallback and only attaching it where re-registration is genuinely the fix.

**Software engineering — test coverage doesn't guard the property the PR is actually about.**
The new tests cover `humanFormOf` / `prefersHtml` / `reconnectCommand` and one `sendOAuthError` negotiation case thoroughly, but nothing asserts that the authorize/callback *handlers* route each error branch through `sendOAuthError`. The bug being fixed (#74) was precisely a wiring omission, and the Critical above is a second one — both are invisible to unit tests of the pure helper. A route-level test that drives each authorize/callback error branch with an HTML `Accept` header and asserts an HTML (not JSON) body would guard exactly the "every branch is wired" property, and would have caught `server.ts:421-427`.

### Information

- **Security (no issue — noted as verified):** the reflected `error_description` from the callback query string flows into the page, but `renderOAuthErrorPage` escapes `heading`/`message`/`action` via `escapeHtml` (`& < > " '`), and the reconnect command is built from `SERVER_CONFIG.publicUrl` rather than the attacker-controlled `Host` header — so no reflected XSS and no host-header injection into the suggested command. Good, and the `publicUrl` rationale is well documented in the code.
- **`src/http/oauth-error-response.ts:86`** — `npx add-mcp <url>` is not documented anywhere else in the repo; `README.md` and `llms-install.md` only reference `npx @insforge/install`. The comments make a strong case that `add-mcp` is the correct *remote*-add command, but there's no in-repo source of truth. Worth confirming it's published/correct and adding it to the setup docs so the error page and the docs don't drift.
- **PR description nit:** the top example still shows the stale `npx @insforge/install` wording; the code and the cubic summary correctly use `npx add-mcp`. Code is right — description is just stale.
- **Performance:** no concerns — pure string building on error-only paths.

---

### Verdict: `request_changes`

One Critical: a browser-reachable authorize error branch (`server.ts:421-427`) still returns raw JSON, contradicting the PR's stated goal that every browser-visible OAuth error goes through `sendOAuthError`. Everything else is genuinely clean — the negotiation, escaping, `publicUrl` choice, and per-code wording are all well done, and the two Suggestions would make the fix complete and regression-proof.

jwfing (commented):
## Review: Every browser-visible OAuth error says what to do

**Summary:** Routes all `/oauth/authorize` and `/oauth/callback` error branches through a single `sendOAuthError` helper that serves an actionable HTML page to browsers and the unchanged OAuth JSON to programmatic clients — a well-scoped, well-tested fix for the "one branch was wired, nine weren't" bug. Zero blocking issues; one functional gap worth addressing before the non-`insforge.dev` cutover path is relied on.

### Requirements context
No `docs/superpowers/` (or `docs/specs`, `docs/plans`) directory exists in this repo — assessing against the PR description and linked context (#74/#82) alone. No invented requirements.

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — `reconnectCommand` hardcodes the remove target `insforge`, which only matches on the `mcp.insforge.dev` host.** `src/http/oauth-error-response.ts:107` returns `['npx add-mcp remove insforge -y', 'npx add-mcp ${mcpUrl}']`. I verified against add-mcp docs (`/neon-solutions/add-mcp`): `remove <name | url | package>` matches by the stored server name, and when you *add* by URL the name is inferred from the hostname (`https://mcp.context7.com/mcp` → `context7`). So on the canonical `mcp.insforge.dev` deployment the inferred name is `insforge` and the command works. But on the exact non-insforge host the test encodes — `keen-pulse-fsjr9.run.mcp-use.com` (`oauth-error-response.test.ts:87-93`) — the user's existing entry is named after *that* host (`mcp-use`/`keen-pulse-fsjr9`), so `remove insforge` matches nothing → the removal is a no-op → the following bare `add-mcp <url>` rewrites an identical entry and changes nothing. That is precisely the "same-hostname cutover no-op" the `remove`-first design exists to prevent, resurfacing on exactly the cutover hosts the dynamic `mcpUrl` was added to serve. The "names the host the person is actually talking to" test only asserts the generated *strings*, not that `remove` matches what `add` stored, so it gives false confidence here. Suggest deriving the remove target from the same URL (add-mcp's `remove` accepts a URL: `npx add-mcp remove ${mcpUrl} -y`), which is correct on every host. Not blocking: the production `mcp.insforge.dev` path works, and this is still strictly better than the pre-PR raw-JSON-with-no-command behavior.

#### Information

- **Functionality — the `default` branch of `humanFormOf` attaches the reconnect command to every unrecognized code.** `src/http/oauth-error-response.ts` default case. This is reached by the authorize `response_type` check (`unsupported_response_type`) and, more importantly, by the callback platform-error branch (`src/http/server.ts:464`) which forwards the platform's raw `error`. For codes like `access_denied` (the user declined consent) the remedy is "start again and approve," not "remove + re-add the server" — re-registering doesn't fix a declined consent. Consider an explicit `access_denied` case (or suppressing `action` for non-reregistration codes). Low blast radius; the description is still escaped and shown.
- **Software engineering — content negotiation and API compatibility preserved.** Verified `prefersHtml` via `req.accepts(['json','html'])` defaults to `json` when no `Accept` header is sent, so conforming MCP clients keep getting spec-compliant JSON — the HTML is genuinely additive. The `{ ...humanFormOf(...), ...human }` merge with explicit `action: undefined` correctly suppresses the code block for the "opened by hand" / expired-state branches. Good scope discipline: only the two browser-facing handlers changed; the `/mcp` transport 401s stay JSON.
- **Security — no concerns.** `mcpUrl` is built from `SERVER_CONFIG.publicUrl`, not the attacker-controlled `Host` header (correctly documented in the source); all rendered fields go through `escapeHtml`, including the platform-supplied `error_description` reflected via the callback; no secrets or tokens newly logged or returned.
- **Performance — no concerns.** Pure per-error string building on an already-terminal (rare) error path; no I/O, loops, or allocations of note.
- **Testing — strong.** The new `oauth-error-response.test.ts` asserts the wording *property* over the error codes rather than a single example (which is exactly the bug shape being fixed), plus content negotiation and the remove-before-add sequence. The pre-existing `templates/oauth-error.test.ts` still passes with the new array-based `action` rendering.

### Verdict
`approved` (informational — a human still gives the GitHub approval). No Critical findings. The Suggestion on `reconnectCommand`'s hardcoded `insforge` name is worth fixing before the non-`insforge.dev` cutover flow is depended upon, but it does not block: the production host works and the change is strictly better than the status quo on every branch.

jwfing (commented):
## Review: Every browser-visible OAuth error says what to do

**Summary:** Routes every browser-reachable `/oauth/authorize` and `/oauth/callback` error through a single `sendOAuthError` helper that serves a human HTML page to browsers while preserving the exact OAuth JSON for programmatic clients — a clean, well-tested consolidation that does what the PR claims.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessing against the PR description and the linked context (#74's single-branch page, #82's port-agnostic matching) alone. The core stated requirement — *"every browser-reachable error now goes through one function"* — is the yardstick I used. Note the prior review round's Critical (the authorize `try/catch` at `server.ts:407` returning raw JSON on a throw) is **addressed** in this PR; it now calls `sendOAuthError`.

The implementation is solid across all four dimensions. Findings below are all non-blocking.

---

### Critical
(none)

---

### Suggestion

**1. Functionality — one browser-reachable authorize branch still returns raw JSON** (`src/http/server.ts:347-350`)
The `readClientId` catch has a non-`InvalidClientIdError` branch that still does `res.status(500).json({ error: 'server_error', ... })` rather than `sendOAuthError`. This is exactly the "guard the consumers you noticed, leave the rest" pattern the PR argues against. In practice it's effectively unreachable — `isOAuthConfigured()` at `:297` guarantees a non-empty `INSFORGE_CLIENT_SECRET`, so `clientIdSigningKey()` can't throw and `readClientId` (`client-id.ts:241`) only throws `InvalidClientIdError` from there on — which is why I'm not blocking on it. But for consistency with the PR's own thesis, either route it through `sendOAuthError` too, or add a one-line comment noting it's unreachable-by-construction so the next reader doesn't re-flag it.

**2. Security/Functionality — user-controlled callback error values aren't string-coerced** (`src/http/server.ts:453-456`, `src/http/oauth-error-response.ts:humanFormOf` default case)
The platform-error branch is the only path feeding request-controlled values into the HTML page: `error` and `error_description` are taken straight from `req.query` and cast `as string`. Express 5's `qs` parser turns duplicate params into arrays, so a crafted `GET /oauth/callback?error=access_denied&error_description=a&error_description=b` (with an invalid `state`, so there's no `redirectUri` to bounce to) makes `humanFormOf`'s default `message: body.error_description` an **array**, and `escapeHtml(array)` throws. Express 5 auto-catches the async throw and returns a generic 500 — so this is a robustness/UX nit, **not** a crash or DoS — but the user then sees the bare 500 instead of the friendly page this PR exists to deliver. Recommend coercing query values to a single string (e.g. take the first element, or `String(x)`) before handing them to `sendOAuthError`. No XSS risk here — `renderOAuthErrorPage` escapes every field, and the template test at `oauth-error.test.ts:23-43` covers array items too.

**3. Software engineering — no test that `sendOAuthError` builds the URL from `publicUrl`** (`src/http/oauth-error-response.ts:181-195`)
The host-header-spoofing safeguard (build `mcpUrl` from `SERVER_CONFIG.publicUrl`, never the request Host) is the security-relevant part of `sendOAuthError`, but the browser test at `oauth-error-response.test.ts:22-26` only asserts `type('html')`/status and never checks the rendered command host. The property is well covered at the `reconnectCommand`/`humanFormOf` unit level, so this is minor — a small integration assertion that the sent HTML contains `publicUrl/mcp` would lock down the whole path end-to-end.

---

### Information

- **`/oauth/register` 500 now routes through `sendOAuthError`** (`src/http/server.ts:267`). `register` is a programmatic (non-browser) endpoint, so this is a slight scope expansion beyond "browser-visible." It's harmless — dynamic-registration clients send `Accept: application/json` or `*/*`, and `prefersHtml` returns the first list entry (`json`) for `*/*`, so they still get JSON — just noting it's technically outside the stated scope.
- **Machine vocabulary can leak into the default-case message.** `humanFormOf`'s default uses `body.error_description` verbatim, so `unsupported_response_type` (`:331`) surfaces `"Only response_type=code is supported"` to a human. The property test asserts no underscores in the *heading* but not the *message*, so this isn't caught. Only reachable by a misconfigured client, so cosmetic.

---

### Dimension coverage
- **Software engineering** — Strong. Pure exported helpers tested without standing up Express, property-based assertions over error codes (not one example), content-negotiation and remove-before-add behavior all covered. Follows repo conventions: `.js` import extensions, `vitest`, the house style of dense rationale comments. Existing `oauth-error.test.ts` still passes under the `string | string[]` template refactor.
- **Functionality** — Solves the stated bug; the redirect-mismatch branch a real user hits (`:374`) and the whole-handler throw (`:413`) now render the page. JSON preserved byte-for-byte for clients. Two small gaps above.
- **Security** — Good. Repair command built from `publicUrl` (not attacker-controlled Host) — correctly documented; all HTML fields escaped; no secrets/tokens logged or returned. One input-type-hardening nit (#2).
- **Performance** — No concerns. Pure synchronous string building, no new I/O, loops, or hot-path work.

---

### Verdict
**approved** (informational — human approval is a separate action). No Critical findings; the three Suggestions and two Information notes are all non-blocking and safe to address in this PR or a follow-up.

jwfing (commented):
## Review: Every browser-visible OAuth error says what to do

**Summary:** Routes all browser-reachable OAuth error paths through a single `sendOAuthError` helper that returns HTML to browsers and unchanged JSON to programmatic clients — a well-executed, well-tested change that closes the #74 gap, with one remaining symmetric branch worth converting.

### Requirements context
This repo has **no `/docs/superpowers/` (or any spec/plan dir)** — confirmed on the head checkout. Assessing against the PR description, the linked #74 intent, and the surrounding code alone. The PR's stated invariant is explicit and testable: *"Every browser-reachable error now goes through one function."* I checked the implementation against that claim.

---

### Critical
(none)

---

### Suggestion

**Functionality — one browser-reachable branch still returns raw JSON (`src/http/server.ts:561-572`).**
The callback's outer `catch` still does `res.status(...).json({ error: 'server_error', ... })`. `/oauth/callback` is *always* a browser navigation (the PR body says exactly this), so an exception during token exchange — a `fetch` throw, a Redis failure — lands raw JSON in the user's tab. This is directly symmetric to the **authorize** outer catch at `server.ts:405-417`, which this same PR *did* convert to `sendOAuthError` (and the diff comment even calls "the tenth consumer… whose whole point was not stopping at the ones I noticed"). Converting `:568` the same way would make the completeness claim actually hold. Not blocking — it's pre-existing behavior, not a regression, and only fires on rare server-side exceptions — but it's the one place the invariant leaks.

**Software engineering — the completeness invariant isn't test-guarded.** `humanFormOf`/`reconnectCommand`/`prefersHtml` are tested thoroughly (the property-over-error-codes approach is the right call). But the route wiring is not: neither the `{ ...humanFormOf(body), ...human }` partial-override merge in `sendOAuthError` (`oauth-error-response.ts:200-210`) nor the per-branch `human` overrides in `server.ts` are exercised. A test asserting each authorize/callback branch's *actual* content-negotiated output is precisely what would have caught the `:568` gap above — i.e. the exact regression class this PR targets. Worth adding a thin route-level or handler-level assertion per branch.

---

### Information

- **Functionality (`oauth-error-response.ts:172-179`).** The `default` case attaches `reconnectCommand` to *every* unknown code, including `access_denied`. "Remove and re-add the server" is the wrong remedy for a user who deliberately clicked Deny — "try again and approve" fits better. Low impact: `access_denied` normally bounces back to the client via the registered redirect (`server.ts:438-448`) and only reaches this page when there's no `redirectUri`.
- **Security (positive).** Reflected platform input (`error_description` from the callback query) now reaches the HTML page where it was previously JSON-only, but it is correctly escaped via `escapeHtml` in the template (`templates/oauth-error.ts:99,108-115`) — no XSS. Using `SERVER_CONFIG.publicUrl` rather than the `Host` header for the reconnect command (`oauth-error-response.ts:203`) correctly avoids host-header spoofing; the extensive rationale comment is appreciated. No secrets logged, no auth checks weakened, no new dependencies.
- **Performance.** No concerns — string building per error, no new I/O, no DB access, nothing in a hot path.
- **Software engineering (positive).** Follows repo conventions (`.js` ESM imports, named exports, vitest); the narrowed `AcceptNegotiable` interface keeps the helper testable without an express mock; content negotiation defaults correctly to JSON when no `Accept` header is present.

---

### Verdict: **approved** (informational — human approval is separate)
No Critical findings. The change is a real improvement that does what it claims for 7 of the 8 browser-reachable branches; the remaining callback outer-catch (`:568`) and the missing per-branch route tests are non-blocking Suggestions. Posting as a comment, not a block.

jwfing (commented):
## Review — Every browser-visible OAuth error says what to do

**Summary:** A clean, well-reasoned consolidation that routes every browser-reachable OAuth error (authorize / callback / register) through a single `sendOAuthError` helper — HTML for browsers, unchanged JSON for programmatic clients — and fixes the reconnect instructions to be host-exact and remove-before-add. I found no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory (gates are vitest + tsc + eslint + tsup). Assessed against the PR description, the referenced #74 bug ("actionable page wired to exactly one branch"), and the surrounding OAuth route code.

### Findings

#### Critical
(none) — the change is additive for JSON clients (spec body is byte-unchanged), the reconnect URL is derived from `SERVER_CONFIG.publicUrl` rather than the attacker-controlled Host header (`oauth-error-response.ts:190-192`, matching the discovery-doc rationale), all user-facing strings pass through `escapeHtml` (`templates/oauth-error.ts:100-102`, `108-115`), and the prior Critical (the authorize outer `catch` returning raw JSON) is now routed through the helper (`server.ts:408-415`). `normaliseBaseUrl` strips trailing slashes so `mcpUrl` composes cleanly (`config.ts:38-40`).

#### Suggestion
- **[software engineering] No route-level regression test for the actual bug class.** `oauth-error-response.test.ts` thoroughly covers the pure units (`humanFormOf`, `reconnectCommand`, `prefersHtml`, `sendOAuthError` negotiation), and the PR argues correctly that the wording is the payload. But the *defect being fixed* was structural — a route branch that forgot to call the helper. A unit test on `humanFormOf` cannot catch a future edit that reverts one `server.ts` branch back to `res.status(...).json(...)`. A thin supertest over `GET /oauth/authorize` and `GET /oauth/callback` with `Accept: text/html`, asserting an HTML body on each failure branch, would guard exactly the regression this PR exists to prevent. Ref: `src/http/oauth-error-response.test.ts` (whole file) and `src/http/server.ts:295-420, 424-577`.
- **[security] Attacker-controlled platform error text is reflected on the branded page.** The callback platform-error branch passes no `human` override (`server.ts:450-456`), so `humanFormOf` falls to `default` and renders `body.error_description`. For `/oauth/callback`, `error`/`error_description` come straight from `req.query` on an unauthenticated GET, so an attacker can craft `…/oauth/callback?error=x&error_description=<prose>` and hand it to a victim. It **is** HTML-escaped (`templates/oauth-error.ts:100-102`), so this is not XSS — but a styled `insforge.dev` page now reflects arbitrary attacker prose next to a legitimate-looking `npx add-mcp` command, which is a modest phishing upgrade over the old raw-JSON body. Consider whitelisting known error codes in the `default` branch (or not echoing `error_description` verbatim for the callback platform-error path). Ref: `src/http/oauth-error-response.ts:135-160`, `src/http/server.ts:450-456`.

#### Information
- **[functionality]** The callback platform-error branch (`server.ts:453-456`) routes every platform error through `humanFormOf`'s `default` case, which attaches a `reconnectCommand`. For a deliberate `access_denied` (user cancelled), "remove and re-add the server" is misleading advice. Minor wording nit — an explicit `human` override for that path would be more accurate. Ref: `oauth-error-response.ts:154-159`.
- **[style]** `error as string` / `error_description as string | undefined` (`server.ts:454-455`) assume single-valued query params; duplicate params (`?error=a&error=b`) yield arrays under Express. Harmless here and consistent with the pre-existing pattern in this file, so noted only for completeness.
- **[performance]** No concerns — all changed code executes only on OAuth error paths; no hot-path, loop, allocation, or query changes.

### Verdict
`approved` (informational). Zero Critical findings; posted as a COMMENT. The Suggestions (a route-level regression test and trimming reflected `error_description`) are worth addressing but non-blocking. Nice work closing the loop on the earlier authorize-`catch` Critical and covering all eight branches.

jwfing (approved):
LGTM - approved.

PR #85, opened 2026-07-31
fix(verify): probe authorize with a complete request, not a bare one

tonychang04:
## Changes

`verify-deploy.mjs` scored **PASS** on a deployment where no user could log in. This adds the check that catches it.

## Why

Both existing authorize assertions are answered by the **argument-validation layer**, which runs before the handler touches its client store. A parameterless probe 400s on missing params and never reaches the dependency.

Live, on the Manufact slug (`keen-pulse-fsjr9.run.mcp-use.com`, master @ `d78a33d`, no Redis):

```
GET  /oauth/authorize            -> 400 invalid_request       "PASS the authorize endpoint is reachable and validating"
POST /oauth/register {}          -> 400 invalid_client_metadata
```
Textbook RFC 4xx — reads as healthy. But with a **valid** body:
```
POST /oauth/register {redirect_uris:[...], ...}   -> 500 MaxRetriesPerRequestError (ioredis)
GET  /oauth/authorize?<full params + PKCE>        -> 500 same
```

## The new check

Send a request complete enough to reach the store, with a client id that cannot exist:

```
400 invalid_client -> store reached, id genuinely unknown   (healthy)
500                -> the lookup itself failed              (broken)
```

Registers nothing, so the script keeps its read-only contract.

## Testing

Verified in both directions against live deployments:

| target | store | result |
|---|---|---|
| `keen-pulse-fsjr9.run.mcp-use.com` (no Redis) | broken | **FAIL** — 23/27 |
| `mcp.insforge.dev` (EC2, Redis-backed) | healthy | **PASS** — 16/25 |

## Backwards Compatibility

Not breaking. Adds one assertion; total count 26 -> 27.

---
## Summary by cubic
Fixes verify-deploy by sending a complete `/oauth/authorize` request so it actually hits the client store. Prevents false PASS results when the backing store (e.g., Redis) is down.

- **Bug Fixes**
  - Add full authorize probe (PKCE + fake client_id) to reach the store.
  - 400 invalid_client = store reachable (healthy); 500 = lookup failed (broken).
  - Read-only check; no registration or state changes.

<sup>Written for commit dab2eaec2eb53a9e293ef7d3f9b9fd60b1b43de0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/85?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/verify-deploy.mjs
+++ b/scripts/verify-deploy.mjs
@@ -222,6 +222,38 @@ check(
   `status ${authorize.status}`
 );
 
+// The two checks above are both satisfied by the argument-validation layer,
+// which answers before the handler touches its client store. That is exactly
+// how this script scored a PASS on a deployment where no user could log in:
+// registration and authorize were both 500ing on
+// `MaxRetriesPerRequestError (ioredis)`, and a parameterless probe never got
+// far enough to see it.
+//
+// So send a request that is complete enough to reach the store. The client id
+// is deliberately one that cannot exist: a server whose store is healthy looks
+// it up, misses, and says so; a server whose store is unreachable cannot get
+// that far. Both outcomes are 4xx-vs-5xx, and neither registers anything, so
+// the script stays read-only.
+//
+//   400 invalid_client -> store reached, id genuinely unknown  (healthy)
+//   500                -> the lookup itself failed             (broken)
+const probeAuthorize = await getJson(
+  '/oauth/authorize?response_type=code' +
+    '&client_id=verify-deploy-probe-not-a-real-client' +
+    '&redirect_uri=' + encodeURIComponent('http://127.0.0.1:33418/callback') +
+    '&code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM' +
+    '&code_challenge_method=S256' +
+    '&state=verify-deploy'
+);
+check(
+  'a fully-formed authorize request reaches the client store',
+  probeAuthorize.status !== 500,
+  probeAuthorize.status === 500
+    ? 'authorize 500s on a complete request — the client store is unreachable, ' +
+      'so no login can complete (check Redis/backing store)'
+    : `status ${probeAuthorize.status}`
+);
+
 // --- follow the chain to whichever AS the resources name -------------------
 // This is the client's next hop, so it gets checked whether the AS is us or the
 // platform. RFC 8414 §3.1 derives the URL the same way RFC 9728 does — insert


jwfing (commented):
## Review — fix(verify): probe authorize with a complete request, not a bare one

**Summary:** Adds one read-only assertion to `verify-deploy.mjs` that sends a fully-formed `/oauth/authorize` request with a non-existent `client_id`, so a deploy whose client store (Redis) is unreachable fails the gate instead of scoring a false PASS on the argument-validation layer — a real, well-motivated fix that is correctly implemented.

### Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` (or `specs/`/`plans/`) directory. Assessed against the PR description, the surrounding script, and the actual `/oauth/authorize` handler in `src/http/server.ts`.

I verified the core claim against the handler (`src/http/server.ts:303-432`): a request with `client_id` + `redirect_uri` + `response_type=code` present clears the arg-validation branches (lines 313, 323) and reaches `await redis.get(clientKey)` at line 333. On a healthy store the unknown client returns `400 invalid_client` (line 352); the `await redis.get` at line 333 is *not* wrapped in a try/catch, so a store failure rejects and surfaces as a 500 — exactly the 4xx-vs-5xx split the PR relies on. The `accept: application/json` header set by `getJson` also keeps the response on the JSON branch rather than the HTML error page, so the status code is preserved. The logic holds.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — false-green if a store outage manifests as a hang rather than a fast 500** (`scripts/verify-deploy.mjs:248-255`). The assertion fails only on exactly `500` (`probeAuthorize.status !== 500`). If the backing store hangs past `VERIFY_TIMEOUT_MS`, `getJson` aborts and returns `status: 'unreachable (…)'` (a string), which is `!== 500` → the check PASSES. The documented failure mode (`MaxRetriesPerRequestError`) fails fast as a 500, and a wholly-unreachable host is already caught by the `/health` early-exit, so blast radius is low — but consider also failing when the status is non-numeric/`unreachable`, or asserting the positive case (`=== 400`) directly. Asserting `=== 400` would additionally catch a stray `2xx`/`3xx` from a proxy or route change on `/oauth/authorize` — the same class of regression the author deliberately guards against with `=== 400` on the sibling "reachable and validating" check two lines up (line 221). Optional, since that sibling check partially covers it.

#### Information
- **Functionality — message misattributes a missing-credentials failure to the store.** `isOAuthConfigured()` returns 500 at `src/http/server.ts:306-311` *before* the store is touched, so when OAuth creds are absent this new probe also FAILs with `"the client store is unreachable … (check Redis/backing store)"` (`scripts/verify-deploy.mjs:251-254`). The cause is missing credentials, not Redis. Cosmetic only — the preceding `OAuth credentials are configured` check (line 211-215) already reports the accurate reason in that scenario, so both fire and the true cause is visible.
- **Software engineering — no automated test.** There is no unit/integration test exercising the new assertion, but this is consistent with the repo: the sibling `scripts/verify-handshake.mjs` has none either, and `scripts/` is operational tooling that probes live deployments rather than app code. The PR documents bidirectional manual verification (broken deployment → FAIL 23/27; healthy → PASS), which is the appropriate evidence for a script of this kind. Not a blocker.

#### Security
No security-relevant concerns. The probe registers nothing (read-only contract preserved), uses a synthetic `client_id` and a loopback `redirect_uri`, and the `code_challenge` is the well-known RFC 7636 example value — no secrets or PII introduced.

#### Performance
No concerns. One additional lightweight HTTP GET in a deploy-gate script; each request already carries the shared 10s deadline.

### Verdict
**approved** (informational — human approval is still a separate action). No Critical findings; the two Suggestion/Information items are non-blocking refinements. Clean, well-reasoned, appropriately scoped fix with strong explanatory comments and documented live verification.

jwfing (approved):
LGTM - approved.

tonychang04:
Superseded — #82 landed an equivalent parameterised probe independently, so this is redundant against current master. But that probe has the same blind spot one layer deeper: a made-up client id now fails signature verification and returns 400 before reaching the authorization-state write, so it reports PASS on the Manufact deployment where no login can start. Replaced by a version that mints a real id first.

jwfing (commented):
## Review: fix(verify): probe authorize with a complete request

**Summary:** Adds a single fully-formed `/oauth/authorize` probe to `verify-deploy.mjs` so the deploy gate reaches the client store and can no longer score PASS when the backing store (Redis) is down — a correct, well-scoped, read-only fix.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/`) directory. Assessed against the PR description, the live evidence in the body, and the surrounding code (`src/http/server.ts`).

I verified the core premise directly against the handler:
- `src/http/server.ts:303-356` — validation order is `isOAuthConfigured()` → 500 (existing "credentials" check), missing `client_id/redirect_uri/response_type` → `400 invalid_request` (the *bare* probe stops here), `response_type` → 400, **then `redis.get(clientKey)` at line 333** (the store hit), then null → `400 invalid_client`. So a request carrying `response_type=code` + `client_id` + `redirect_uri` genuinely reaches the store, exactly as the PR claims.
- `express ^5.1.0` — async handler rejections auto-forward to the error handler → `500`, so the "broken store = 500" direction holds.
- `src/http/redis.ts:30` — `maxRetriesPerRequest: 3` makes an unreachable store reject *fast* (`MaxRetriesPerRequestError`) rather than hang, so the probe observes a clean 500, matching the author's live capture.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — assertion is looser than its sibling (`scripts/verify-deploy.mjs:248-255`).** The bare-probe check just above (`:216-223`) tightly asserts `authorize.status === 400`, but the new probe only asserts `probeAuthorize.status !== 500`. For a healthy store the response is deterministically `400 invalid_client` (`server.ts:334-356`), so tightening this to `=== 400` (or at least a 4xx test) would both match the sibling's rigor and close a residual false-PASS gap: if this endpoint *stalls* past `VERIFY_TIMEOUT_MS`, `getJson` returns `status: 'unreachable(...)'` (a string), and `'unreachable(...)' !== 500` evaluates **true → PASS** — the very "green on a broken deploy" failure mode this PR exists to kill. In the observed/likely failure mode (fast `MaxRetriesPerRequestError` → 500) the check works correctly, which is why this is a Suggestion and not a blocker, but the fast-fail is what makes it work, not the assertion.

#### Information
- **Software engineering — no automated test added, and that is consistent with repo convention.** `scripts/*.mjs` are standalone operational scripts with no test harness (`test` = `vitest run --passWithNoTests`; neither `verify-deploy.mjs` nor `verify-handshake.mjs` is covered), and the change was validated live in both directions (broken store → FAIL, healthy → PASS) per the PR body. Noting for completeness, not requesting a test.
- **Security — no concerns.** Read-only: `createAuthorizationState` (`server.ts:407`) runs only *after* a successful client lookup, which the deliberately non-existent `client_id` never passes, so the probe writes/registers nothing and the read-only contract holds. No secrets, no new dependencies, fixed PKCE value is the RFC 7636 example (harmless).
- **Performance — no concerns.** One extra `GET` per deploy-gate run; negligible.

### Verdict
**approved** (informational — a human still gives the GitHub approval via the approve flow). Zero Critical findings; the one Suggestion above is worth a quick tightening to `=== 400` but does not block merge.

PR #86, opened 2026-07-31
fix(verify): start the login with a minted client id, not a made-up one

tonychang04:
## Changes

The parameterised `/oauth/authorize` probe that landed with #82 is still short of the state write it means to test — so **the gate currently reports `PASS  the client lookup on /oauth/authorize works (a login could start)` against a deployment where no login can start.**

## Why

Signed client ids changed where the request stops. An unregistered id now fails signature verification and returns `400 invalid_client` *before* the authorization-state write. Measured against the live Manufact deployment (master @ `33fee08`, no Redis):

```
client_id=verify-deploy-probe (made up)   -> 400 invalid_client        <- reads as healthy
client_id minted by /oauth/register       -> 500 Failed to initiate…   <- the truth
```

The 500 is `oauth-manager.ts:139`, `redis.setex('mcp:auth:state:'…)` — #82 took Redis out of *registration*, not out of the flow.

## The fix

Mint an id via `/oauth/register`, then authorize as that client. Post-signed-ids **registration stores nothing server-side — the id IS the registration** — so this keeps the script's read-only contract while reaching the first step that genuinely needs backing state.

Also adds a check that registration itself works, which nothing covered before.

## Testing

Verified in both directions against live deployments:

| target | backing state | result |
|---|---|---|
| `keen-pulse-fsjr9.run.mcp-use.com` (Manufact, no Redis) | unreachable | **FAIL** — 24/28 |
| `mcp.insforge.dev` (EC2, Redis-backed) | healthy | **PASS** — 17/26 |

## Backwards Compatibility

Not breaking. Replaces one assertion with two; total 27 -> 28.

Supersedes #85 (closed).

---
## Summary by cubic
Fixes the deploy verification to start OAuth with a `client_id` minted via `/oauth/register`, reaching the authorization state write and correctly detecting missing backing state. Adds a gateway bypass for `/health` via `VERIFY_HEALTH_URL` to assert the actual deployed build.

- **Bug Fixes**
  - Use minted `client_id` from `/oauth/register` when calling `/oauth/authorize`.
  - Add `postJson()` helper for JSON POST requests.
  - Validate registration returns `200/201` with a `client_id`; surface `500` when state is unavailable (avoid false PASS from `400 invalid_client`).
  - Support `VERIFY_HEALTH_URL` to read `/health` behind the gateway and assert the deployed version.

<sup>Written for commit ce9ada543581277df0d475a10cb445df3d8e75c4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/86?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/scripts/verify-deploy.mjs
+++ b/scripts/verify-deploy.mjs
@@ -95,6 +95,28 @@ async function getJson(path) {
   return { status: res.status, body, headers: res.headers };
 }
 
+async function postJson(path, payload) {
+  let res;
+  try {
+    res = await fetch(base + path, {
+      method: 'POST',
+      headers: { 'content-type': 'application/json', accept: 'application/json' },
+      body: JSON.stringify(payload),
+      signal: deadline(),
+    });
+  } catch (error) {
+    return { status: unreachable(error), body: undefined, headers: undefined };
+  }
+  const text = await res.text();
+  let body;
+  try {
+    body = JSON.parse(text);
+  } catch {
+    body = undefined;
+  }
+  return { status: res.status, body, headers: res.headers };
+}
+
 // --- health, and the deployed version -------------------------------------
 const health = await getJson('/health');
 check('GET /health is 200', health.status === 200, `status ${health.status}`);
@@ -230,21 +252,54 @@ check(
 
 // The bare probe above stops at the missing-parameter check, which is BEFORE
 // the client lookup — so it passes on a server that cannot log anyone in. A
-// probe carrying parameters reaches the lookup, and an unknown client there is
-// a 400 `invalid_client`. A 500 means the lookup itself failed: no Redis, or
-// no signing key. This is the difference between "the routes exist" and "a
-// login could start", and everything else in this file only proves the former.
-const authorizeWithParams = await getJson(
-  '/oauth/authorize?client_id=verify-deploy-probe' +
-    '&redirect_uri=http%3A%2F%2F127.0.0.1%3A1%2Fcb&response_type=code'
-);
+// probe carrying parameters gets further.
+//
+// But a MADE-UP client id is not far enough, and this is the second time this
+// file has been fooled by the same shape. Since signed client ids landed, an
+// unregistered id fails signature verification and returns 400 `invalid_client`
+// — still short of the authorization-state write. Measured against the Manufact
+// deployment, which cannot log anyone in:
+//
+//   client_id=verify-deploy-probe (made up)  -> 400   <- reads as healthy
+//   client_id minted by /oauth/register      -> 500   <- the truth
+//
+// So mint one. Post-signed-ids `/oauth/register` stores nothing server-side —
+// the id IS the registration — so this keeps the script's read-only contract
+// while reaching the first thing in the flow that actually needs backing state.
+const registration = await postJson('/oauth/register', {
+  client_name: 'verify-deploy probe',
+  redirect_uris: ['http://127.0.0.1:1/cb'],
+  grant_types: ['authorization_code'],
+  response_types: ['code'],
+});
 check(
-  'the client lookup on /oauth/authorize works (a login could start)',
-  authorizeWithParams.status === 400,
-  authorizeWithParams.status === 500
-    ? 'status 500 — the client lookup failed; the remaining Redis uses are not gone yet'
-    : `status ${authorizeWithParams.status}`
+  'a client can register (/oauth/register)',
+  registration.status === 200 || registration.status === 201,
+  registration.body?.error_description || `status ${registration.status}`
 );
+
+const mintedId = registration.body?.client_id;
+if (mintedId) {
+  const authorizeAsClient = await getJson(
+    '/oauth/authorize?response_type=code' +
+      `&client_id=${encodeURIComponent(mintedId)}` +
+      '&redirect_uri=' + encodeURIComponent('http://127.0.0.1:1/cb')
+  );
+  check(
+    'a real registered client can start a login (/oauth/authorize)',
+    authorizeAsClient.status !== 500,
+    authorizeAsClient.status === 500
+      ? `status 500 — ${authorizeAsClient.body?.error_description || 'authorize failed'}; ` +
+        'the flow still needs backing state it cannot reach (remaining Redis uses)'
+      : `status ${authorizeAsClient.status}`
+  );
+} else {
+  check(
+    'a real registered client can start a login (/oauth/authorize)',
+    false,
+    'not attempted — registration did not return a client_id'
+  );
+}
 
 // --- follow the chain to whichever AS the resources name -------------------
 // This is the client's next hop, so it gets checked whether the AS is us or the
--- a/scripts/verify-deploy.mjs
+++ b/scripts/verify-deploy.mjs
@@ -118,7 +118,47 @@ async function postJson(path, payload) {
 }
 
 // --- health, and the deployed version -------------------------------------
-const health = await getJson('/health');
+//
+// On Manufact the mcp-use gateway answers /health ITSELF — a fixed
+// {"status":"healthy","timestamp":…} returned whether or not a container is
+// running behind it. Our handler never reaches the client, so the version and
+// session numbers are invisible and "what is actually deployed?" becomes
+// unanswerable from outside. That is precisely the blindness that let
+// production sit on a 139-day-old build.
+//
+// The app underneath is reachable directly: the platform exposes the Fly
+// hostname as `config.internalUrl` on the server record
+// (https://mcp-<serverId>-<deployNo>.fly.dev), and it serves OUR /health. Point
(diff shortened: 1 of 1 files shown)


jwfing (commented):
## Review: fix(verify): start the login with a minted client id, not a made-up one

**Summary:** A correct, well-reasoned fix that replaces a probe fooled by signed-client-id semantics (made-up id → `400 invalid_client`, reads as healthy) with a register-then-authorize probe that actually reaches the authorization-state write, closing a false-green in the deploy gate.

### Requirements context
No `/docs/superpowers/` or `docs/specs/` directory exists in this repo (consistent with insforge-mcp's shape — it has no spec/plan tree). Assessed against the PR description, the code paths in `src/http/server.ts` / `src/http/oauth-manager.ts`, and the file's own contract docstring.

I verified the PR's central claims against the source at head:
- `/oauth/register` returns **201** with a `client_id` and stores nothing server-side (the id *is* the registration) — `src/http/server.ts:263-298`. The check accepting `200 || 201` is correct.
- A made-up id at `/oauth/authorize` fails signature verification → **`400 invalid_client`** — `src/http/server.ts:337-373`. This is exactly the shape the old assertion mistook for health.
- A minted id passes signature + exact redirect_uri match (`http://127.0.0.1:1/cb` is loopback-plaintext, which `mintClientId` allows) and reaches `createAuthorizationState` → `redis.setex(...)` at `src/http/oauth-manager.ts:139`. On Redis failure this throws and `statusForHttpError` returns **500** (`src/http/error-status.ts:9-19`), so the new `status !== 500` assertion fails in the correct (broken) direction. Claims check out.

### Findings

#### Critical
(none)

#### Suggestion

- **[software engineering / documentation] Module docstring is now stale — `scripts/verify-deploy.mjs:6-9`.** The header still states *"Read-only ... It registers nothing and writes nothing."* That is no longer true: the script now POSTs `/oauth/register`, and on a healthy deployment the minted-client authorize triggers a Redis auth-state write via `createAuthorizationState` → `setex` (`oauth-manager.ts:139`, a 10-minute-TTL row). The behavior is intentional and harmless (throwaway client, short-lived state), but the stated contract other maintainers rely on now contradicts the code. Update the docstring to describe the throwaway registration + short-lived state write.

- **[functionality / robustness] `getJson` follows redirects, so the authorize assertion reflects the upstream AS, not our server — `scripts/verify-deploy.mjs:283-295`.** `getJson` uses `fetch` with the default `redirect: 'follow'`. A *healthy* `/oauth/authorize` returns `302` to the upstream InsForge AS (`server.ts:412`), which `fetch` will follow — so `authorizeAsClient.status` is the upstream response's status, not our `302`. Consequences: (a) an upstream `5xx` while our server is fine would trip `status !== 500` and print the misleading *"remaining Redis uses"* detail; (b) the probe fires a real authorize request at the external AS. This is safe-direction (it can only produce a false FAIL, never a false PASS on a broken deploy — the only way to see a bare `500` here is our own server), which is why it's a Suggestion, not blocking. Consider `redirect: 'manual'` and asserting a `3xx` so the check reflects *this* server's decision. (Pre-existing `getJson` calls never surfaced this because they can't legitimately 302; this is the first probe that can.)

#### Information

- **[software engineering / test coverage] No automated test for the script.** Consistent with repo convention — `scripts/**/*.mjs` are run directly (`npm run verify:deploy`), not unit-tested, and `vitest` runs with `--passWithNoTests`. Verification is manual and is documented in the PR against two live deployments (Manufact → FAIL, `mcp.insforge.dev` → PASS). Acceptable given conventions; noting for completeness. `postJson` mirrors the existing `getJson` shape (deadline, unreachable-as-value, JSON-or-undefined), and eslint scopes `scripts/**/*.mjs` with the needed globals, so lint/format are unaffected.
- **[security] No concern — net neutral.** The throwaway client uses a loopback redirect_uri; `client_name` is logged server-side via `JSON.stringify` (existing log-injection guard, `server.ts:284-288`); no secrets, tokens, or PII are handled, logged, or emitted by the new code.
- **[performance] Negligible.** Two extra sequential HTTP calls added to a deploy gate, each bounded by `TIMEOUT_MS` (`deadline()`). No hot path.

### Verdict
**approved** (informational; the human still approves via the approve flow). Zero Critical findings — the fix is correct and the two Suggestions are quality/robustness improvements the author may address or defer.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review — `fix(verify): start the login with a minted client id`

**Summary:** Sound fix — the deploy probe now mints a real signed `client_id` via `/oauth/register` and authorizes as it, so it reaches the authorization‑state write and correctly fails against a deployment with no backing state (no false `PASS`); a few non‑blocking accuracy/clarity notes below.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, and there is no per-script spec — assessing against the PR description, the linked history (#82 signed client ids, #83 boot‑without‑Redis), and the actual handler behavior. I verified the core claims against source:
- `/oauth/register` mints a self-contained signed id and stores nothing server-side (`src/http/client-id.ts:190-232`, handler `src/http/server.ts:247-299`, returns `201` + `client_id`) — the "registration stores nothing" premise holds.
- `/oauth/authorize` with a *valid* minted id + matching `redirect_uri` reaches `createAuthorizationState`, which does `redis.setex(...)` (`src/http/oauth-manager.ts:139`). With no Redis that throws and the catch returns `statusForHttpError(err)` → **500** for any non-`InsforgeApiError` (`src/http/error-status.ts:9-19`). So the new `status !== 500` assertion is a terminal, reliable detector of the failure mode — the gate cannot green-wash a broken deploy. ✅

### Findings

#### Critical
(none)

#### Suggestion

1. **Functionality — the authorize probe follows the 302 and scores the *platform's* status, not the MCP server's** (`scripts/verify-deploy.mjs:322-335`). `getJson` (`:81-96`) uses `fetch` with the default `redirect: 'follow'`. On a healthy server the authorize handler returns a `302` to `${INSFORGE_CONFIG.apiBase}/api/oauth/v1/authorize` (`src/http/server.ts:412`), so the assertion actually issues a live request to the InsForge platform's authorize endpoint and measures *its* status. The intended failure (no Redis → terminal `500`, not a redirect) is still caught correctly, so the gate direction is safe; but two edges make the check measure more than it claims: (a) it now depends on the upstream platform being healthy, and (b) if that followed hop is unreachable, `getJson` returns an `"unreachable(...)"` string which satisfies `!== 500` and reads as `PASS`. Consider `fetch(..., { redirect: 'manual' })` for this one probe so you observe the `302` from the server under test directly (e.g. assert `status === 302 || (status !== 500 && status < 400)`).

2. **Accuracy — the file's "writes nothing" contract is now stale** (`scripts/verify-deploy.mjs:6-9`). The header still says *"It registers nothing and writes nothing."* The PR's justification correctly covers `/oauth/register` (no server-side write), but the second probe drives `createAuthorizationState`, which writes a TTL'd auth-state row to Redis on a healthy deployment (`src/http/oauth-manager.ts:139`). Blast radius is negligible (an orphan state that auto-expires and is never completed), but the top docstring is a deliberately emphasized property future readers will rely on — please update it to reflect that the login probe now writes a short-lived state row.

#### Information

3. **Software engineering — no automated test for the script, which is expected here.** `scripts/**/*.mjs` are explicitly "standalone operational scripts" run directly by node and are outside vitest/tsc (`eslint.config.js:11-13`); this file is itself a live-deployment probe, so the absence of a unit test is consistent with repo convention rather than a gap to fix. Coding standards, ESM import style, and the declared script globals (`fetch`/`URL`/`AbortSignal`) all match.

4. **Software engineering — minor duplication.** The `VERIFY_HEALTH_URL` inline IIFE (`scripts/verify-deploy.mjs:141-160`) re-implements the fetch + `text()` + `JSON.parse` dance that `getJson` (`:81-96`) and the new `postJson` (`:98-118`) already do, differing only in base URL. A small `getJsonFrom(origin, path)` helper would collapse all three. Non-blocking.

5. **Security / performance — no concerns.** No secrets logged or returned (`VERIFY_HEALTH_URL` echoed at `:161` is a hostname, not a credential); `postJson` sends only client metadata; no new user input reaches SQL/shell; no auth check weakened; no new dependencies; two extra HTTP calls in a gate script are immaterial.

### Verdict

**`approved`** (informational — human approval is still a separate action). The fix is correct, in-scope (one file, replaces one assertion with two as described), and achieves its stated goal of detecting a deployment that cannot start a login. All five items above are Suggestion/Information — none block merge.

jwfing (approved):
LGTM - approved.

PR #87, opened 2026-07-31
Authorization state as a sealed value, not a Redis row

tonychang04:
The Redis use that stops a login from *starting*, and the one Max measured on
the slug an hour ago:

```
GET /oauth/authorize  500  {"error":"server_error",
                            "error_description":"Failed to initiate authorization"}
oauth-manager.ts      createAuthorizationState -> getRedisClient()
```

A client could register (thanks to #82) and then never open a login.

## Sealed, not signed — the one real difference from client ids

| | client id (#82) | auth state (this) |
|---|---|---|
| contents | public registration | includes `insforgeCodeVerifier` |
| mechanism | HMAC signature | **AES-256-GCM encryption** |

The verifier is our PKCE secret for the platform call. A signed-but-readable
envelope would print it in the authorize URL, the browser's address bar and the
platform's logs — **precisely the party PKCE defends against.** The ceremony
would remain and the protection would not.

- fresh 96-bit nonce per seal (a repeated nonce is how GCM fails catastrophically)
- expiry lives **inside** the ciphertext, so it is covered by the auth tag and
  cannot be extended by whoever holds the blob — there is a test that flips bytes
  across the body and requires every one to be rejected
- one primitive for confidentiality and integrity, so there is no
  encrypt-then-MAC ordering to get wrong

## Key separation

Derived from `INSFORGE_CLIENT_SECRET` with a **different label** from the client
id key, so the two have no computable relationship. That is what lets a public
signed value and a secret-bearing encrypted one live under one environment
variable — still no new config. A test asserts the keys differ and that each
rejects the other's output.

## Replay, stated rather than discovered

A sealed state is replayable inside its ten minutes, where a deleted row was
single-use. The bound is the platform: a replay re-presents an authorization code
the platform has already consumed, and it refuses. So a replay reaches our method
and then fails at the exchange — **it cannot mint a second session.** "Signed" and
"single-use" are easy to conflate, so this is written next to the code.

## Size — the #82 trap one hop further along

The sealed state *contains* the client id, and then travels in the **platform's**
authorize URL, so a large registration inflates a request line we do not own.
Measured on a running server rather than estimated:

```
realistic registration   client_id  171   platform authorize URL  1037
largest #82 accepts      client_id 3980   platform authorize URL  6494   (limit 8192)
```

`MAX_CLIENT_ID_LENGTH` bounds this hop too, with room. That is luck rather than
design, and it is noted in the code: anyone raising that constant has to
re-measure *this* number, not only the one it was chosen for.

## Verified with no Redis process anywhere

```
POST /oauth/register    201   client_id 171 chars
GET  /oauth/authorize   302   -> the platform          (was 500)
```

171 tests.

## What this does NOT do

Sign-in still will not **complete**. Two Redis uses remain on the login path —
`/oauth/callback` (`setex` the token) and `/oauth/select-project` (`get` + `del`)
— and both are needed before Quinn's end-to-end pass can run. This removes the
one that stopped a login from starting; it does not make one finish, and I would
rather say so than let a green `/oauth/authorize` imply otherwise.

---
## Summary by cubic
Move authorization state from a Redis row to a sealed AES-256-GCM value carried in the OAuth `state` parameter. This removes Redis from `/oauth/authorize`, keeps the PKCE verifier secret, and lets logins start reliably.

- **New Features**
  - Seal and verify auth state with AES-256-GCM: fresh nonce per seal, expiry inside the ciphertext, and tamper rejection.
  - Key separation: derive the auth-state key from `INSFORGE_CLIENT_SECRET` with a different label than the client-id signing key.
  - Replay is safe: a replay within 10 minutes cannot mint a second session because the platform refuses reused codes.
  - Size guard: measured authorize URL sizes; `MAX_CLIENT_ID_LENGTH` still bounds this hop, and must be re-measured if raised.

- **Bug Fixes**
  - `/oauth/authorize` works without Redis (302 to the platform instead of 500).
  - `getAuthorizationState` opens sealed values and logs rejections; returns null for invalid or expired state.

<sup>Written for commit b71caf97c768f11c03560b19aa0ebb2902981b09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/87?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/auth-state.test.ts
+++ b/src/http/auth-state.test.ts
@@ -1,0 +1,135 @@
+import { describe, it, expect } from 'vitest';
+import { createHmac } from 'crypto';
+import { sealAuthState, openAuthState, InvalidAuthStateError, AUTH_STATE_TTL_SECONDS } from './auth-state.js';
+import { deriveAuthStateKey, deriveClientIdSigningKey } from './config.js';
+
+const KEY = deriveAuthStateKey('a-platform-client-secret');
+const OTHER = deriveAuthStateKey('a-different-client-secret');
+
+const STATE = {
+  clientId: 'mcp_abc.def',
+  redirectUri: 'http://127.0.0.1:8765/callback',
+  scope: 'mcp:read mcp:write',
+  state: 'client-csrf-value',
+  codeChallenge: 'a'.repeat(43),
+  codeChallengeMethod: 'S256',
+  insforgeCodeVerifier: 'v'.repeat(43),
+  createdAt: 1785480000000,
+};
+
+describe('authorization state round-trips', () => {
+  it('returns exactly what was sealed', () => {
+    expect(openAuthState(sealAuthState(STATE, KEY), KEY)).toEqual(STATE);
+  });
+
+  it('produces a different blob every time for the same input', () => {
+    // A fresh nonce per seal. A repeated nonce under one key is how GCM fails
+    // catastrophically, so this is the property, not a nicety.
+    const a = sealAuthState(STATE, KEY);
+    const b = sealAuthState(STATE, KEY);
+    expect(a).not.toBe(b);
+    expect(openAuthState(a, KEY)).toEqual(openAuthState(b, KEY));
+  });
+});
+
+describe('the PKCE verifier is not readable from the sealed value', () => {
+  // This is the whole reason auth state is encrypted rather than signed, like
+  // client ids are. The state parameter travels through the platform and sits
+  // in the browser's address bar; a signed-but-readable envelope would publish
+  // the verifier to exactly the party PKCE defends against.
+  it('does not contain the verifier in any obvious encoding', () => {
+    const sealed = sealAuthState(STATE, KEY);
+    const body = sealed.slice('v1.'.length);
+    const decoded = Buffer.from(body, 'base64url').toString('binary');
+
+    for (const haystack of [sealed, decoded]) {
+      expect(haystack).not.toContain(STATE.insforgeCodeVerifier);
+      expect(haystack).not.toContain('insforgeCodeVerifier');
+    }
+    // ...nor the client's own redirect, which would make the blob a target list.
+    expect(sealed).not.toContain('127.0.0.1');
+  });
+});
+
+describe('a sealed state cannot be forged or altered', () => {
+  it('refuses another key', () => {
+    expect(() => openAuthState(sealAuthState(STATE, KEY), OTHER)).toThrow(InvalidAuthStateError);
+  });
+
+  it('refuses a key derived with the other label from the same secret', () => {
+    // Same environment variable, different purpose. If these collided, a public
+    // client id and a secret-bearing auth state would share a key.
+    const secret = 'one-secret-two-purposes';
+    const clientIdKey = Buffer.from(deriveClientIdSigningKey(secret), 'base64url');
+    expect(clientIdKey.length).toBe(32);
+    expect(clientIdKey.equals(deriveAuthStateKey(secret))).toBe(false);
+    expect(() => openAuthState(sealAuthState(STATE, deriveAuthStateKey(secret)), clientIdKey)).toThrow(
+      InvalidAuthStateError
+    );
+  });
+
+  it('refuses a single flipped byte anywhere in the blob', () => {
+    const sealed = sealAuthState(STATE, KEY);
+    const raw = Buffer.from(sealed.slice(3), 'base64url');
+    for (const i of [0, 6, 12, 20, raw.length - 1]) {
+      const tampered = Buffer.from(raw);
+      tampered[i] ^= 0x01;
+      expect(() => openAuthState(`v1.${tampered.toString('base64url')}`, KEY)).toThrow(
+        InvalidAuthStateError
+      );
+    }
+  });
+
+  it('refuses values that are not ours at all', () => {
+    for (const bad of ['', 'v1.', 'v1.notbase64!!', 'v2.abc', 'abc', 'v1.' + 'A'.repeat(20)]) {
+      expect(() => openAuthState(bad, KEY), bad).toThrow(InvalidAuthStateError);
+    }
+  });
+
+  it('refuses non-strings', () => {
+    for (const bad of [undefined, null, 42, {}]) {
+      expect(() => openAuthState(bad as unknown as string, KEY)).toThrow(InvalidAuthStateError);
+    }
+  });
+});
+
+describe('expiry travels inside the seal', () => {
+  const T0 = 1_700_000_000_000;
+
+  it('accepts up to the TTL and refuses after it', () => {
+    const sealed = sealAuthState(STATE, KEY, T0);
+    expect(openAuthState(sealed, KEY, T0)).toEqual(STATE);
+    expect(openAuthState(sealed, KEY, T0 + AUTH_STATE_TTL_SECONDS * 1000 - 1)).toEqual(STATE);
+    expect(() => openAuthState(sealed, KEY, T0 + AUTH_STATE_TTL_SECONDS * 1000)).toThrow(
+      InvalidAuthStateError
+    );
+  });
+
+  it('cannot have its expiry extended by whoever holds the blob', () => {
+    // exp is inside the ciphertext, so it is covered by the auth tag. If it sat
+    // beside the blob it could simply be rewritten.
+    const sealed = sealAuthState(STATE, KEY, T0);
+    const raw = Buffer.from(sealed.slice(3), 'base64url');
+    // Flip bytes across the ciphertext body; every attempt must be rejected
+    // rather than yielding a state with a later expiry.
+    for (let i = 28; i < Math.min(raw.length, 60); i += 7) {
+      const tampered = Buffer.from(raw);
+      tampered[i] ^= 0xff;
+      expect(() => openAuthState(`v1.${tampered.toString('base64url')}`, KEY, T0)).toThrow(
+        InvalidAuthStateError
+      );
+    }
+  });
+});
+
+describe(`a wrong key is our fault, not the caller's`, () => {
+  it('throws a plain Error, never InvalidAuthStateError', () => {
+    // Distinct on purpose: a misconfigured server must not report itself to a
+    // client as "your state is invalid", which is the mislabelling that sends
+    // someone to re-add a working install.
+    for (const bad of [Buffer.alloc(16), Buffer.alloc(0), 'not a buffer' as unknown as Buffer]) {
+      expect(() => sealAuthState(STATE, bad)).toThrow(Error);
+      expect(() => sealAuthState(STATE, bad)).not.toThrow(InvalidAuthStateError);
+    }
+  });
+});
--- a/src/http/auth-state.ts
+++ b/src/http/auth-state.ts
@@ -1,0 +1,137 @@
+import { createCipheriv, createDecipheriv, randomBytes } from 'crypto';
+
+/**
+ * Authorization state as a sealed value rather than a Redis row.
+ *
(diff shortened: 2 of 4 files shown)


jwfing (commented):
## Review: Authorization state as a sealed value

**Summary:** A well-constructed change that replaces the `/oauth/authorize` Redis write with a self-contained AES-256-GCM sealed `state`, correctly reasoned and thoroughly tested; no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the surrounding OAuth code, and the sibling `client-id.ts` (#82) pattern the PR builds on.

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality / robustness — the size bound is incomplete for the client-supplied `state`.** `src/http/auth-state.ts:22-40` and the seal at `src/http/oauth-manager.ts:129-139` argue that `MAX_CLIENT_ID_LENGTH` bounds the platform authorize-URL hop. That holds for `client_id`, and `redirect_uri` is bounded indirectly because `isRegisteredRedirectUri` (`src/http/server.ts:379`) rejects anything not in the registration (itself length-capped). But the client's own `state` CSRF value (`src/http/server.ts:306,397`) is passed straight into the seal and now rides in the **platform's** authorize URL — where in the pre-PR design only the random 43-char stateId did. It is not length-checked anywhere, so a client sending a multi-KB `state` can push the platform authorize URL past the 8192 request-line limit and break its own login. Blast radius is limited to that one client (self-inflicted), so this is non-blocking, but either cap the client `state` length at `/authorize` or soften the in-code claim that `MAX_CLIENT_ID_LENGTH` alone bounds this hop.

- **Software engineering — no test for the `oauth-manager` wrapper.** `auth-state.ts` itself is exercised exhaustively, but the new `getAuthorizationState` contract (`src/http/oauth-manager.ts:147-161`) — `InvalidAuthStateError` → `return null`, any *other* error re-thrown — isn't directly covered. A small unit test would lock in the "expired/forged is null, unexpected errors propagate" branch so a future refactor can't silently swallow a real fault.

#### Information

- **Dead payload in the auth-code record.** `createAuthorizationCode` still stores the full sealed blob under `stateId` (`src/http/oauth-manager.ts:215`), but `exchangeCode` never reads it (`:255`). Harmless, but now that the sealed value is ~1 KB+ rather than a 43-char id it is non-trivial dead weight in every auth-code row — safe to drop.

- **Replay-within-TTL is a genuine semantic change, and the PR states it honestly.** A deleted row was single-use; a sealed state is replayable for its 10 minutes (`src/http/oauth-manager.ts:222-231`). The reasoning that the platform's single-use authorization code is the real bound is sound — a replay reaches `createAuthorizationCode` but dies at the platform exchange. Flagging only so a future reader doesn't mistake "sealed" for "single-use."

- **Unbounded input reaches the cipher.** `openAuthState` (`src/http/auth-state.ts:76`) decrypts an arbitrary-length attacker-supplied `state` on `/oauth/callback` with no size cap before `createDecipheriv`. Negligible in practice (GCM is cheap), noted for completeness.

### Dimension notes
- **Security:** Strong. AES-256-GCM gives confidentiality+integrity from one primitive (no encrypt-then-MAC ordering to get wrong); a fresh 96-bit nonce per seal; `exp` sealed *inside* the ciphertext so it's tag-covered and cannot be extended; keys derived from `INSFORGE_CLIENT_SECRET` under a distinct label (`mcp-auth-state-v1`) so the public client-id key and the secret-bearing state key have no computable relationship. No secrets leak to logs — the rejection log messages are static strings and the verifier only ever exists inside the ciphertext (which is exactly why encrypting rather than signing is the right call here). The `InvalidAuthStateError`-vs-plain-`Error` split (caller-fault vs server-misconfig) is a nice touch and is tested.
- **Performance:** Net positive — removes a Redis round-trip from the authorize hot path and a `del` from the code path; no N+1 or new blocking I/O.
- **Test coverage:** Excellent for `auth-state.ts` — round-trip, non-determinism (fresh nonce), verifier-not-readable, single-flipped-byte rejection across the blob, key separation both directions, expiry boundary + non-extendability, malformed/non-string inputs, and wrong-key-is-server-fault. The removed constants (`AUTH_STATE_PREFIX`/`AUTH_STATE_TTL`) have no remaining references, so the build stays green.

### Verdict
**approved** (informational — zero Critical findings; posted as a COMMENT, human approval via the approve flow is separate). The two Suggestions are worth a look but none block merge.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review — Authorization state as a sealed value

**Summary:** Moves OAuth authorization state off Redis into an AES-256-GCM sealed value carried in the `state` parameter; the crypto is sound, the tests are thorough, and the core claim (no Redis on `/oauth/authorize`) holds — no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/` directory. Assessed against the PR description, the sibling `client-id.ts` module it deliberately mirrors, and the OAuth flow in `server.ts`.

I verified the PR's central claims directly:
- **Redis is off the authorize path.** `createAuthorizationState` (`src/http/oauth-manager.ts:124-141`) no longer calls `getRedisClient()`/`setex`; the only work is a `sealAuthState(...)`. `readClientId` upstream is pure crypto. Confirmed.
- **Crypto is correct.** AES-256-GCM with a fresh 12-byte nonce per seal, 16-byte tag, `nonce‖tag‖body` layout parsed back correctly (`auth-state.ts:73-115`), `exp` sealed *inside* the ciphertext so it's tag-covered, key length asserted `=== 32`, and two independent keys derived from one secret via distinct HMAC labels (`config.ts:129,166`). No nonce reuse, no encrypt-then-MAC ordering to get wrong.
- **Replay is bounded — and more tightly than the PR states.** Beyond the platform refusing a reused code, the select-project token in Redis is `del`'d after use (`server.ts:614`), so a replayed `state` inside its 10 minutes can't even reach a second code mint on this server.

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality — the size guard bounds `client_id`, but not everything that rides in the sealed blob.** `src/http/oauth-manager.ts:129-139` seals the whole `AuthorizationState`, which includes the client-supplied `redirect_uri` (≤2048), `scope`, and the client's own `state` — and `state`/`scope` are taken straight from the query with **no length bound** (`server.ts:306,322-323,397`). That sealed blob then becomes the platform authorize URL's `state` (`server.ts:393-407`). The PR's measurement ("largest #82 accepts → platform authorize URL 6494 < 8192") holds `client_id` at the limit but treats the other fields as small; a client that also sends a large-but-legal `state` or `scope` can push the platform authorize request line past 8192 and get a 431 / failed login. This is a genuine robustness regression versus the old fixed-size random `stateId`, which was constant regardless of payload. Low blast radius (a client only harms its own login), so non-blocking — but the code comment claiming `MAX_CLIENT_ID_LENGTH` "bounds this hop too" is incomplete and should either bound `state`/`scope` too or be reworded.

- **Software engineering — no test at the `OAuthManager` seam.** `auth-state.ts` itself is excellently covered, but the new `getAuthorizationState` try/catch that converts `InvalidAuthStateError` → `null` (and rethrows anything else) at `oauth-manager.ts:147-161`, plus the `createAuthorizationState` → `getAuthorizationState` round-trip through `authStateKey()`, have no direct test (there is no `oauth-manager.test.ts`). A one-test regression guard on "invalid sealed state ⇒ null, unexpected error ⇒ throws" would lock in the oracle-avoidance behavior the comment carefully describes.

#### Information

- **Performance — Redis key is now the full sealed blob.** `mcp:oauth:token:${state}` (`server.ts:520-526,555,595,614`) keys Redis by the ~1KB (realistic) sealed value instead of the former 43-char id. Functionally fine and URL-safe (base64url + `.`), just noting the larger keys/URL now flow through four Redis ops and the select-project redirect.
- **Software engineering — decrypted payload isn't shape-revalidated.** Unlike `readClientId`, which re-checks structure after signature verification, `openAuthState` returns `parsed.v` as `T` with only an `exp` check (`auth-state.ts:105-115`). Acceptable here because the value is authenticated (only this server could produce it); noted only for parity with the sibling module.

#### Security
No security regressions. Encryption keeps `insforgeCodeVerifier` out of the authorize URL, address bar, and platform logs — a net improvement over a signed-only envelope; the "verifier not readable" test enforces it. No new secrets logged (`oauth-manager.ts:156` logs only `error.message`), no auth checks weakened, no new dependencies (Node `crypto` builtins only).

#### Performance
Sealing is a single in-process AES-GCM op replacing a Redis round-trip — a latency win on the authorize path. See the Information note on key size for the one minor cost.

### Verdict
**approved** (informational — human approval via the separate approve flow). Zero Critical findings; two Suggestions and two Information notes above are non-blocking. Nicely reasoned change with genuinely strong crypto tests.

tonychang04:
Non-blocking, but the size argument in `auth-state.ts:31-41` is load-bearing for this PR and I re-measured both halves of it. Two corrections, one in your favour.

**1. The 8192 budget is not the limit of the hop it protects.** Probed the actual platform authorize endpoint today (`GET https://api.insforge.dev/api/oauth/v1/authorize`, junk `client_id`, padded `state`):

```
request-line ~ 8,079   400   (reaches the app)
request-line ~16,079   400   (reaches the app)
request-line ~16,279   431
request-line ~16,479   414
request-line ~30,079   414
```

So the binding constraint on this hop is Node's 16,384 `max-http-header-size`, not the 8k nginx buffer the comment assumes — nothing in front of `api.insforge.dev` cuts at 8k. The PR has ~2x the headroom it claims.

**2. `MAX_CLIENT_ID_LENGTH` does not bound this hop.** The reported `client_id 3980 -> 6494` is measured with a *short* chosen `redirect_uri`. The `redirect_uri` is sealed too, and it can legally be 2048 chars (`MAX_URI_LENGTH`) while the `client_id` stays under 4096. Running the real `mintClientId` + `sealAuthState` from this branch:

```
registration                                 client_id  sealed  platform request-line
cursor:// x1, no client state                    186     738      1,018   ok
max client_id, short chosen redirect_uri       4,066   5,853      6,133   ok   <- the PR's case
max client_id + 2048-char chosen redirect_uri  4,035   8,515      8,795   over the PR's 8192
realistic client_id + 6KB scope                  118   8,555      8,835   over the PR's 8192
```

Row 3 mints fine and breaks the PR's own stated budget with no oversized client input at all. Under the *measured* 16,384 it still works, which is why this isn't blocking — but "MAX_CLIENT_ID_LENGTH turns out to bound this hop too, with room" is true of the case that was measured, not of the worst case our own bounds permit.

And rows 3-4 are the ceiling-free ones: `scope` and the client's `state` (`server.ts:306,322-323`) have no length bound anywhere, so 12KB of either clears 16,384 too. Unbounded input, unbounded URL — the same shape #82 landed `MAX_CLIENT_ID_LENGTH` to close.

**Suggested fix, same one-liner #82 already uses:** bound the finished sealed string inside `sealAuthState`/`createAuthorizationState` rather than adding two more per-field bounds. One check, in the units that fail (bytes on a request line), on the thing that actually travels — it cannot drift out of sync with the fields the way `MAX_CLIENT_ID_LENGTH` just did. A 400 at `/oauth/authorize` naming the oversized parameter beats a 431 from someone else's server mid-redirect.

Measurement script and raw probes available if useful. — Max

PR #88, opened 2026-07-31
The platform token rides in the state, not a row keyed by it

tonychang04:
Third of the five Redis uses, and the one that made `/oauth/callback` and
`/oauth/select-project` need Redis at all:

```
callback        SETEX mcp:oauth:token:<state>   the platform access token
select-project  GET it twice (page render, then POST), then DEL
```

All three are keyed by the state the token already travels with. So fold it in:
`attachPlatformToken` re-seals the existing authorization state with the token
attached, and the callback redirects with that as the new `state_id`.

**The `state_id` changes at the callback, deliberately.** The sealed value *is*
the record, so a record with one more field is a different string — there is
nothing to update in place, which is the point.

## Why this is only safe because the envelope is encrypted

A platform access token is a bearer credential. A signed-but-readable blob would
put it in a URL, the address bar and the platform's logs. That is the property
`auth-state.ts` was built for, and it is the only reason the token can live here
rather than in a row.

Expiry restarts at the callback, matching what it replaces: the Redis token row
carried its own fresh 10-minute TTL from the moment it was written, and the
person still has a project to choose.

**`server.ts` now has no Redis call anywhere in the OAuth path** — the only
`getRedisClient()` left in that file is the boot ping.

## Measured with no Redis process anywhere

```
select-project, state WITHOUT a token   400  "Session expired..."
select-project, state WITH a token      401  from api.insforge.dev
                                             (fake token, so the platform
                                             refuses — but it GOT there)
select-project, garbage state           400
Redis errors in the server log          0

sealed state, no token            337 chars
sealed state, ~900-char token    1570 chars   (fits a URL with room)
```

The 401 is the result that matters: it means every state check passed and the
handler reached the platform, which is the whole path minus a real credential.

## Not a complete login yet

`createAuthorizationCode` still writes `mcp:auth:code:` and the token binding
writes `mcp:auth:binding:`, so the code exchange at `/oauth/token` is next.
**Two of five left.**

171 tests.

---
## Summary by cubic
Moves the platform access token from Redis into the encrypted OAuth state, removing Redis from `/oauth/callback` and `/oauth/select-project`. The callback reseals the state with the token and redirects with a new `state_id`.

- **Refactors**
  - Add `platformAccessToken?` to the authorization state and `attachPlatformToken` to reseal with the token.
  - Callback stores the token by resealing and redirects with the new, encoded `state_id`.
  - Project selection reads the token from state; removes Redis GET/DEL; consistent “Session expired” handling.
  - Expiry resets at callback to preserve the 10‑minute TTL behavior.
  - OAuth path in `server.ts` no longer calls Redis (only the boot ping remains).

<sup>Written for commit 632c8526c2118bcf69326c98ece0460eddaad174. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/88?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/oauth-manager.ts
+++ b/src/http/oauth-manager.ts
@@ -57,6 +57,18 @@ interface AuthorizationState {
   // Our PKCE verifier for calling Insforge OAuth
   insforgeCodeVerifier: string;
 
+  /**
+   * The platform access token, present only AFTER the callback has exchanged
+   * the code for it.
+   *
+   * It used to live in its own Redis row (mcp:oauth:token:<state>) read twice
+   * by the project-selection page. Folding it into the state it was already
+   * keyed by removes the row and the second lookup — and it is safe here for
+   * exactly one reason: this envelope is encrypted, not signed. A bearer token
+   * in a signed-but-readable blob would be a bearer token in a URL.
+   */
+  platformAccessToken?: string;
+
   createdAt: number;
 }
 
@@ -103,6 +115,19 @@ function generateCode(): string {
  * OAuthManager handles the OAuth authorization flow and token-to-project binding
  */
 export class OAuthManager {
+  /**
+   * Re-seal an existing state with the platform token attached.
+   *
+   * Returns a NEW state_id: the sealed value IS the record, so adding a field
+   * necessarily produces a different string. The expiry restarts here, which
+   * matches what it replaces — the Redis token row carried its own fresh
+   * 10-minute TTL from the moment the callback wrote it, and the person still
+   * has a project to choose.
+   */
+  attachPlatformToken(authState: AuthorizationState, platformAccessToken: string): string {
+    return sealAuthState({ ...authState, platformAccessToken }, authStateKey());
+  }
+
   /**
    * Create a new authorization state (step 1 of OAuth flow)
    * Returns a state ID and the PKCE code challenge for Insforge OAuth
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -516,14 +516,14 @@ app.get(OAUTH_ENDPOINTS.callback, async (req: Request, res: Response) => {
 
     console.log('[OAuth] Token received, redirecting to project selection...');
 
-    const redis = getRedisClient();
-    await redis.setex(
-      `mcp:oauth:token:${state}`,
-      10 * 60,
-      tokens.access_token
-    );
+    // The token rides inside the state instead of a row keyed by it. The
+    // state_id therefore CHANGES here — it is the record, so a record with one
+    // more field is a different string.
+    const stateWithToken = oauthManager.attachPlatformToken(authState, tokens.access_token);
 
-    res.redirect(`${SERVER_CONFIG.publicUrl}${OAUTH_ENDPOINTS.selectProject}?state_id=${state}`);
+    res.redirect(
+      `${SERVER_CONFIG.publicUrl}${OAUTH_ENDPOINTS.selectProject}?state_id=${encodeURIComponent(stateWithToken)}`
+    );
   } catch (error) {
     console.error('OAuth callback error:', error);
     getAnalyticsService().trackOAuthFailure({
@@ -550,16 +550,16 @@ app.get(OAUTH_ENDPOINTS.selectProject, async (req: Request, res: Response) => {
 
   try {
     const oauthManager = getOAuthManager();
-    const redis = getRedisClient();
-
-    const token = await redis.get(`mcp:oauth:token:${state_id}`);
-    if (!token) {
-      return res.status(400).send('Session expired. Please start the authorization process again.');
-    }
 
     const authState = await oauthManager.getAuthorizationState(state_id as string);
     if (!authState) {
-      return res.status(400).send('Invalid or expired state');
+      return res.status(400).send('Session expired. Please start the authorization process again.');
+    }
+    const token = authState.platformAccessToken;
+    if (!token) {
+      // A state that never went through the callback. Same message: from the
+      // person's side these are the same event — a sign-in that did not finish.
+      return res.status(400).send('Session expired. Please start the authorization process again.');
     }
 
     const projectGroups = await oauthManager.getAvailableProjects(token);
@@ -590,20 +590,15 @@ app.post(OAUTH_ENDPOINTS.selectProject, async (req: Request, res: Response) => {
 
   try {
     const oauthManager = getOAuthManager();
-    const redis = getRedisClient();
-
-    const token = await redis.get(`mcp:oauth:token:${state_id}`);
-    if (!token) {
-      return res.status(400).send('Session expired. Please start the authorization process again.');
-    }
 
     const authState = await oauthManager.getAuthorizationState(state_id as string);
-    if (!authState) {
+    if (!authState?.platformAccessToken) {
       return res.status(400).json({
         error: 'invalid_request',
         error_description: 'Invalid or expired state',
       });
     }
+    const token = authState.platformAccessToken;
 
     const code = await oauthManager.createAuthorizationCode(
       state_id as string,
@@ -611,7 +606,8 @@ app.post(OAUTH_ENDPOINTS.selectProject, async (req: Request, res: Response) => {
       project_id as string
     );
 
-    await redis.del(`mcp:oauth:token:${state_id}`);
+    // Nothing to delete: the token was never stored. It stops being usable
+    // when the sealed state it rides in expires.
 
     const redirectUrl = new URL(authState.redirectUri);
     redirectUrl.searchParams.set('code', code);


jwfing (commented):
## Review — #88 "The platform token rides in the state, not a row keyed by it"

**Summary:** Solid, well-reasoned refactor that folds the `mcp:oauth:token:<state>` Redis row into the encrypted `AuthorizationState` envelope, correctly removing all Redis from the `/oauth/callback` → `/oauth/select-project` path; I found no blocking issues, only a few non-blocking notes.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — no matching spec/plan found, so I assessed against the PR description, the surrounding OAuth code, and the design rationale documented in `src/http/auth-state.ts`. Verified the "expiry restarts at the callback" and "state_id changes deliberately" claims against the code, and confirmed the AES-256-GCM sealing makes the token-in-URL move safe.

---

### Critical
(none)

---

### Suggestion

**Security — loss of the single-use property (`src/http/oauth-manager.ts:192-257`, `src/http/server.ts:581-607`).** The old `mcp:oauth:token:<state>` row was `DEL`'d on the select-project POST, so the state was single-use for project selection. The sealed token-state cannot be invalidated, so within its 10-minute TTL the same `state_id` can drive `POST /oauth/select-project` repeatedly — each call mints a fresh token binding + a fresh authorization code. Blast radius is genuinely low: the envelope is encrypted so a leaked `state_id` does **not** expose the token, minted codes are delivered only to the client-fixed `redirectUri`, and (when the client used PKCE) the exchange still needs the `code_verifier`. The author's replay note at `oauth-manager.ts:247-255` is correct for the `/oauth/token` exchange path, but it reasons about re-presenting *the same* code; it doesn't cover that the select-project POST now mints *new* codes on replay. Worth extending that comment — or, if cheap, binding a one-time nonce into the sealed state and rejecting reuse.

**Software engineering — no test for the new behavior (`src/http/oauth-manager.ts:127-129`).** `attachPlatformToken` and the "read the token from `authState.platformAccessToken` instead of Redis" branch in both select-project handlers land with no unit coverage. There's no existing `oauth-manager`/server-endpoint test, so this refactor of a previously-untested path isn't a regression — but this is a security-sensitive credential path, the repo follows TDD, and `src/http/auth-state.test.ts` already shows the seal primitive is trivially unit-testable. A round-trip test (`attachPlatformToken` → `openAuthState` → assert `platformAccessToken` present and `exp` refreshed) would lock in the two claims the PR body rests on.

---

### Information

**Functionality — dead field now larger (`src/http/oauth-manager.ts:240`).** `createAuthorizationCode` still stores `stateId` inside the `mcp:auth:code:` record, but `exchangeCode` (`oauth-manager.ts:265-311`) never reads it back. Pre-existing, but folding the token in grew that stored value from ~337 to ~1570 chars — dead data written to Redis for every code. Consider dropping `stateId` from the code record.

**Functionality — `createdAt` vs `exp` drift (`src/http/oauth-manager.ts:72`, `src/http/auth-state.ts:65`).** The re-sealed state keeps the original authorize-time `createdAt` while the envelope `exp` resets to now+10m. Harmless today (nothing reads `createdAt` for expiry), but the two timestamps are now inconsistent — flagging in case future logic ever trusts `createdAt`.

**Performance — size budget for the new hop (`src/http/server.ts:525`, `src/http/auth-state.ts:30-41`).** The select-project redirect URL grows to ~1570 chars (measured in the PR body), comfortably under Node's 16KB header limit and the ~8KB request-line budget the codebase tracks. But the platform-access-token length is an unstated assumption, and the careful size analysis in `auth-state.ts` covers only the `client_id` → platform-authorize-URL hop, not this new token-bearing select-project hop. A one-line note there would help whoever next bumps a length constant.

**Security — token-in-URL is sound (positive).** Putting the bearer token in the redirect URL is safe precisely because the envelope is AES-256-GCM (`auth-state.ts`): the select-project page's third-party subresources (Google Fonts, the `insforge.dev` logo) receive at most opaque ciphertext, and modern browsers' default `strict-origin-when-cross-origin` policy strips the query from cross-origin `Referer` anyway. `encodeURIComponent` on the redirect is a harmless no-op for base64url but good defensive hygiene. The hidden `state_id` form field is `escapeHtml`'d. No token leak.

---

### Verdict

**approved** (informational — posted as a comment; a human still gives the explicit GitHub approval). Zero Critical findings. The refactor is correct, the encryption-based safety argument holds, and the flow (callback re-seal → escaped hidden field → POST) round-trips properly. The Suggestions above (replay note + a cheap round-trip test) are worth addressing but non-blocking.

jwfing (commented):
## Review — #88 "The platform token rides in the state, not a row keyed by it"

**Summary:** A clean, well-reasoned refactor that folds the platform access token into the encrypted auth-state envelope and removes all three Redis calls (`SETEX` / `GET`×2 / `DEL`) from `/oauth/callback` and `/oauth/select-project`; the crypto is sound and the code is correct, but the change quietly drops the single-use property that the deleted `redis.del` provided.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, so there is **no matching spec/plan** — assessed against the PR description and the surrounding OAuth code (`src/http/auth-state.ts`, `src/http/oauth-manager.ts`, `src/http/server.ts`). This is the third in a documented series ("Two of five left") of Redis-removal refactors, consistent with #82 (stateless client ids) and #87 (stateless auth state).

---

### Critical
(none)

### Suggestion

**Security — single-use enforcement is lost; the sealed state is replayable for the full 10 minutes** — `src/http/oauth-manager.ts:247-256`, `src/http/server.ts:603-610`

The old design deleted the token row (`redis.del(mcp:oauth:token:<state>)`) at the end of the `select-project` POST, making the whole selection step effectively single-use: a second POST with the same `state_id` failed with "Session expired". After this change the token lives in the sealed state for its full TTL and nothing consumes it, so `POST /oauth/select-project` can be replayed as many times as desired within ~10 minutes.

The in-code justification (`oauth-manager.ts:249-255`) only defends against replay of the *authorization code* at `/oauth/token` (protected by `GETDEL`). It does not cover replay of the *select-project POST itself*: each replay re-enters `createAuthorizationCode`, mints a **fresh** code, and the handler responds with a 302 whose `Location` carries that code (`server.ts:612-619`). An actor who obtains the sealed `state_id` (it now travels in the `select-project` URL — browser history, `Referer`, access/proxy logs — and as a hidden form field in the rendered page) can therefore mint fresh codes for the window. PKCE closes the exploit *when a client used it*, but `createAuthorizationState` treats `codeChallenge` as optional (`oauth-manager.ts:145`, `152-156`), so a non-PKCE client leaves the code directly exchangeable. Please either (a) re-establish single-use (e.g. a tiny `state`-hash consumption marker), or (b) make PKCE mandatory for this flow, or (c) explicitly document the accepted risk. If PKCE is not enforced for all MCP clients, treat this as blocking.

**Software engineering — no test exercises the new branches** — `src/http/oauth-manager.ts:127-129`

`attachPlatformToken` and the two new "no token in state → 400" branches (`server.ts:558-563`, `594-600`) have no direct coverage. `attachPlatformToken` is a thin delegate to the well-tested `sealAuthState`, and the `server.ts` handlers were already untested (no `server.test.ts`/`oauth-manager.test.ts`), so this isn't a regression — but a small round-trip test asserting `platformAccessToken` survives seal/open and that a token-less state is rejected would lock in the contract this PR depends on. `src/http/auth-state.test.ts`'s `STATE` fixture (line 9-18) does not include the new field.

### Information

- **Effective auth-state lifetime is extended.** Re-sealing at the callback with a fresh `Date.now()` (`auth-state.ts:60-66`) restarts the 10-minute TTL. The comment (`oauth-manager.ts:121-125`) correctly notes this matches the old token-row TTL, but the *authorization state* previously expired 10 min after `/authorize`; now the callback grants it another 10 min, so the end-to-end window can approach ~2× the original. Benign, but worth noting since the same sealed blob is what the replay note above is about.
- **State size in the server's own redirect URL.** The token now inflates the `select-project` URL to ~1570 chars for a ~900-char token (per the PR body). Fits comfortably, but `openAuthState` has no length guard, and the 4096-char bound discussed in `auth-state.ts:30-41` was measured for the *platform* authorize URL, not this hop — a future token-size change should be re-measured against this URL too.
- **Performance (positive, no findings).** Removes one `SETEX` and two `GET`s plus a `DEL` from the hot OAuth path; no new queries, loops, or blocking work introduced.
- `encodeURIComponent` on the redirect (`server.ts:525`) is correct and slightly hardens the pre-existing un-encoded `state` redirect. No dangling `mcp:oauth:token` references remain and the `getRedisClient` import is still needed for the boot ping (`server.ts:1192`).

---

### Verdict: **approved** (informational — human approval via the approve flow)

No Critical findings. The refactor is correct and the crypto envelope (AES-256-GCM, per-seal nonce, exp inside the tag) properly keeps the bearer token confidential. The one thing I'd want addressed before merge is the lost single-use guarantee — please confirm PKCE is mandatory for these clients, or restore single-use / document the accepted replay window.

jwfing (approved):
LGTM - approved.

PR #89, opened 2026-07-31
The state parameter is a handle again; the record rides in a cookie

tonychang04:
Clears the blocker: the platform stores `state` in a **255-character column**, #87 put a 681-character sealed envelope in it, and a login could not start. Quinn found it by following the redirect I stopped short of.

Shrinking does not reach 255 — ~410 is the floor for a state that can still complete a callback. And a cookie alone is not sufficient either; it moves the bound to ~4096 rather than removing one. **Both halves, so every component is bounded rather than hoped about:**

```
drop clientId from the seal   nothing read it after authorize, and it is the
                              largest field (up to 4096 on its own)
bound the client's state      512, rejected at authorize with invalid_request
handle in `state`             32 hex chars
record in a cookie            HttpOnly, Path=/oauth, Max-Age=600
```

**The handle is not decoration.** It is in the `state` parameter *and* inside the sealed record, and the callback requires them to match. Without that binding the cookie alone would authorise any callback carrying one — exactly the CSRF `state` exists to prevent.

**`SameSite=Lax` is load-bearing, not a weaker choice.** The callback is a top-level GET navigation from the platform's origin: Lax sends cookies on exactly that, Strict withholds them. A Strict cookie would be invisible at the one moment it is needed. There is a test asserting Lax so a later "hardening" fails instead of breaking sign-in. `secure` follows `publicUrl` for the same class of reason.

## Measured against a running server

```
state parameter          32 chars   (was 681; the column is 255)
Set-Cookie total        610 bytes   (cap ~4096)

callback, cookie + matching handle   reaches the PLATFORM token exchange
callback, cookie + wrong handle      400, and the mismatch is logged
callback, no cookie                  400
authorize, 513-char client state     400, naming the 512 limit
```

The first line is the one that matters: with a matching cookie the callback got all the way to `token_exchange_failed: invalid_client` — the platform rejecting a fake code, which means every state check passed.

## What still needs doing

**This has not been driven against the real platform.** The remaining verification is to deploy and follow the redirect on the slug — the exact step whose absence hid this bug. `mcp:auth:code:` and `mcp:auth:binding:` also remain, so a login still cannot *complete*.

183 tests.

---
## Summary by cubic
Fixes OAuth sign-in by making `state` a short handle and moving the sealed record to a cookie. Fits the platform’s 255-char `state` column and preserves CSRF protection by binding the handle to the cookie; now enforced by API and tests.

- **Bug Fixes**
  - Enforce client `state` length ≤ 512 at `/oauth/authorize`; longer values return `invalid_request`.
  - Set HttpOnly cookie `mcp_oauth_state` with SameSite=Lax, Path=/oauth, Max-Age=600; `secure` follows `publicUrl`.
  - Require cookie + matching `state` handle at callback and select-project; mismatch or missing cookie returns 400.
  - Clear `mcp_oauth_state` after successful code handoff to reduce replay risk.

- **Refactors**
  - Add `newStateHandle()` (32 hex chars) and carry `{ handle, sealedState }` through the flow.
  - Make `OAuthManager.getAuthorizationState(sealed, handle)` require the handle and verify it matches.
  - Update routes: authorize sets the cookie and sends the handle; callback/select-project read the cookie and pass the handle.
  - Drop `clientId` from the sealed state to keep the cookie under 4 KB.
  - Add tests for CSRF binding, handle size/unpredictability, cookie attributes/parsing, and client `state` bounds.

<sup>Written for commit d487b72736d40127deca8711536c379c640dc119. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/89?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added improved OAuth authorization-state protection using secure, short-lived cookies and matching request handles.
  * Added validation for OAuth state values, including length limits and malformed cookie handling.
  * Authorization and callback flows now require a matching state cookie and handle.
  * Authorization cookies are cleared after the flow completes.

* **Tests**
  * Added comprehensive coverage for cookie security, parsing, uniqueness, validation, state binding, and invalid input handling.

### DIFF
--- a/src/http/auth-state-cookie.test.ts
+++ b/src/http/auth-state-cookie.test.ts
@@ -1,0 +1,94 @@
+import { describe, it, expect } from 'vitest';
+import {
+  AUTH_STATE_COOKIE,
+  cookieAttributes,
+  readCookie,
+  newStateHandle,
+  isAcceptableClientState,
+  MAX_CLIENT_STATE_LENGTH,
+} from './auth-state-cookie.js';
+
+describe('the handle fits the column that broke us', () => {
+  it('is far inside 255 characters', () => {
+    // The whole reason this file exists: the platform stores `state` in a
+    // 255-char column, and the sealed record could not be shrunk below ~410.
+    for (let i = 0; i < 50; i++) {
+      const handle = newStateHandle();
+      expect(handle.length).toBe(32);
+      expect(handle.length).toBeLessThan(255);
+      expect(handle).toMatch(/^[0-9a-f]{32}$/);
+    }
+  });
+
+  it('is unpredictable', () => {
+    const seen = new Set(Array.from({ length: 200 }, () => newStateHandle()));
+    expect(seen.size).toBe(200);
+  });
+});
+
+describe('cookie attributes', () => {
+  it('is SameSite=Lax, because Strict would withhold it on the callback', () => {
+    // The callback is a top-level GET navigation from the platform's origin.
+    // Lax sends cookies on exactly that; Strict does not — a Strict cookie
+    // would be invisible at the one moment it is needed and every sign-in
+    // would fail. This assertion exists to stop a later "hardening".
+    expect(cookieAttributes('https://mcp.insforge.dev').sameSite).toBe('lax');
+  });
+
+  it('is HttpOnly and scoped to the OAuth routes', () => {
+    const attrs = cookieAttributes('https://mcp.insforge.dev');
+    expect(attrs.httpOnly).toBe(true);
+    expect(attrs.path).toBe('/oauth');
+  });
+
+  it('follows the public URL for Secure rather than hardcoding it', () => {
+    // A Secure cookie is silently dropped on plaintext http, and local runs are
+    // http://127.0.0.1 — hardcoding true makes every local sign-in fail in a
+    // way that looks like a code bug.
+    expect(cookieAttributes('https://mcp.insforge.dev').secure).toBe(true);
+    expect(cookieAttributes('http://127.0.0.1:3000').secure).toBe(false);
+  });
+});
+
+describe('readCookie', () => {
+  it('finds the value among others', () => {
+    expect(readCookie(`a=1; ${AUTH_STATE_COOKIE}=v1.abc; b=2`, AUTH_STATE_COOKIE)).toBe('v1.abc');
+  });
+
+  it('percent-decodes, because that is how it was written', () => {
+    expect(readCookie(`${AUTH_STATE_COOKIE}=v1.a%2Bb%3D`, AUTH_STATE_COOKIE)).toBe('v1.a+b=');
+  });
+
+  it('does not match a name that merely contains ours', () => {
+    expect(readCookie(`x_${AUTH_STATE_COOKIE}=nope`, AUTH_STATE_COOKIE)).toBeUndefined();
+    expect(readCookie(`${AUTH_STATE_COOKIE}_x=nope`, AUTH_STATE_COOKIE)).toBeUndefined();
+  });
+
+  it('treats a malformed value as absent rather than throwing mid-sign-in', () => {
+    expect(readCookie(`${AUTH_STATE_COOKIE}=%E0%A4%A`, AUTH_STATE_COOKIE)).toBeUndefined();
+  });
+
+  it('handles no header, an empty header and a valueless entry', () => {
+    expect(readCookie(undefined, AUTH_STATE_COOKIE)).toBeUndefined();
+    expect(readCookie('', AUTH_STATE_COOKIE)).toBeUndefined();
+    expect(readCookie('justaname', AUTH_STATE_COOKIE)).toBeUndefined();
+  });
+});
+
+describe('the client state is bounded before we seal it', () => {
+  it('accepts what real clients send, and absence', () => {
+    expect(isAcceptableClientState(undefined)).toBe(true);
+    expect(isAcceptableClientState('')).toBe(true);
+    expect(isAcceptableClientState('s'.repeat(43))).toBe(true);
+    expect(isAcceptableClientState('s'.repeat(MAX_CLIENT_STATE_LENGTH))).toBe(true);
+  });
+
+  it('rejects one character over, and non-strings', () => {
+    // Unbounded client state means an unbounded cookie, which is the same
+    // whole-payload mistake one layer down.
+    expect(isAcceptableClientState('s'.repeat(MAX_CLIENT_STATE_LENGTH + 1))).toBe(false);
+    for (const bad of [42, {}, ['a'], null]) {
+      expect(isAcceptableClientState(bad)).toBe(false);
+    }
+  });
+});
--- a/src/http/auth-state-cookie.ts
+++ b/src/http/auth-state-cookie.ts
@@ -1,0 +1,123 @@
+import { randomBytes } from 'crypto';
+
+/**
+ * Where the sealed authorization state actually travels.
+ *
+ * #87 put the whole sealed envelope in the OAuth `state` parameter. Every byte
+ * budget it was measured against — Node's max-http-header-size, nginx's request
+ * line — was satisfied, and the platform still returned 500: it stores `state`
+ * in a **255-character column**. A limit nobody greps for, enforced by a
+ * different system, one hop past where I stopped measuring.
+ *
+ * Shrinking does not reach 255. The floor for a state that can still complete a
+ * callback is ~410 characters, measured across every variant the flow permits.
+ * So the envelope has to travel somewhere else, and `state` goes back to being
+ * a short handle.
+ *
+ * A cookie on our own origin is that somewhere, and it is not sufficient on its
+ * own either — it moves the bound to roughly 4096 bytes rather than removing
+ * one. Both halves are needed: drop the client id from the seal (nothing reads
+ * it after authorize) and bound the client's own `state`, so every component is
+ * bounded rather than hoped about.
+ *
+ *   client_id at its 4096 cap, client state 512   6639   OVER a 4096 cookie
+ *   no client_id, client state bounded to 512     1159   fits
+ *   realistic (loopback, 43-char state)            544   ample
+ *
+ * The handle is not decoration. It is in the `state` parameter AND inside the
+ * sealed payload, and the callback requires them to match — a double-submit
+ * binding. Without it the cookie alone would authorise any callback that
+ * arrived with a cookie, which is the CSRF the `state` parameter exists to
+ * prevent.
+ */
+
+/** The cookie carrying the sealed state. Scoped to the OAuth routes only. */
+export const AUTH_STATE_COOKIE = 'mcp_oauth_state';
+
+/**
+ * How long the cookie lives. Matches the seal's own expiry — the seal is
+ * authoritative, since a cookie's Max-Age is a request from us that the browser
+ * is free to ignore.
+ */
+export const AUTH_STATE_COOKIE_MAX_AGE_SECONDS = 10 * 60;
+
+/**
+ * A handle for one authorization, short enough for a 255-char column with room
+ * for anything else the platform decides to put beside it.
+ */
+export function newStateHandle(): string {
+  return randomBytes(16).toString('hex'); // 32 chars
+}
+
+export interface CookieAttributes {
+  httpOnly: true;
+  secure: boolean;
+  sameSite: 'lax';
+  path: string;
+  maxAge: number;
+}
+
+/**
+ * Attributes for the state cookie.
+ *
+ * `sameSite: 'lax'` is load-bearing and NOT a weaker choice made for
+ * convenience. The callback arrives as a top-level GET navigation from the
+ * platform's origin. Lax sends cookies on exactly that; **Strict withholds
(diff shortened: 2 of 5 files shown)


jwfing (commented):
## Review — #89 "The state parameter is a handle again; the record rides in a cookie"

**Summary:** Sound, well-reasoned fix that moves the sealed OAuth record out of the length-capped `state` parameter into an encrypted `Path=/oauth` cookie while keeping a 128-bit handle as the double-submit CSRF binding — I found no blocking issues, and a few things worth tightening before deploy.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory, so this was assessed against the PR description, the linked motivation (#87's 681‑char state vs. the platform's 255‑char column), and the surrounding code. Verified against the actual code paths in `src/http/server.ts`, `src/http/oauth-manager.ts`, and `src/http/auth-state.ts`.

### Findings

#### 🔴 Critical
(none)

#### 🟡 Suggestion

- **[software-engineering] The security-critical binding has no test.** The new `src/http/auth-state-cookie.test.ts` thoroughly covers the helper module (handle length/entropy, cookie attributes, `readCookie` edge cases, client-state bounds) — but the actual CSRF property this PR exists to enforce is untested. Nothing exercises `oauth-manager.ts:189-195` (handle mismatch → `null`) or the "both halves required" wiring in `server.ts` (a callback with a valid cookie but wrong/missing `state` must fail). `oauth-manager.ts` has no test file at all. Per the repo's TDD convention, a regression test asserting `getAuthorizationState(sealed, wrongHandle) === null` and that each endpoint returns 400 when the cookie is absent would lock in the behavior a future refactor could silently break.

- **[functionality] The post-token cookie size is unmeasured.** The 610-byte measurement in the description is the *authorize* cookie (pre-token). The second `res.cookie(...)` in the callback (`server.ts:552`) re-seals with `platformAccessToken` folded in (`oauth-manager.ts:139-141`). If the platform token is a large JWT, the sealed+base64url+percent-encoded value could approach the browser's ~4096-byte per-cookie limit; over it, the browser silently drops the cookie and `/oauth/select-project` fails with "Session expired." Dropping `clientId` removed the dominant contributor and the realistic estimate stays well under 4096, and the PR is honest that this hasn't been driven against the real platform — so non-blocking, but this is the specific value to measure during that verification step (a value the size of a real platform token, all the way through project selection).

#### 🔵 Information

- **[functionality] The handle check is silently optional.** `getAuthorizationState(sealed, expectedHandle?)` skips the binding when `expectedHandle === undefined` (`oauth-manager.ts:189`). All four current callers guard param presence first (`server.ts:459` `sealed && state`, `:476` `if (!code || !state)`, `:576` `if (!state_id)`, `:616` `if (!state_id || !project_id)`), so it's safe today. But a future caller that forgets the guard loses the CSRF binding with no compile error. Consider making the handle required, so the signature enforces the safe usage.

- **[security] Completion cookie is not cleared.** After POST `/oauth/select-project` redirects back to the client with the code, the encrypted token cookie lingers (HttpOnly, `Path=/oauth`, ~10 min). As the `createAuthorizationCode` comment notes (`oauth-manager.ts:265-274`), the sealed state is replayable within its window — bounded by the platform rejecting an already-consumed code and by `GETDEL` on the local auth code. A `res.clearCookie(AUTH_STATE_COOKIE, { path: '/oauth' })` on completion would be cheap defense-in-depth. Low risk given the value is encrypted, HttpOnly, and SameSite=Lax.

**Things done well (verified):** AES-256-GCM seal with `exp` inside the AEAD-authenticated payload (`auth-state.ts:60-73`) so expiry can't be extended; the `SameSite=Lax` reasoning is correct and load-bearing — it sends the cookie on the top-level GET callback navigation while withholding it on a cross-site POST to `/oauth/select-project`, which incidentally blocks CSRF on that endpoint; `secure` correctly follows `publicUrl`; the handle is a proper 128-bit random double-submit token bound inside the sealed record; PKCE S256-only and `redirect_uri` exact-match are preserved; `readCookie` is dependency-free, prefix-safe, and treats malformed percent-encoding as absent rather than throwing mid-sign-in.

- **[performance] No concerns — net improvement.** The change removes a Redis row (`mcp:oauth:token:<state>`) and its second lookup; `readCookie` is O(header length); no new blocking I/O, N+1, or per-request allocation of note.

### Verdict
**approved** (informational — no Critical findings; the two Suggestions and the size verification are worth addressing before the real-platform deploy the PR already scopes as remaining work). Posted as a comment; a human still gives the explicit GitHub approval.

jwfing (commented):
## Review — #89 "The state parameter is a handle again; the record rides in a cookie"

**Summary:** Sound, well-reasoned fix that puts a 32-char handle back in the platform `state` parameter and moves the sealed record into a `Path=/oauth` cookie bound by a double-submit handle check; the flow traces end-to-end, gates are green (tsc clean, 183/183 tests), and I found no blocking issues.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, so no matching spec/plan was found — assessed against the PR description, the referenced #87 regression (681-char sealed envelope vs. the platform's 255-char `state` column), and the surrounding OAuth code.

I verified the full flow in `src/http/server.ts`: authorize sets the sealed cookie + sends the handle → callback reads cookie + matches handle → re-seals with the platform token → select-project GET renders the handle into the hidden form field (`templates/project-selection.ts:32`) → POST re-reads the cookie + handle. `OAUTH_ENDPOINTS` are all under `/oauth/*` (`config.ts:282-288`), so `Path=/oauth` covers every route. `SameSite=Lax` is correct: the callback is a cross-site top-level GET (Lax sends it), and the select-project POST is same-site (SameSite doesn't gate it). `maxAge: …*1000` is right — Express treats `maxAge` as ms and emits `Max-Age=600`, matching the seal's own `exp`. Confirmed no remaining reads of the dropped `clientId` field anywhere in `oauth-manager.ts`/`server.ts`.

### Critical
(none)

### Suggestion

- **Software engineering / security — the CSRF handle-binding branch is untested (`src/http/oauth-manager.ts:189-195`).** The pure helpers in `auth-state-cookie.ts` have excellent coverage (12 tests: handle length/entropy, `SameSite=Lax`, `secure`-follows-URL, `readCookie` percent-decode/substring-name/malformed, client-state bounds), and `auth-state.ts` seal/open is separately tested. But the one line the whole security story rests on — `state.handle !== expectedHandle → return null` — has no direct test, and there's no automated exercise of the server-side cookie round-trip. A cheap unit test (seal a state with handle A, `getAuthorizationState(sealed, "B")` → `null`; matching handle → returns the state) would lock in the binding so a future refactor that drops the check fails loudly instead of silently reopening the CSRF the `state` parameter exists to prevent. The PR body already flags that this "has not been driven against the real platform," so this is the automated safety net for the part that can be tested without one.

### Information

- **Security — post-token cookie size isn't bounded/asserted (`src/http/server.ts:548-552`).** The file header reasons carefully about staying under ~4096: drop `clientId`, bound client `state` to 512. That reasoning covers the *pre-token* cookie; the callback then re-seals with `platformAccessToken` added, and that field isn't length-bounded. Measured at 610B in practice, so no action needed now — but a large platform JWT plus a 512-char client `state` is the one path that could approach the cap, and unlike the pre-token case there's no guard or test pinning it. Worth a one-line note or assert if platform tokens ever grow.
- **Functionality — single fixed cookie name (`AUTH_STATE_COOKIE = 'mcp_oauth_state'`).** Two concurrent OAuth flows in the same browser share one cookie; the second `authorize` overwrites the first, so the older callback then fails the handle match with a 400. This is a pre-existing single-cookie limitation and a rare case, noted only for awareness — not something this PR needs to solve.
- **Security (positive) — this reduces secret exposure.** The encrypted envelope (which carries the PKCE verifier) no longer travels in the platform authorize URL or the `[OAuth] Redirecting…` log line (`server.ts:426,430`) — only the opaque 32-char handle does. The platform token in the cookie is AES-256-GCM sealed, `HttpOnly`, `Secure` (in prod, correctly following `publicUrl`), `Lax`, and `Path`-scoped. No secrets or PII are newly logged; the new mismatch log line is content-free.

### Verdict
`approved` (informational — human approval is still a separate action). Zero Critical findings; the Suggestion and Information items are non-blocking. Net a security and performance improvement (removes a Redis token row and its second lookup) that cleanly clears the #87 blocker.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review — #89 `state` is a handle again; the record rides in a cookie

**Summary:** A correct and well-reasoned fix that moves the sealed OAuth record from the 255-char-bounded `state` parameter into an HttpOnly cookie on our own origin while keeping the CSRF guarantee via a handle bound into both halves — no Critical findings.

### Requirements context
No `/docs/superpowers/` (or any `docs/`) directory exists in this repo, so there is **no matching spec/plan** — assessed against the PR description, the referenced #87 regression, and the surrounding code (`auth-state.ts`, `oauth-manager.ts`, `server.ts`). The stated goal (fit the platform's 255-char `state` column without losing CSRF protection) is met.

I verified the full flow end-to-end in the workspace: authorize sets the cookie + sends the 32-char handle (`server.ts:420-424`); callback reads cookie + handle, re-seals with the platform token and re-sets the cookie preserving the handle via `attachPlatformToken` (`oauth-manager.ts:139-141`, `server.ts:548-554`); GET/POST select-project read the cookie and re-verify (`server.ts:583-586`, `:626-643`); POST clears it (`server.ts:649`). The double-submit binding holds at every hop.

---

### Findings

#### Critical
(none)

#### Suggestion

- **[functionality] Single fixed cookie name breaks concurrent sign-ins in one browser** — `server.ts:420` and `:552` both write the fixed cookie `mcp_oauth_state`. If a user starts two authorize flows in the same browser, the second overwrites the first's cookie; the first callback then presents a `state` handle that no longer matches the cookie's sealed handle and 400s (`oauth-manager.ts:192-198`). #87's URL-carried state was self-contained and per-flow, so this is a narrow behavioral regression. Blast radius is low (sign-ins are near-always sequential, and #87 couldn't start a login at all), so this is non-blocking — but worth either a line in the design notes or, if you want to support it, keying the cookie name by handle.

- **[software engineering / test coverage] No route-level test for the cookie wiring this PR introduces** — `auth-state-cookie.test.ts` and `oauth-state-binding.test.ts` cover the helpers and the manager-level binding thoroughly, and this is consistent with the repo's convention (no `server.test.ts` / supertest harness exists). But the actual glue in `server.ts` — cookie set on authorize, re-set-with-token on callback, the `sealed && state` guards on all three read sites, and `clearCookie` on completion — is exercised by nothing. Given this PR's own thesis ("the exact shape of gap this repo keeps finding" — a property tested not at all) and the author's note that it "has not been driven against the real platform," a request-level test asserting the `Set-Cookie` on `/oauth/authorize` and the missing-cookie/mismatched-handle 400s on `/oauth/callback` would close precisely the loop flagged as un-driven.

#### Information

- **[software engineering] `clearCookie` omits the set-time attributes** — `server.ts:649` clears with only `{ path: '/oauth' }`. Deletion matches on name+path so this works, but mirroring `secure`/`sameSite` from `cookieAttributes` avoids browser warnings and reads more symmetrically with the set call.

- **[performance] Redundant seal open on select-project POST** — `getAuthorizationState` opens the seal at `server.ts:628`, then `createAuthorizationCode` opens it a second time (`oauth-manager.ts:222`). Two AES-256-GCM decrypts per submit is negligible; noting only because the already-opened `authState` could be threaded through if this path ever gets hot.

- **[security] `getAuthorizationState` handle check is safe by construction** — the `state.handle !== expectedHandle` comparison (`oauth-manager.ts:192`) would technically pass if both sides were `undefined`, but this is unreachable: handles are always set at seal time, sealed cookies are unforgeable (AES-GCM), and every real handle is a 32-char hex string. No change needed — flagging so a future refactor doesn't accidentally make `handle` optional (the code comment already guards this intent).

---

### Dimension coverage

- **Software engineering:** Strong helper/binding tests, clear naming, hand-rolled `readCookie` avoids a dependency and correctly treats malformed values as absent. Only gap is route-level coverage (Suggestion above).
- **Functionality:** Flow is correct across all four endpoints; handle survives the re-seal; `secure` correctly follows `publicUrl` so local http still works. One concurrency edge case (Suggestion).
- **Security:** Sound. Double-submit CSRF binding (handle in both `state` and the sealed payload), HttpOnly + SameSite=Lax (correctly justified for the top-level callback navigation), PKCE verifier stays inside the encrypted seal, client `state` bounded to 512 at authorize (`server.ts:332-341`), `clientId` dropped from the payload. No secrets logged (`oauth-manager.ts:196` logs only a generic mismatch message); no token in any URL. No new dependencies.
- **Performance:** Net improvement — removes a Redis token row and its second lookup. No N+1, no blocking work added.

### Verdict
**approved** (informational — human approval still required via the approve flow). Zero Critical findings; the two Suggestions are non-blocking.

jwfing (approved):
LGTM - approved.

PR #90, opened 2026-07-31
The page promised a repair that does not repair

tonychang04:
Blair ran #84's printed commands as a stuck user instead of reasoning about them, and answered the question left open in that very file — *"whether removing the config entry also drops the client's saved OAuth registration is client-specific"*.

**For Claude Code it does not.** The saved authorisation lives in `mcpOAuth` under `name|sha256(type,url,headers)`, a store `add-mcp` never touches. Removing and re-adding the **same** url rebuilds the identical key and picks the dead registration straight back up.

Which makes the merged page wrong where it mattered most:

> …add it back with the command below, and **this will complete normally**.

On an unchanged hostname that is false for Claude Code, and it was aimed at exactly the population it would fail. A page that hands someone a confident non-fix is worse than one that says nothing, because they stop looking.

## What actually works

```
Claude Code   /mcp -> pick the server -> Clear authentication   (a UI action)
Codex         codex mcp logout insforge
```

The Claude Code step is not a command, so it belongs in the sentence rather than in a code block someone would try to paste. Codex's is real and joins the list **first**, because order is the instruction.

On a **new** hostname the remove/add pair does repair by itself — different url, different key, no saved state — so all three are printed and the wording no longer claims more than any of them can deliver.

## Rendered

```
Your client is holding a sign-in that no longer works

…it will keep reusing it until you clear it. In Claude Code: run /mcp, pick
this server, and choose Clear authentication. In Codex: run the first command
below. Then reconnect with the last one.

1. codex mcp logout insforge
2. npx add-mcp remove https://mcp.insforge.dev/mcp -y
3. npx add-mcp https://mcp.insforge.dev/mcp
```

Zero occurrences of the old promise. A test now **forbids** `complete normally` and **requires** the Clear authentication step, so the claim cannot quietly come back.

## Still unverified, stated rather than implied

Nobody has watched a real Claude Code reuse that record against a live server. Blair proved the record survives the commands; the reuse is inferred from how the key is derived. That remains Quinn's one-step editor test.

189 tests.

---
## Summary by cubic
Fixes the OAuth error page to stop promising a reconnect that doesn’t work and adds clear steps to reset saved auth, especially for Claude Code. Improves reliability by updating command order and adding tests to prevent regressions.

- **Bug Fixes**
  - Updated error copy to tell users to clear saved authentication (Claude Code: use /mcp → Clear authentication; Codex: `codex mcp logout insforge`).
  - Changed `reconnectCommand(mcpUrl)` to return three steps: `codex mcp logout insforge`, `npx add-mcp remove <url> -y`, then `npx add-mcp <url>`.
  - Removed the “complete normally” promise and clarified behavior on same vs new hostnames.
  - Strengthened tests to assert the new text, command order, and correct host usage.

<sup>Written for commit 2e45e7d31d64767b54592a91d1c09ac71796e65d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/90?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -65,6 +65,7 @@ describe('every browser-visible failure says what to do', () => {
     for (const error of ['invalid_client', 'invalid_request']) {
       const action = humanFormOf({ error }, MCP_URL).action as string[];
       expect(action).toEqual([
+        'codex mcp logout insforge',
         `npx add-mcp remove ${MCP_URL} -y`,
         'npx add-mcp https://mcp.insforge.dev/mcp',
       ]);
@@ -72,16 +73,32 @@ describe('every browser-visible failure says what to do', () => {
     }
   });
 
-  it('removes before adding, so it is not a no-op on an unchanged URL', () => {
+  it('clears the saved authorisation before reconnecting', () => {
     // add-mcp deep-merges under the server key and byte-preserves the rest, and
     // the key is inferred from the hostname. So on a same-hostname cutover a
     // bare `add-mcp <url>` rewrites an identical entry and changes nothing:
     // the stuck user runs it, fails identically, and opens a ticket. Removing
     // first is what makes the instruction do something — and it is correct on
     // a new hostname too, where it clears the stale entry anyway.
-    const [first, second] = reconnectCommand(MCP_URL);
-    expect(first).toContain('remove');
-    expect(second).toContain(MCP_URL);
+    // Blair ran these as a stuck user: on an unchanged hostname remove+add
+    // rebuilds the SAME mcpOAuth key and picks the dead registration back up,
+    // so the commands alone repair nothing. Clearing the client's saved
+    // authorisation is the step that works, and it comes first.
+    const [logout, remove, add] = reconnectCommand(MCP_URL);
+    expect(logout).toContain('logout');
+    expect(remove).toContain('remove');
+    expect(add).toContain(MCP_URL);
+  });
+
+  it('does not promise the reconnect alone will work', () => {
+    // The page used to say "add it back ... and this will complete normally".
+    // On an unchanged hostname that promise is false for Claude Code, and it
+    // was aimed at exactly the people it would fail.
+    for (const error of ['invalid_client', 'invalid_request']) {
+      const { message } = humanFormOf({ error }, MCP_URL);
+      expect(message).not.toContain('complete normally');
+      expect(message).toMatch(/Clear authentication/);
+    }
   });
 
   it('names the host the person is actually talking to, in BOTH commands', () => {
@@ -95,12 +112,15 @@ describe('every browser-visible failure says what to do', () => {
     // so passing the URL is exact on every host.
     const slug = 'https://keen-pulse-fsjr9.run.mcp-use.com/mcp';
     expect(reconnectCommand(slug)).toEqual([
+      'codex mcp logout insforge',
       `npx add-mcp remove ${slug} -y`,
       `npx add-mcp ${slug}`,
     ]);
-    for (const command of reconnectCommand(slug)) {
+    // The two add-mcp commands are host-derived; the codex one names the
+    // server as that client stored it, which is our product name on any host.
+    for (const command of reconnectCommand(slug).slice(1)) {
       expect(command).toContain('keen-pulse-fsjr9');
-      expect(command).not.toContain('insforge');
+      expect(command).not.toContain('insforge.dev');
     }
   });
 
--- a/src/http/oauth-error-response.ts
+++ b/src/http/oauth-error-response.ts
@@ -76,11 +76,30 @@ export interface HumanForm {
  * name, and that the package stores no tokens of its own (its single "oauth"
  * reference is an oauthScopes CONFIG field).
  *
- * What this still cannot promise: whether removing the config entry also drops
- * the client's saved OAuth registration is client-specific — add-mcp deletes
- * its config key and nothing more. If some client keys its registration by
- * server name rather than URL, even remove+add leaves it stuck. That is a real
- * test on a real editor, not something to settle in a comment.
+ * AND THE ANSWER, which is no. Blair ran these commands as a stuck user
+ * instead of reasoning about them: for Claude Code the saved authorisation
+ * lives in `mcpOAuth` under `name|sha256(type,url,headers)`, a store add-mcp
+ * never touches. Removing and re-adding the SAME url rebuilds the identical key
+ * and picks the dead registration straight back up. So on an unchanged
+ * hostname these commands do not repair anything on their own, and the page
+ * used to promise "this will complete normally" — a promise that would have
+ * been broken for exactly the population it was written for.
+ *
+ * The step that actually works is clearing the client's saved authorisation,
+ * and it is client-specific:
+ *
+ *   Claude Code   /mcp -> pick the server -> Clear authentication   (a UI action)
+ *   Codex         codex mcp logout insforge
+ *
+ * The Claude Code one is not a command, so it belongs in the sentence rather
+ * than in a code block a person would try to paste. On a NEW hostname the
+ * remove/add pair does repair by itself — different url, different key, no
+ * saved state — so both are printed and the wording no longer claims more than
+ * either can deliver.
+ *
+ * Still unverified, and stated rather than implied: nobody has watched a real
+ * Claude Code reuse that record against a live server. Blair proved the record
+ * survives the commands; the reuse is inferred from the key derivation.
  *
  * Built from publicUrl, which is the CANONICAL host this deployment is
  * configured to be — deliberately not the host the request arrived on. The Host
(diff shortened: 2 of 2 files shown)


jwfing (commented):
## Review — #90 "The page promised a repair that does not repair"

**Summary:** Rewrites the browser-visible OAuth error page (`invalid_client` / `invalid_request`) to stop promising that a bare remove+add "will complete normally" and instead tell users to clear the client's saved authorisation first, backed by tests that forbid the old promise — a correct and well-tested content fix, with a couple of consistency gaps worth a look before merge.

**Requirements context:** This repo has no `/docs/superpowers/` (nor `docs/specs/`, nor any `docs/`) — no matching spec/plan found, so I assessed against the PR description, the retained design commentary in `oauth-error-response.ts`, and the referenced prior PRs (#74, #84). Intent is clear from those.

---

### Critical
(none)

### Suggestion

**1. Functionality — the `npx add-mcp remove …` command is rendered but the prose never tells anyone to run it.** `src/http/oauth-error-response.ts:161-166` (and `:177-181`) say *"In Codex: run the first command below … Then reconnect with the last one."* That positionally references `action[0]` (`codex mcp logout insforge`) and `action[2]` (`npx add-mcp <url>`), leaving the middle `npx add-mcp remove <url> -y` (`reconnectCommand`, `:144`) orphaned — shown, but with no described purpose in either the Claude Code or Codex path. Compounding this, the template renders each command as an *unnumbered* `<code>` block (`src/http/templates/oauth-error.ts:100-102`), so "the first"/"the last" require the reader to count. The PR body's "Rendered" mock shows a numbered 1/2/3 list, but the shipped page is not numbered. Recommend either numbering the list or sequencing the steps explicitly, and clarifying whether `remove` is still needed in the same-host case (the PR's own thesis is that remove+add is a no-op there).

**2. Functionality — `codex mcp logout insforge` hardcodes the server name.** `src/http/oauth-error-response.ts:143`. The rest of `reconnectCommand` deliberately derives everything from the URL specifically to stay correct on non-canonical hosts — see the comment at `:126-141` noting add-mcp's key for `keen-pulse-fsjr9.run.mcp-use.com` is `keen-pulse-fsjr9`, not `insforge`. If Codex likewise stores the server under a host-derived name, `logout insforge` matches nothing on a cutover host — the exact hardcoded-name bug the URL approach was built to avoid. It is correct on the canonical production host (the page is built from `publicUrl` → `mcp.insforge.dev` → `insforge`, and that is the only host serving this page in prod), so it is **not broken today**; but the test at `oauth-error-response.test.ts:119-124` codifies the unverified assumption ("our product name on any host"). Recommend confirming the Codex-stored name (or deriving it) before relying on it for a Manufact-style migration host. Note: I verified against the Codex CLI docs that `codex mcp logout <name>` is a real subcommand ("remove OAuth credentials for MCP server") — the command itself is not hallucinated; only the hardcoded `insforge` argument is the concern.

### Information

- **SE — a test guard was quietly relaxed.** `oauth-error-response.test.ts:121-123` now does `.slice(1)` and relaxes `.not.toContain('insforge')` → `.not.toContain('insforge.dev')`. After the slice, the remaining two commands (the add-mcp pair on the slug URL) contain no `insforge` at all, so the original stricter assertion would still pass — the relaxation is unnecessary and slightly weakens the intent of guarding the canonical host name out of slug commands.
- **UX — all three commands render for every viewer**, including `codex mcp logout insforge` shown to Claude Code users who don't run Codex. This is an acknowledged single-page-for-multiple-clients tradeoff; noting for awareness.

### Dimension coverage

- **Software engineering:** Good TDD. The new tests forbid `complete normally` and require `Clear authentication` across both branches (`:93-102`), and the array-destructure test was updated to `[logout, remove, add]` (`:87`). Coverage is reasonable; the Claude Code UI ("Clear authentication") path is inherently not unit-testable and is honestly flagged as unverified in both the code and PR body.
- **Functionality:** Core fix is correct — the false "complete normally" promise is gone and the clear-auth step is conveyed. Gaps are the orphaned `remove` command and the hardcoded codex name above (both non-blocking).
- **Security:** No security-relevant changes. `mcpUrl` derives from `publicUrl` (config, deliberately *not* the attacker-controlled `Host` header — see `:104-116`), and all output is HTML-escaped (`templates/oauth-error.ts:100-102, 108-115`). No user input reaches the page unescaped.
- **Performance:** No performance-relevant changes — pure string construction on a cold error path.

---

Zero Critical findings. The content fix is correct, in-scope, and well-tested. The two Suggestions (orphaned `remove` command / brittle positional wording, and the hardcoded `codex … insforge` name) are non-blocking consistency improvements. Posting as a COMMENT — explicit GitHub approval remains a separate human action.

jwfing (commented):
## Review — #90 "The page promised a repair that does not repair"

**Summary:** Correctly replaces the false "this will complete normally" promise with client-specific "clear the saved authorisation" guidance and adds a `codex mcp logout insforge` step, with good regression tests; no blocking issues, but the new Codex step reintroduces the hardcoded-`insforge` host-portability problem the rest of this function was written to avoid.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessing against the PR description, linked #74/#84 context, and the surrounding code alone. No spec/plan was invented.

Verified against upstream docs: `codex mcp logout <name>` is a real subcommand (`codex-rs/cli/src/mcp_cmd.rs` — "logout — remove OAuth credentials for MCP server"), so the command itself is not hallucinated. It takes the server **name** as configured in `~/.codex/config.toml`.

---

### Critical
(none)

---

### Suggestions

**1. Functionality — `codex mcp logout insforge` hardcodes the server name, contradicting this file's own host-derivation invariant** — `src/http/oauth-error-response.ts:142-146`, `src/http/oauth-error-response.test.ts:119-124`

The `remove`/`add` steps are deliberately URL-derived precisely because "the name is derived from the hostname, so a hardcoded `insforge` matches nothing on any host that infers a different key — including the Manufact slug we spent a day testing on" (lines 126-141; the same lesson is restated at 65-70). The new `codex mcp logout insforge` reintroduces exactly that hardcoded `insforge`. On the canonical `mcp.insforge.dev` deployment it's correct, but on a white-label/Manufact host (`keen-pulse-fsjr9…`) Codex would hold the server under a host-derived name and `logout insforge` targets nothing. The test at line 119-124 codifies the assumption ("our product name on any host") by slicing the codex command out of the host-portability check — but that assumption is the very thing lines 65-70 argue against for `add-mcp`. Non-blocking because the worst case is a harmless "no such server" error and the Claude Code prose guidance still stands; but on non-canonical hosts a Codex user gets a dead command. Consider deriving the logout name from the host the same way the other two steps are, or explicitly scoping this page to canonical-host deployments.

**2. Functionality — the middle command is printed but no prose ever tells anyone to run it** — `src/http/oauth-error-response.ts:144` rendered via `162-166` & `177-181`

Both branch messages route the user to "the first command" (codex logout) and "reconnect with the last one" (the bare `add`), skipping `npx add-mcp remove <url> -y` (index 1) entirely. So all three commands render as pasteable code blocks, yet neither the Claude Code nor the Codex instruction references the middle one. Given the PR's own thesis (the dead state lives in the client's OAuth store, not in add-mcp's config), the `remove` step is now largely vestigial for the same-host case — but leaving it displayed-but-uninstructed is confusing. Either drop it from the same-host messaging or have the prose account for it.

**3. Software engineering — the `default` fallback branch wasn't updated to match the new messaging** — `src/http/oauth-error-response.ts:193-201`

The default case still attaches `action: reconnectCommand(mcpUrl)` (now including `codex mcp logout insforge`) while its prose only says "Removing the InsForge MCP server … and adding it back is the usual remedy" — no mention of Clear authentication or of what the codex line does. It also pairs an arbitrary `error_description` (e.g. the test's "the disk melted") with three repair commands. Pre-existing shape, but the PR's new codex line now surfaces unexplained here; worth a consistency pass.

---

### Information

- **Test assertion loosened** — `src/http/oauth-error-response.test.ts:121-124`: the host-portability loop now runs over `.slice(1)` and checks `not.toContain('insforge.dev')` instead of the previous `not.toContain('insforge')` over all commands. Reasonable given the codex line legitimately contains `insforge`, but the guard on the `add-mcp` commands is now slightly weaker.
- **Redundant-ish assertion** — `src/http/oauth-error-response.test.ts:76-91`: the `contains('logout'/'remove'/URL)` checks are fully subsumed by the exact-equality test at lines 60-74; harmless, just overlapping coverage.
- The PR body is admirably honest that the Claude Code reuse is inferred from key derivation, not observed ("Quinn's one-step editor test"). Recommend that manual verification land before relying on this page for the white-label case, since Suggestion 1 hinges on the same key-derivation reasoning.

---

### Security
No security-relevant changes. `mcpUrl` is built from `SERVER_CONFIG.publicUrl` (canonical host, deliberately **not** the attacker-controlled `Host` header — lines 104-116), all rendered command/message lines pass through `escapeHtml`, and no secrets, tokens, or new dependencies are introduced.

### Performance
No performance-relevant changes. The touched functions are pure string builders with no I/O, loops, or allocations of concern; no consumer outside this file depends on the array length (`reconnectCommand` widening from 2→3 elements breaks nothing).

---

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings: the fix is correct on the canonical deployment, well tested, and the regression guards (`not "complete normally"`, requires `Clear authentication`) are exactly right. The Suggestions above — chiefly the hardcoded `insforge` in the Codex step on non-canonical hosts — are worth addressing but non-blocking.

jwfing (approved):
LGTM - approved.

PR #91, opened 2026-07-31
Authorization codes carry their own record, and PKCE becomes required

tonychang04:
Fourth of five. `mcp:auth:code:` is gone; `oauth-manager`'s only remaining Redis use is the token binding.

## The one place this rewrite tightens behaviour

`GETDEL` made a stored code **single-use**: redeemed once, gone. A sealed code cannot be single-use — there is nothing to delete — so it is **replayable for its lifetime**.

What makes that acceptable is PKCE: a replayed code without the verifier is useless, and the verifier never leaves the client. **Without PKCE a replay is a full second session**, so the honest choice is to refuse at issue time rather than hand out a code we cannot protect.

Nothing real is lost by requiring it. The MCP spec requires PKCE of public clients, the SDK always sends it, and our AS metadata already advertises S256 as the *only* supported method — so this refuses a combination we never told anyone we would accept.

## Measured, including the part that is a genuine loosening

```
correct verifier             ok     {"tokenHash":"th"}
REPLAY, correct verifier     ok     <- replayable, as documented
no verifier                  THROW  Code verifier required
wrong verifier               THROW  Code verifier mismatch
wrong redirect_uri           THROW  Redirect URI mismatch
garbage code                 THROW  Invalid or expired authorization code
code with no challenge       THROW  Authorization code is missing its challenge

sealed code length           275 chars
```

The replay line is in the table on purpose. It is the cost of the change and it should be visible in review, not buried in a comment.

## Smaller things

- **Five minutes, not the state's ten.** RFC 6749 §4.1.2 wants a code short-lived, and one that cannot be single-use wants it more. `sealAuthState` takes a TTL per value now instead of assuming one.
- **A third derived key, third label.** Three purposes with no computable relationship between their keys: a client id is public and signed, an auth state is secret and encrypted, an auth code is a bearer credential in its own right. A test asserts each key rejects the others' output.

## Remaining

`mcp:auth:binding:` — the last one, and the one that needs the token to become self-describing rather than a Redis key. Until it lands a login still cannot complete.

210 tests.

---
## Summary by cubic
Authorization codes are now stateless, sealed records with a 5-minute TTL, and PKCE is required to redeem them. Redis `mcp:auth:code:` storage is removed; only `mcp:auth:binding:` remains.

- **Refactors**
  - Replace stored codes with sealed payloads; TTL is per value (5 minutes for codes).
  - Enforce PKCE at issue and redeem; codes without a challenge are rejected.
  - Add a dedicated auth code key derived from the client secret; keys for client ID, state, and code are isolated.
  - Redeem path opens the sealed code, validates redirect URI, and verifies the code verifier; replay limited by PKCE and expiry.

- **Migration**
  - No client changes needed; SDKs already send PKCE and the AS advertises `S256` only.

<sup>Written for commit c044d3c81d29d4e00ab67c958528b16193d0a88b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/91?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/auth-state.test.ts
+++ b/src/http/auth-state.test.ts
@@ -133,3 +133,42 @@ describe(`a wrong key is our fault, not the caller's`, () => {
     }
   });
 });
+
+describe('the TTL is per-value, because a code is not a state', () => {
+  const KEY2 = deriveAuthStateKey('another-secret');
+  const T0 = 1_700_000_000_000;
+
+  it('honours a shorter TTL than the default', () => {
+    // RFC 6749 §4.1.2 wants an authorization code short-lived, and a code that
+    // cannot be single-use (nothing to delete) wants it more. Five minutes for
+    // codes, ten for state.
+    const sealed = sealAuthState({ code: true }, KEY2, T0, 5 * 60);
+    expect(openAuthState(sealed, KEY2, T0 + 5 * 60 * 1000 - 1)).toEqual({ code: true });
+    expect(() => openAuthState(sealed, KEY2, T0 + 5 * 60 * 1000)).toThrow(InvalidAuthStateError);
+  });
+
+  it('still defaults to the state TTL when none is given', () => {
+    const sealed = sealAuthState({ state: true }, KEY2, T0);
+    expect(openAuthState(sealed, KEY2, T0 + AUTH_STATE_TTL_SECONDS * 1000 - 1)).toEqual({ state: true });
+    expect(() => openAuthState(sealed, KEY2, T0 + AUTH_STATE_TTL_SECONDS * 1000)).toThrow();
+  });
+});
+
+describe('three purposes, three keys', () => {
+  it(`gives each label a key that rejects the other's output`, async () => {
+    const { deriveAuthCodeKey } = await import('./config.js');
+    const secret = 'one-secret-three-purposes';
+    const stateKey = deriveAuthStateKey(secret);
+    const codeKey = deriveAuthCodeKey(secret);
+
+    expect(stateKey.equals(codeKey)).toBe(false);
+    // An auth code must not be openable as an auth state, or a weakness in the
+    // most exposed value reaches the other.
+    expect(() => openAuthState(sealAuthState({ x: 1 }, codeKey), stateKey)).toThrow(
+      InvalidAuthStateError
+    );
+    expect(() => openAuthState(sealAuthState({ x: 1 }, stateKey), codeKey)).toThrow(
+      InvalidAuthStateError
+    );
+  });
+});
--- a/src/http/auth-state.ts
+++ b/src/http/auth-state.ts
@@ -57,12 +57,17 @@ const TAG_BYTES = 16;
  * covered by the authentication tag and cannot be extended by an attacker who
  * holds the blob.
  */
-export function sealAuthState<T>(value: T, key: Buffer, nowMs: number = Date.now()): string {
+export function sealAuthState<T>(
+  value: T,
+  key: Buffer,
+  nowMs: number = Date.now(),
+  ttlSeconds: number = AUTH_STATE_TTL_SECONDS
+): string {
   assertKey(key);
 
   const payload = JSON.stringify({
     v: value,
-    exp: nowMs + AUTH_STATE_TTL_SECONDS * 1000,
+    exp: nowMs + ttlSeconds * 1000,
   });
 
   const nonce = randomBytes(NONCE_BYTES);
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -179,6 +179,28 @@ export function authStateKey(): Buffer {
   return deriveAuthStateKey(INSFORGE_CONFIG.clientSecret);
 }
 
+/**
+ * And a third key, for authorization codes. Same secret, third label.
+ *
+ * Three purposes, three keys, no relationship between them: a client id is
+ * public and signed, an auth state is secret and encrypted, an auth code is a
+ * bearer credential in its own right. Sharing one key across all three would
+ * mean a weakness in the most exposed use reaches the other two.
+ */
+const AUTH_CODE_KEY_LABEL = 'mcp-auth-code-v1';
+
+export function deriveAuthCodeKey(clientSecret: string): Buffer {
+  if (!clientSecret) {
+    throw new Error('INSFORGE_CLIENT_SECRET is required: the authorization code key is derived from it');
+  }
+  return createHmac('sha256', clientSecret).update(AUTH_CODE_KEY_LABEL).digest();
+}
+
+/** Lazy for the same reason as the others. */
+export function authCodeKey(): Buffer {
+  return deriveAuthCodeKey(INSFORGE_CONFIG.clientSecret);
+}
+
 // ============================================================================
 // Redis Configuration
 // ============================================================================
--- a/src/http/oauth-manager.ts
+++ b/src/http/oauth-manager.ts
@@ -2,7 +2,7 @@ import { createHash, randomBytes } from 'crypto';
 import { getRedisClient } from './redis.js';
 import { sealAuthState, openAuthState, InvalidAuthStateError } from './auth-state.js';
 import { newStateHandle } from './auth-state-cookie.js';
-import { authStateKey } from './config.js';
+import { authStateKey, authCodeKey } from './config.js';
 import {
   validateToken,
   getProjectAccess,
@@ -103,7 +103,6 @@ interface TokenBinding {
 
 // Redis key prefixes
 const TOKEN_BINDING_PREFIX = 'mcp:auth:binding:';
-const AUTH_CODE_PREFIX = 'mcp:auth:code:';
 
 // TTLs
 const AUTH_CODE_TTL = 5 * 60; // 5 minutes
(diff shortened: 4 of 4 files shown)


jwfing (commented):
## Review: Authorization codes carry their own record, and PKCE becomes required

**Summary:** Converts OAuth authorization codes from Redis-stored single-use handles into stateless, self-contained AES-256-GCM sealed tokens (5-min TTL, third derived key) and makes PKCE mandatory at issue and exchange — a clean, well-reasoned refactor with no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory (confirmed), so this is assessed against the PR title/body and the surrounding code (`auth-state.ts`, `config.ts`, `oauth-manager.ts`, `server.ts`). The change is a coherent continuation of the prior stateless-`auth-state` work (`client-id.ts` → sealed `AuthorizationState`), applying the same pattern to the authorization code.

---

### Critical
(none)

The core security properties hold up. Three-key separation via distinct HMAC labels is real and tested (`config.ts:190-202`, `auth-state.test.ts:157-174`); the expiry lives inside the AEAD-authenticated payload so it can't be extended (`auth-state.ts:60-79`); only S256 is accepted with no `plain` downgrade path (`oauth-manager.ts:345-347`); and a code without the PKCE verifier is unusable. No secrets are newly logged or returned. I could not construct a forge/replay path that mints a session.

### Suggestion

**Functionality — mandatory PKCE is enforced late and ungracefully** (`oauth-manager.ts:268-273`, `server.ts:301-409`, `server.ts:698-703`)
`/oauth/authorize` still accepts a request with no `code_challenge` (`server.ts:408` passes it through as optional). The new requirement only fires much later, inside `createAuthorizationCode`, *after* the user has authenticated on the platform and selected a project — and the resulting thrown `Error` surfaces at the project-selection endpoint as a generic `server_error` (`server.ts:700-703`), not a spec-compliant `invalid_request`. A non-PKCE client therefore burns the entire login before failing with a confusing 5xx-style error. If PKCE is now required, reject it at `/oauth/authorize` with `error=invalid_request` so the client fails fast, and consider advertising PKCE as required in the AS metadata (`server.ts:203`).

**Functionality — token binding is persisted before the PKCE check** (`oauth-manager.ts:248-273`)
The 30-day `TOKEN_BINDING` is written to Redis (lines 248-252) *before* the `!authState.codeChallenge` guard throws (line 268). A rejected (non-PKCE) authorization still leaves a binding row behind. Low blast radius (the binding is keyed by `tokenHash` and identical to what a valid flow would write), but moving the PKCE check up to right after state validation would mean a refused authorization leaves nothing behind.

**Software engineering — no direct tests for the two behavioral changes** (`oauth-manager.ts:212-360`)
The added tests thoroughly cover the *primitive* (`sealAuthState` TTL param, `deriveAuthCodeKey` separation) but not the manager behavior this PR changes: (a) `createAuthorizationCode` now throwing when `codeChallenge` is absent, and (b) `exchangeCode` redeeming a sealed code + rejecting a missing challenge / verifier mismatch. Notably, `exchangeCode` is now **Redis-free** (pure crypto), so it is trivially unit-testable without mocks — seal a code with `authCodeKey()` and assert redemption, the missing-challenge rejection (`oauth-manager.ts:335-337`), and a verifier mismatch. `oauth-state-binding.test.ts` covers only the state/handle binding. Given the repo's TDD convention and the security sensitivity, a small `exchangeCode` test would be cheap insurance.

**Software engineering — leftover dead code** (`oauth-manager.ts:121-123`)
`generateCode()` is now unused (the old `const code = generateCode()` was replaced by `sealAuthState(...)`). It won't fail CI — `@typescript-eslint/no-unused-vars` is set to `'warn'` (`eslint.config.js:29`) and there's no `noUnusedLocals` — but it's leftover from the refactor and should be removed.

**Software engineering — `exchangeCode`'s catch collapses config errors into a client error** (`oauth-manager.ts:320-324`)
The `try { payload = openAuthState(code, authCodeKey()); } catch { throw new Error('Invalid or expired authorization code'); }` swallows *everything*, including `authCodeKey()`/`assertKey` throwing on a missing `INSFORGE_CLIENT_SECRET` (a server misconfiguration). That's exactly the mislabelling the rest of this module is careful to avoid — compare `getAuthorizationState` (`oauth-manager.ts:199-205`), which rethrows anything that isn't an `InvalidAuthStateError`, and the deliberate plain-`Error` vs `InvalidAuthStateError` split in `auth-state.ts:136-142`. Consider catching only `InvalidAuthStateError` here and letting a real misconfiguration propagate.

### Information

**Security — codes are no longer single-use (by design, documented)** (`oauth-manager.ts:254-298`, `305-313`)
RFC 6749 §4.1.2 / OAuth 2.1 expect an authorization code to be single-use with revoke-on-reuse; the old `GETDEL` provided that, the sealed code cannot. The PR documents this and the mitigation is genuine: (a) mandatory PKCE means a code intercepted without the verifier is inert, and (b) redemption is deterministic — a replay yields the same `tokenHash`, i.e. the same session, not a fresh one. The residual gap is a client that leaks *both* code and verifier (e.g. a client that logs its verifier) getting a 5-minute replay window; acceptable here, just worth stating that statelessness was traded for single-use.

**Software engineering — stale JSDoc on `exchangeCode`** (`oauth-manager.ts:301-306`)
The method's doc comment still reads *"Uses atomic GETDEL to prevent authorization code replay attacks"*, which now directly contradicts the implementation and the accurate inline comment seven lines below (`oauth-manager.ts:312`). Please update it — a stale security claim on a security method is worse than no comment.

**Security — pre-existing (out of PR scope):** `server.ts:696` logs the full client redirect URL, which now includes the sealed `code`. The code is inert without the PKCE verifier so risk is low, but since codes are now self-contained bearer values rather than opaque Redis handles, redacting the `code` query param from that log line would be a reasonable follow-up.

**Performance:** net improvement — the code `SETEX` at issue and `GETDEL` at exchange are gone, and `exchangeCode` no longer touches Redis at all. No new loops, allocations, or blocking I/O. No findings.

---

Zero Critical findings. The design is sound and the crypto discipline is careful. The Suggestions (early PKCE rejection, binding-before-check ordering, `exchangeCode` tests, dead-code + stale-doc cleanup) are all non-blocking.

jwfing (commented):
## Review — Authorization codes carry their own record, PKCE required

**Summary:** Sound, well-documented stateless-auth-code change that trades single-use for a PKCE-mandatory sealed code; the crypto and key-separation are correct and the real CI gates (tsc, eslint, vitest) pass — the notes below are all non-blocking.

### Requirements context
No `/docs/superpowers/` or any spec/plan directory exists in this repo (consistent with prior reviews of `insforge-mcp`), so no matching spec was found — assessed against the PR description, the four-of-five series context, and the surrounding code. Verified locally on the head commit: `npm ci` → `tsc --noEmit` (0 errors), `eslint .` (0 errors, warnings only), and the added `auth-state` / `client-id` tests (52 passed). Prettier is **not** a CI gate here (not in `ci.yml`, not wired into `eslint.config.js`, and `master` already has pre-existing prettier violations), so formatting is cosmetic.

---

### Critical
(none)

The central design decision — a sealed code is replayable for its 5-minute lifetime, made acceptable by mandatory S256 PKCE — is an intentional, clearly-surfaced tradeoff, not a defect. The mitigation is genuinely enforced (`oauth-manager.ts:268-273` at issue, `oauth-manager.ts:335-356` at exchange), the verifier never leaves the client, and a replay yields only the same `tokenHash`/session the legitimate holder already has. The three-key separation (`config.ts:182-202`, distinct HMAC label `mcp-auth-code-v1`) correctly prevents a sealed *state* from being redeemed as a *code*, and the new test asserts exactly that. Nothing here rises to blocking.

---

### Suggestions (non-blocking)

**Software engineering — no regression tests for the new `oauth-manager` behavior** (`src/http/oauth-manager.ts:254-357`)
The added tests cover the crypto primitives (`auth-state.test.ts` per-value TTL) and key separation (`config.ts` three-keys), but the *behavioral* changes this PR is actually about — refuse-at-issue when `codeChallenge` is absent, refuse-on-exchange when the sealed code lacks a challenge, replay-with-correct-verifier succeeds — have no committed coverage (there is no `oauth-manager.test.ts` at all). The "Measured" table in the PR body reads like a one-off script rather than a checked-in test. Given how security-critical these paths are, a small `oauth-manager` suite exercising that table would lock the behavior in.

**Functionality — PKCE is rejected late, after auth + a Redis write** (`src/http/oauth-manager.ts:248-273`, surfaced at `src/http/server.ts:698-704`)
A client that authorizes *without* `code_challenge` is accepted at `/oauth/authorize` (`createAuthorizationState` never checks), completes the full platform login and project selection, and only then hits the `PKCE is required` throw in `createAuthorizationCode`. Two consequences worth considering: (1) the token binding is `setex`'d into Redis (`:248-252`) *before* the PKCE check at `:268`, so the reject path leaves an orphaned 30-day binding (inert — the client can't derive its `tokenHash` — but wasteful); (2) the failure surfaces to the client as `server_error` (`server.ts:701`) rather than a client-side `invalid_request`. Since the whole point is to *require* PKCE, rejecting in `createAuthorizationState` (at authorize time, before any work) would fail fast and classify correctly. Low blast radius in practice because the MCP SDK always sends PKCE.

**Software engineering — stale JSDoc contradicts the new behavior** (`src/http/oauth-manager.ts:305`)
`exchangeCode`'s doc comment still reads "Uses atomic GETDEL to prevent authorization code replay attacks," which is the exact opposite of the new implementation (line 312: "No GETDEL, because there is nothing stored"). In a PR whose thesis is being honest about the replay tradeoff, this leftover line is actively misleading — please update or drop it.

**Software engineering — dead code** (`src/http/oauth-manager.ts:121-123`)
`generateCode()` was the random-code generator for the old stored-code path; this PR removed its only caller, so it is now unused (eslint flags it). Safe to delete.

---

### Information
- **Security — the replay/single-use tradeoff is real and residual.** Beyond interception (which PKCE covers), the one thing lost versus single-use is the RFC 6749 §4.1.2 *reuse-detection* signal: an AS that sees a code redeemed twice can revoke the derived tokens. A sealed code can't be marked spent, so that detection is gone within the 5-minute window. Acceptable for public clients under mandatory S256, and correctly called out in the PR — noting it so it's on the record, not asking for a change.
- **Performance — net positive.** The change removes one Redis `SETEX` (code issue) and one `GETDEL` (exchange) per flow, trading them for local AES-GCM seal/open. 275-char sealed code is comfortably within URL/cookie limits. No N+1, no hot-loop, no blocking I/O introduced.
- **Style — one prettier violation in new code** (`src/http/config.ts:194`): the `deriveAuthCodeKey` throw exceeds the 100-col `printWidth` and would be reflowed by `prettier --write`. Cosmetic only — prettier is not enforced in CI here.

---

### Verdict
**approved** (informational — the GitHub green-check remains a separate human action). Zero Critical findings; the crypto, key-derivation separation, expiry-inside-AAD, redirect-URI binding, and S256-only enforcement are all correct, and the intentional replay tradeoff is soundly mitigated by mandatory PKCE. The suggestions above (oauth-manager regression tests, earlier PKCE rejection, stale JSDoc, dead `generateCode`) are worth addressing but none block merge.

jwfing (approved):
LGTM - approved.

PR #92, opened 2026-07-31
A co-tenant could substitute or kill our state cookie

tonychang04:
Quinn found the class, Max verified it on master. I reproduced both against the shipped `readCookie` rather than taking either:

```
Cookie: NAME=v1.ATTACKER; NAME=<ours>   -> we read "v1.ATTACKER"
Cookie: NAME=%E0%A4%A;    NAME=<ours>   -> we read undefined, forever
```

**The second is the worse one and the cheaper one.** `catch { return undefined }` *returns* rather than continues, so a single malformed percent-escape in a neighbour's cookie is a permanent sign-in outage — no crypto, no timing, no knowledge of our internals. My comment claimed the read "fails closed"; it never asked closed **for whom**. Closed for us, open for them.

## Three fixes, in reverse order of how much they matter

```
read EVERY match              shadowing becomes inert — a shadowing value fails
                              to open and the next candidate is tried
malformed value: continue     one bad escape stops being fatal
__Host- prefix, Path=/        a browser refuses to store a __Host- cookie with a
                              Domain, so no co-tenant can set our name at all
```

The first two matter **more** than the prefix, because they survive someone dropping the prefix later — which is exactly how this class comes back.

## Measured after

```
attacker first             ["v1.ATTACKER","v1.OURS"]   ours reachable
attacker MALFORMED first   ["v1.OURS"]                 ours reachable
name on https              __Host-mcp_oauth_state
name on http               mcp_oauth_state
```

`Path=/` is what `__Host-` requires and it costs nothing: a cookie path was never a security boundary — any path on a host can read or set that host's cookies. The prefix buys a guarantee the path never did.

The name falls back **only** when there is no TLS at all, because `__Host-` requires Secure and a Secure cookie is dropped on plaintext http. That is local development and nothing else, and it is keyed on the scheme rather than on a flag someone could set — fallbacks are how protections get lost.

211 tests.

---
## Summary by cubic
Prevents co-tenants on shared domains from substituting or breaking our OAuth state cookie. We now use a `__Host-` cookie on TLS and read all candidate cookies, skipping malformed values, so our state stays reachable.

- **Bug Fixes**
  - Use `authStateCookieName()` to set `__Host-mcp_oauth_state` on HTTPS (fallback to `mcp_oauth_state` only on plain HTTP). Path=/, HttpOnly; Secure follows the public URL.
  - Replace `readCookie` with `readCookies` to collect all matches and continue on decode errors; callers try each candidate.
  - Update server flows to use `openStateFromRequest()` and to set/clear the correct cookie name and attributes.

<sup>Written for commit 6c980888e5fc76f8b72944faa8ce19d49b0277be. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/92?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/auth-state-cookie.test.ts
+++ b/src/http/auth-state-cookie.test.ts
@@ -1,8 +1,9 @@
 import { describe, it, expect } from 'vitest';
 import {
   AUTH_STATE_COOKIE,
+  authStateCookieName,
   cookieAttributes,
-  readCookie,
+  readCookies,
   newStateHandle,
   isAcceptableClientState,
   MAX_CLIENT_STATE_LENGTH,
@@ -35,10 +36,8 @@ describe('cookie attributes', () => {
     expect(cookieAttributes('https://mcp.insforge.dev').sameSite).toBe('lax');
   });
 
-  it('is HttpOnly and scoped to the OAuth routes', () => {
-    const attrs = cookieAttributes('https://mcp.insforge.dev');
-    expect(attrs.httpOnly).toBe(true);
-    expect(attrs.path).toBe('/oauth');
+  it('is HttpOnly', () => {
+    expect(cookieAttributes('https://mcp.insforge.dev').httpOnly).toBe(true);
   });
 
   it('follows the public URL for Secure rather than hardcoding it', () => {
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Review — OAuth state-cookie hardening

**Summary:** A well-reasoned, security-positive fix that neutralizes co-tenant cookie shadowing on shared parent domains by reading *every* candidate cookie (skipping undecodable ones) and switching to a `__Host-` prefixed name on TLS — the core logic is correct and well unit-tested.

**Requirements context:** No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory and no repo-local `.claude/skills/`. Assessed against the PR description, the linked reproduction, and existing conventions in `src/http/`. Verified head SHA `6c98088`.

I confirmed the three-part approach is sound: `authStateCookieName()` gates the `__Host-` prefix on `https://` (so it always pairs with `secure: true`, `path: '/'`, no `Domain` — the exact conditions the prefix requires); `readCookies()` returns all matches and `continue`s past decode errors; and `openStateFromRequest()` tries each candidate against `getAuthorizationState()`, so a shadowing value simply fails to open (AES-256-GCM auth-tag rejection in `openAuthState`) and the next is tried. A forged candidate can't open without the seal key, and the `handle` binding still catches cross-authorization replay. All old `readCookie`/`AUTH_STATE_COOKIE`/`cookieAttributes(url, path)` call sites were updated; no dangling references.

---

### Critical
(none)

### Suggestion

**Functionality — `res.clearCookie` will not remove the `__Host-` cookie on HTTPS** — `src/http/server.ts:701`
```ts
res.clearCookie(authStateCookieName(SERVER_CONFIG.publicUrl), { path: '/' });
```
On TLS the name resolves to `__Host-mcp_oauth_state`, but the deletion `Set-Cookie` here omits `Secure`. Browsers reject *any* `Set-Cookie` carrying a `__Host-` prefix that isn't `Secure` (RFC 6265bis prefix rules apply to expiring cookies too), so the clear is silently dropped and the cookie lingers until its 10-minute `maxAge` — i.e. the defense-in-depth clear this line intends is defeated in exactly the (HTTPS) environment it targets. The old code cleared a non-prefixed cookie and worked, so this is a small regression of the clear step. Blast radius is low (the sealed state self-expires in 10 min and the flow is already complete), hence Suggestion rather than Critical — but if immediate invalidation is considered part of the posture, treat it higher. Fix: pass matching attributes, e.g. `{ path: '/', secure: SERVER_CONFIG.publicUrl.startsWith('https://') }`.

**Software engineering — server-flow behavior is untested** — `src/http/server.ts:113-124`
The array/skip-malformed semantics are thoroughly covered in `auth-state-cookie.test.ts`, which is the right place for the security-critical logic. But `openStateFromRequest()`'s "try each candidate until one opens" loop and the set/clear attribute wiring have no test (there is no `server.test.ts`). This is consistent with the repo's existing lack of a server test harness, so it's not a regression — noting it because a test asserting the clear attributes would have caught the finding above.

### Information

- **Software engineering — inline return type is hard to read** — `src/http/server.ts:115-117`: `NonNullable<Awaited<ReturnType<ReturnType<typeof getOAuthManager>['getAuthorizationState']>>>` resolves correctly to `AuthorizationState`, but importing that type directly would read far more clearly.
- **Functionality — cookie-name migration blip across deploy**: in-flight sign-ins started before this deploy hold the old `mcp_oauth_state` (path `/oauth`) cookie, which won't be read under the new name and won't be cleared by the new `path: '/'` clear. Acceptable given the 10-minute cookie lifetime (stale flows just restart), noted for awareness.

### Security
Net-positive and no new findings: the only new user input (the `Cookie` header) reaches a safe percent-decode with errors caught, then symmetric decrypt with auth-tag verification; no secrets are logged; the `handle`/CSRF binding is preserved; no new dependencies.

### Performance
No concern. `openStateFromRequest` loops over candidate cookies calling `getAuthorizationState`, which is pure AES-256-GCM decrypt + JSON parse (no DB/network in `openAuthState`), and the candidate count is bounded by the HTTP header-size limit. Malformed values are filtered before the loop.

---

### Verdict: `approved` (informational)
Zero Critical findings. Posted as a COMMENT — explicit GitHub approval remains a separate human action. The two Suggestions (notably the HTTPS `__Host-` clear) are worth addressing but are non-blocking.

jwfing (commented):
## Review — #92 A co-tenant could substitute or kill our state cookie

**Summary:** Solid, well-reasoned defense against OAuth state-cookie shadowing on a shared parent domain — the two behavioral fixes (`readCookies` returns every match and skips malformed values; callers try each candidate) plus the `__Host-` name are all sound and correctly test-covered; one non-blocking bug in the cookie-clear path.

### Requirements context
No spec/plan directory in this repo (`docs/superpowers/` and `docs/` contain no specs for `insforge-mcp`) — assessed against the PR description, the shipped-code behavior it cites, and surrounding code in `src/http/`.

I verified the central premise of the loop in `openStateFromRequest` (server.ts:114-124): "a shadowing value simply fails to open and the next candidate is tried." `getAuthorizationState` (oauth-manager.ts:182-207) converts every attacker-reachable failure into `null` because `openAuthState` (auth-state.ts:84-127) throws only `InvalidAuthStateError` on bad prefix / bad base64url / GCM-verify failure / non-JSON / expiry — the sole raw `throw` is the 32-byte-key server-misconfig, which is not attacker-controlled. So a hostile candidate never escapes the loop as an exception. The premise holds. ✅

---

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — the defense-in-depth clear silently no-ops on HTTPS** — `src/http/server.ts:701`
  ```ts
  res.clearCookie(authStateCookieName(SERVER_CONFIG.publicUrl), { path: '/' });
  ```
  On TLS the cookie is set with `{ secure: true, path: '/' }` (auth-state-cookie.ts:102-112, `cookieAttributes`), but the clear passes only `{ path: '/' }`. Express's `clearCookie` does not add `secure` on its own, and browsers clear a cookie only when the emitted attributes match those it was set with (path/secure/domain). Worse, the emitted `Set-Cookie: __Host-mcp_oauth_state=; Path=/; …` lacks `Secure`, and the `__Host-` prefix *requires* Secure — so a compliant browser rejects the clearing header outright. Net effect: on the real (HTTPS) deployment the cookie is never proactively cleared; it lingers until `maxAge` (10 min). Not a re-introduction of the shadowing bug and the state is replayable within its window anyway (as the comment at server.ts:698-700 notes), so it's genuinely defense-in-depth — but the code intends to clear and doesn't. Suggest matching the set options, e.g.
  ```ts
  res.clearCookie(authStateCookieName(SERVER_CONFIG.publicUrl), {
    path: '/',
    secure: SERVER_CONFIG.publicUrl.startsWith('https://'),
  });
  ```
- **Software engineering — `openStateFromRequest` has no direct test** — `src/http/server.ts:114-124`. The parse layer (`readCookies`) is very well covered, including attacker-first vs ours-first ordering and malformed-value skipping. But the "try each candidate until one opens, return the first that does" orchestration is only exercised transitively. A small unit test (attacker-sealed value first, ours second → returns ours) would lock in the behavior the PR title is about at the layer that actually gates sign-in.

#### Information
- **Software engineering — verbose inline return type** — `src/http/server.ts:117`. The `NonNullable<Awaited<ReturnType<ReturnType<typeof getOAuthManager>['getAuthorizationState']>>>` return annotation is hard to read; extracting a named `AuthorizationState`-based alias would be clearer. Purely taste.
- **Software engineering — `AUTH_STATE_COOKIE` is now referenced only by tests** — `src/http/auth-state-cookie.ts:61`. It's correctly marked `@deprecated` and still exported; after this PR nothing in production code reads it (server.ts uses `authStateCookieName(...)` everywhere). Fine to keep for the http-name literal and test fixtures — noting it so it isn't mistaken for a live production path.

---

### Dimension coverage
- **Software engineering:** Strong TDD — tests renamed to describe the threat, cover every branch of `readCookies` (multi-match ordering, malformed-skip, no-header/empty/valueless, substring non-match) and `authStateCookieName`/`cookieAttributes`. Only gap is the server-level orchestration test above (Suggestion).
- **Functionality:** The fix does what it claims; all callers migrated to `openStateFromRequest`/`authStateCookieName` consistently, no leftover `readCookie`/`AUTH_STATE_COOKIE` call sites, and the removed `oauthPathPrefix` param has no other callers. One clear-path bug (Suggestion).
- **Security:** This is a security *improvement* and introduces no new input path — the sealed state remains encrypted + HttpOnly, `getAuthorizationState` stays a non-oracle (uniform `null`, distinct-wording-same-type logging). Broadening `Path=/oauth` → `Path=/` is correctly reasoned (path was never a boundary; cookie stays same-origin + HttpOnly). No new deps.
- **Performance:** The candidate loop is bounded and cheap — on HTTPS a co-tenant *cannot* inject a `__Host-`-named cookie (the prefix forbids a `Domain`, locking it to the exact host), so at most our own value is present, and each `getAuthorizationState` is in-memory AES-GCM with no Redis/network round-trip. No N+1, no blocking I/O.

---

### Verdict
**approved** (informational — human approval still required via the approve flow). Zero Critical findings. The two Suggestions and two Information notes are non-blocking; the `clearCookie` fix is worth a one-line follow-up but does not gate the merge.

jwfing (approved):
LGTM - approved.

PR #93, opened 2026-07-31
Refuse a missing code_challenge at authorize, not at the end

tonychang04:
#91 required PKCE — and put the refusal in `createAuthorizationCode`, which runs **after** the browser sign-in and the project picker. Measured on the slug once it deployed:

```
authorize with no code_challenge  ->  302
```

So a client without PKCE gets redirected, the person signs in, chooses a project, and *only then* hits the refusal. They did all of the work before we told them we were never going to accept it.

Same mistake as validating a `redirect_uri` at authorize instead of at registration: **reject where the caller is listening**, not where only a browser can see it. The MCP client reads the authorize response; nothing reads the one at the end of the flow.

```
no code_challenge     400  invalid_request, naming code_challenge
with code_challenge   302
```

The error says *why* rather than just *what*: we advertise S256 as our only `code_challenge_method` and issue codes that carry their own state, so PKCE is what makes a code safe to accept.

211 tests.

---
## Summary by cubic
Reject authorize requests missing a PKCE `code_challenge` at the authorize endpoint so clients get an immediate, actionable error. This prevents users from completing sign-in and project selection only to be rejected at the end.

- **Bug Fixes**
  - Validate `code_challenge` in the authorize handler; missing now returns 400 `invalid_request` with a clear message.
  - With `code_challenge`: unchanged (302).

<sup>Written for commit 022db38727b55a2451c0934726ae348167154eca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/93?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -343,6 +343,29 @@ app.get(OAUTH_ENDPOINTS.authorize, async (req: Request, res: Response) => {
     });
   }
 
+  if (!code_challenge) {
+    // Refused HERE, not at the end.
+    //
+    // #91 made authorization codes stateless, which means they cannot be
+    // single-use, which means PKCE is what stops a replay. I put the refusal in
+    // createAuthorizationCode — and measured on the slug that a client without
+    // PKCE still gets a 302, signs in through the browser, picks a project, and
+    // only THEN fails. The person did all the work before we told them we were
+    // never going to accept it.
+    //
+    // That is the same mistake as validating a redirect_uri at authorize
+    // instead of at registration: reject where the caller is listening, not
+    // where only a browser can see it. The MCP client reads this response; it
+    // never reads the one at the end of the flow.
+    return sendOAuthError(req, res, 400, {
+      error: 'invalid_request',
+      error_description:
+        'code_challenge is required. This server advertises S256 as its only ' +
+        'code_challenge_method and issues authorization codes that carry their own state, ' +
+        'so PKCE is what makes a code safe to accept.',
+    });
+  }
+
   if (response_type !== 'code') {
     return sendOAuthError(req, res, 400, {
       error: 'unsupported_response_type',


jwfing (commented):
## Review: Refuse a missing `code_challenge` at authorize, not at the end

**Summary:** Correctly moves the PKCE refusal from `createAuthorizationCode` (after browser sign-in + project picker) up to the `/oauth/authorize` GET where the MCP client actually reads the response — a small, well-motivated security/UX fix with no Critical issues.

### Requirements context
This repo has no `/docs/superpowers/` (or any `docs/`) directory, so there is no matching spec/plan — assessed against the PR description, the referenced #91 (which introduced stateless authorization codes and the original end-of-flow refusal), and the existing OAuth flow in `src/http/server.ts` / `src/http/oauth-manager.ts`.

### Findings

#### Critical
(none) — the change is consistent with #91's intent, returns a proper `400 invalid_request` directly to the caller (no redirect to an unvalidated `redirect_uri`, so no open-redirect), and the defense-in-depth refusals remain in place downstream: `oauth-manager.ts:268` (`createAuthorizationCode`) and the token-endpoint verifier at `oauth-manager.ts:335-354`. No existing client is newly broken — #91 already rejected PKCE-less clients, just later in the flow.

#### Suggestion

- **Software engineering — no test for the new behavior** (`src/http/server.ts:346-367`). The PR body cites "211 tests" but changes exactly one file and adds no test. The behavior it advertises — *"no code_challenge → 400 invalid_request naming code_challenge"* — is asserted nowhere. There is currently no supertest/route-level harness in this repo (every `app.get/post` handler in `server.ts` is untested; `auth-challenge.test.ts` only covers pure metadata helpers), so this is consistent with existing coverage — but given the repo's TDD convention, extracting the guard into a testable predicate (as was done for `state` via `isAcceptableClientState`) or adding a route test would lock in the regression this PR is fixing.

- **Security — `code_challenge` is unbounded in the sealed cookie** (`src/http/server.ts:336-344` vs `346-367`; `oauth-manager.ts:165-175`). `state` is length-bounded to `MAX_CLIENT_STATE_LENGTH` (512) *precisely because* it rides inside the sealed `AUTH_STATE_COOKIE`. `code_challenge` rides in the same cookie (spread via `...params` into `authState`) but has no length/format check — an oversized value produces an oversized cookie, the exact class the `state` guard exists to prevent. For S256 a `code_challenge` is always a 43-char base64url string, so a cheap `isAcceptable…`-style bound here would be a natural companion to this PR's theme. (Pre-existing — `code_challenge` was already passed through before this PR — but adjacent to the code you're touching.)

- **Functionality — the error message promises method validation this branch doesn't do** (`src/http/server.ts:360-366`). The message states the server "advertises S256 as its only code_challenge_method," but the guard only checks *presence*. A request with `code_challenge` present and `code_challenge_method=plain` passes this check and is instead rejected inside `createAuthorizationState` (`oauth-manager.ts:155-156`) via a thrown `Error`, which the `catch` at `server.ts:449-458` turns into a generic `server_error` ("Failed to initiate authorization") rather than a clean `invalid_request` naming the method. By this PR's own "reject where the caller is listening" principle, an unsupported method would ideally be refused right here alongside the missing-challenge case. (Also pre-existing; noting for consistency since you're editing this exact block.)

#### Information

- **Ordering.** The new guard runs before `response_type`, `client_id`, and `redirect_uri` validation. Because the error is returned directly (never redirected) this introduces no open-redirect and no meaningful client-existence oracle — validating request shape before crypto-verifying the `client_id` is reasonable. Noted only for awareness.
- **Performance:** no relevant changes — the addition is a single synchronous presence check with no added I/O or allocation in the hot path.
- **Comment style** (`server.ts:347-359`) references internal narrative ("measured on the slug", "john-bot's Critical") — this matches the surrounding file's existing prose-comment convention, so purely informational.

### Verdict
**approved** (informational — the GitHub green check remains a separate human action). No Critical findings; posted as a `COMMENT`. The three Suggestions are non-blocking; the missing test is the most worth addressing since it's the guarantee this PR exists to provide.

jwfing (commented):
## Review: Refuse a missing `code_challenge` at authorize, not at the end

**Summary:** A tight, well-motivated fix that moves the PKCE-presence refusal from `createAuthorizationCode` (end of flow, after browser sign-in + project pick) to the `/oauth/authorize` handler (start), so a non-PKCE client is told immediately instead of after doing all the interactive work — no Critical issues found.

### Requirements context
This repo has no `/docs/superpowers/` (or `docs/specs/`) directory — `docs/` is empty. No matching spec/plan found, so I assessed against the PR description and the linked prior work (#91, which made auth codes stateless and thus required PKCE). The change is faithful to the stated intent: *reject where the caller is listening.*

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — no test for the new behavior** (`src/http/server.ts:346-367`). The PR adds no test, and the diff touches only `server.ts`. The repo's convention is to extract validatable logic into pure helpers that get unit tests (e.g. `isAcceptableClientState`, `isRegisteredRedirectUri` in `client-id.test.ts`); this check is inline in the un-exported `app` handler, so it isn't directly exercised. Note the broader PKCE-required behavior (`oauth-manager.ts:268`) also lacks a direct test, so this isn't a coverage regression — but a close-call refusal on the security-critical authorize path is exactly the kind of thing a regression test should pin. Consider extracting a small `requirePkce(query)`-style predicate (mirroring the existing helpers) and unit-testing the presence/empty/array cases.

#### Information
- **Functionality — check precedence** (`src/http/server.ts:346` vs `:379-418`). The new `!code_challenge` refusal runs *before* client-id signature verification (`readClientId`) and the registered-`redirect_uri` match. So a request that is both missing PKCE *and* carries a bogus/unregistered `client_id` now reports the PKCE error first rather than "Unknown client_id". This is safe — both paths return a direct 400 JSON (no redirect, so no open-redirect / code-theft exposure), and the PKCE requirement is already public in the AS metadata (`code_challenge_methods_supported`, `:203`) — so nothing is disclosed that isn't already advertised. Flagging only so the ordering is a deliberate choice on record.
- **Security — presence-only validation** (`src/http/server.ts:346`). The check gates on presence, not shape. A compliant S256 challenge is a fixed ~43-char base64url string, but an oversized `code_challenge` is forwarded into `createAuthorizationState` and sealed into the state cookie unbounded, unlike `state` which is length-guarded by `isAcceptableClientState`. This is pre-existing (the value has been forwarded since #91) and out of scope for this PR — noting it only as a candidate for a future follow-up alongside the other authorize-time bounds.

### Positives
- Defense-in-depth is preserved rather than replaced: the end-of-flow refusal still stands in `createAuthorizationCode` (`oauth-manager.ts:268`), so the early check layers on top of the backstop instead of removing it.
- No analogous late-refusal gap for `code_challenge_method`: a non-`S256` method is already rejected early inside `createAuthorizationState` (`oauth-manager.ts:155`), whose throw is caught by the handler's `try/catch` and returned before any redirect.
- The refusal is a net performance win — it short-circuits before the browser round-trip and platform/Redis work.
- Style matches the surrounding validations and the file's comment-heavy convention; `sendOAuthError` is called correctly per its signature (`oauth-error-response.ts:222`) and content-negotiates HTML vs JSON, so both a human in a tab and an MCP client get an appropriate response.

### Verdict
**approved** (informational — no Critical findings; the single Suggestion is a non-blocking test-coverage nit). Posted as a COMMENT; explicit GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #94, opened 2026-07-31
The token stops being a Redis key: `mcp:auth:binding:` cutover

tonychang04:
**This is the cutover.** `mcp:auth:binding:` — the last Redis use in the OAuth path — is deleted here, not prepared for. The prefix, the `setex`/`get`/`del` calls and the `getRedisClient` import all leave `oauth-manager.ts`, and `server.ts` is rewired to the new token in the same commit. After this, OAuth runs with Redis unconfigured; only MCP **sessions** still need it.

An earlier draft of this description said "two modules, no wiring yet". That was true of the first commit and stopped being true; read the file list, not that sentence.

## What changes, in order of blast radius

**1. The bearer stops being a Redis key.** `resolveProjectFromToken` no longer looks anything up: it opens the token. `/oauth/token` returns the sealed value instead of a token hash, and `expires_in` goes from 30 days to 24h. Every token issued before this deploy stops resolving — the binding rows they point at are no longer read. Clients re-authorize; there is no migration path and none is worth building for a value with a 30-day tail and a working `/authorize`.

**2. A stolen token now reaches every project the user can, not one.** The sealed envelope carries the **platform access token**, because `list-projects` / `create-project` / `switch-project` need platform authority and today we discard that token after the picker uses it once. Tony chose this explicitly ("they can have more permission", 07-29 10:07) and Max has it on the record. It is the largest security change in the rewrite and it lands with this merge — a reviewer should see it as chosen, not inherited.

**3. Revocation moves to the platform.** There is no row to delete, so `/oauth/revoke` drops the cached project key and returns 200 (RFC 7009 §2.2), and enforcement is the platform's `validateAccessToken` `revoked` check on every call. The project-key cache TTL is therefore the bound on how stale a revocation can be — 60s, fixed. That is the design, not a tuning knob: caching bounded by `exp`, which I first specced, would have made revoke a no-op from our side.

**4. `POST /api/projects/{id}/bind` can no longer re-point a token.** The project is sealed inside the token the client already holds. 200 with `{id}` only if it is already scoped there, 409 otherwise. Silently succeeding would lie; issuing a new token from an endpoint that never authenticated anyone would be worse.

**5. `/api/projects` unseals the bearer before calling the platform** — the one real bug fixed here. It forwarded our bearer straight through, which worked only while the bearer *was* a platform token; since the seal it got a 401, taking `list-projects` with it. Every place that forwarded the bearer now has to forward what is *inside* it. This was the last one, and nothing in the suite calls this endpoint.

**6. The `project_binding_required` 401 becomes `authorization_required`** on `/mcp` and `/sse`, with reauth guidance. Clients matching on the old string see a new one.

## The two new modules

`access-token.ts` — AES-256-GCM, fourth derived key, fourth label. Encrypted rather than signed for a stronger reason than the auth state: a signed-but-readable bearer would publish a platform credential to anyone who saw the header, including our own logs. The payload is `userId` / `projectId` / `platformAccessToken` and nothing else — every field platform-issued and bounded, after `projectName` (upstream `z.string().min(2)`, no `.max`) turned out to let one org member's long project name mint an unusable token for every other member. `MAX_ACCESS_TOKEN_LENGTH` 4096 is the belt: measured against the real proxy chain, ~3.3x a real token, and it throws at sign-in where the error page can explain it rather than at a proxy that will not.

`project-key-cache.ts` — the project API key is deliberately *not* in the token: it is a credential with its own lifetime, and sealing it would freeze it for the token's whole life so a rotated key would keep working. Fetched per request, cached 60s, keyed by user **and** project (two users can reach one project with different grants; a project-only key would serve one user's key to the other). Iris measured what the cache buys: **~60ms of platform server time, location-independent**, on a route that does nothing — dominant on tools that otherwise take 20ms, which is why it ships with the change rather than after it. Eviction is rate-limited on the write path, and there is a test that it does not leak — the binding leaked because nothing ever removed it.

## Testing

221 tests, and they cover the two new modules thoroughly. **They do not cover the wiring**: `server.ts` `/api/projects` and `/bind`, and `resolveProjectFromToken`'s cache hit/miss/refusal branches have no tests — which is exactly where the bug in (5) was hiding. Per the auth/sessions norm, a QA pass on the real path — login through restart, against a running server — is a merge condition here, and is worth more than more unit coverage.

## Still on Redis after this

MCP session state (`/mcp`, `/sse`). That is the next one.

---
## Summary by cubic
Replaces Redis-bound tokens with a sealed, self-describing access token and removes `mcp:auth:binding:` from the OAuth path. OAuth now works without Redis; revocation is enforced by the platform with a 60s user+project cache that is leak-free and collision-safe.

- **Migration**
  - OAuth no longer uses Redis; only MCP sessions still do.
  - `/oauth/token` returns the sealed access token (not a token hash); `expires_in` is 24h.
  - Token payload is only `userId`, `projectId`, and `platformAccessToken`; tokens over 4096 chars are refused at issue.
  - Project API key is fetched per request and cached 60s (keyed by user+project); cache evicts stale entries, prevents key collisions, and `forgetUser` drops all per-user entries.
  - `/oauth/revoke` always returns 200 per RFC 7009 and drops cached keys for that user.
  - `POST /api/projects/{id}/bind` cannot re-point a token; 200 if already scoped (returns only `projectId`), 409 if scoped to a different project.
  - 401s now use `authorization_required` with reauth guidance.

- **Bug Fixes**
  - `/api/projects` opens the sealed bearer before calling the platform; `list-projects` works again.

<sup>Written for commit e1f44e81a408982576668d7931013d381ea90ea2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/94?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * OAuth exchanges now issue encrypted access tokens that expire after 24 hours.
  * Tokens securely preserve project and account context without persistent token bindings.
  * Project access information is cached briefly to improve request performance.
* **Bug Fixes**
  * Invalid, expired, oversized, or tampered tokens are rejected consistently.
  * Unauthorized MCP and SSE requests now receive appropriate authorization prompts.
  * Attempts to switch a token between projects are prevented.
  * Revoking access now promptly clears cached project information.

### DIFF
--- a/src/http/access-token.test.ts
+++ b/src/http/access-token.test.ts
@@ -1,0 +1,132 @@
+import { describe, it, expect } from 'vitest';
+import { createHmac } from 'crypto';
+import {
+  issueAccessToken,
+  readAccessToken,
+  ACCESS_TOKEN_TTL_SECONDS,
+  MAX_ACCESS_TOKEN_LENGTH,
+  type AccessTokenPayload,
+} from './access-token.js';
+import { deriveAuthStateKey } from './config.js';
+
+/**
+ * The most valuable envelope we issue, and it shipped without a test file.
+ * That is the gap this closes — the fields it carries were argued about in
+ * review and pinned nowhere.
+ */
+
+const key = createHmac('sha256', 'a-test-secret').update('mcp-access-token-v1').digest();
+const otherKey = createHmac('sha256', 'a-different-secret').update('mcp-access-token-v1').digest();
+const T0 = 1_700_000_000_000;
+
+const payload: AccessTokenPayload = {
+  userId: 'user_1',
+  platformAccessToken: 'eyJ.a-platform-jwt.sig',
+  projectId: 'proj_1',
+};
+
+describe('the access token', () => {
+  it('comes back as what went in', () => {
+    const token = issueAccessToken(payload, key, T0);
+    expect(readAccessToken(token, key, T0)).toEqual(payload);
+  });
+
+  it('carries ONLY platform-issued, bounded fields', () => {
+    // The test Max's finding earns. `projectName` was in here, a project name
+    // is `z.string().min(2)` upstream with no `.max`, and it rode inside every
+    // token issued for that project — so one org member's 10k-character name
+    // minted an unusable token for every other member.
+    //
+    // Asserting the exact key set rather than the absence of one name: the
+    // defect was a class (caller-influenced data in a credential), and
+    // `organizationId`, `accessHost` and `userEmail` were in the same class,
+    // three of them read from the payload by nothing at all.
+    const token = issueAccessToken(payload, key, T0);
+    const opened = readAccessToken(token, key, T0)!;
+    expect(Object.keys(opened).sort()).toEqual(['platformAccessToken', 'projectId', 'userId']);
+  });
+
+  it('refuses to issue a token no proxy would carry', () => {
+    // The belt. Bounded by construction is what the payload is FOR, so this
+    // should never fire in production — it fires when someone adds a field that
+    // is not bounded, which is precisely the mistake that has now been made
+    // three times in one day on three different envelopes.
+    const huge = { ...payload, platformAccessToken: 'x'.repeat(MAX_ACCESS_TOKEN_LENGTH) };
+    expect(() => issueAccessToken(huge, key, T0)).toThrow(/Refusing to issue an access token/);
+  });
+
+  it('does not print the payload when it refuses', () => {
+    // A length is enough to act on. This message reaches logs.
+    const secret = 'super-secret-platform-token-'.repeat(200);
+    try {
+      issueAccessToken({ ...payload, platformAccessToken: secret }, key, T0);
+      throw new Error('expected a refusal');
+    } catch (error) {
+      expect((error as Error).message).not.toContain(secret);
+      expect((error as Error).message).toMatch(/limit 4096/);
+    }
+  });
+
+  it('leaves realistic room: a real token is far under the cap', () => {
+    // If this ever gets close, the cap is doing nothing and the number needs
+    // re-deriving from a fresh measurement of the chain, not raising.
+    const token = issueAccessToken(payload, key, T0);
+    expect(token.length).toBeLessThan(MAX_ACCESS_TOKEN_LENGTH / 2);
+  });
+
+  it('is null, never a throw, for anything that is not ours', () => {
+    // Every caller treats "not our token" as an ordinary 401, and #95 is what
+    // happens when one of these paths throws instead: a 500 with no
+    // WWW-Authenticate, which tells an MCP client to give up rather than to
+    // sign in again.
+    expect(readAccessToken('', key, T0)).toBeNull();
+    expect(readAccessToken('not-a-token', key, T0)).toBeNull();
+    expect(readAccessToken('v1.STALE_FROM_AN_EARLIER_FORMAT', key, T0)).toBeNull();
+    expect(readAccessToken('v2.' + 'A'.repeat(80), key, T0)).toBeNull();
+  });
+
+  it('is null under a different key, so rotation signs everyone out', () => {
+    // The stated and only revocation mechanism for the sealed values: rotating
+    // INSFORGE_CLIENT_SECRET. Worth a test because it is also the operational
+    // cost Max flagged — client ids, auth states and access tokens all die at
+    // once.
+    const token = issueAccessToken(payload, key, T0);
+    expect(readAccessToken(token, otherKey, T0)).toBeNull();
+  });
+
+  it('is null once tampered with', () => {
+    const token = issueAccessToken(payload, key, T0);
+    const flipped = token.slice(0, -1) + (token.endsWith('A') ? 'B' : 'A');
+    expect(readAccessToken(flipped, key, T0)).toBeNull();
+  });
+
+  it('expires 24 hours out, not 30 days', () => {
+    // The binding it replaced lived 30 days and refreshed on use. A value that
+    // cannot be revoked directly should not live for a month — the TTL IS the
+    // revocation window for the token itself, so a shortening here is a
+    // security property and a lengthening needs an argument.
+    const token = issueAccessToken(payload, key, T0);
+    const ttlMs = ACCESS_TOKEN_TTL_SECONDS * 1000;
+    expect(readAccessToken(token, key, T0 + ttlMs - 1)).toEqual(payload);
+    expect(readAccessToken(token, key, T0 + ttlMs)).toBeNull();
+    expect(ACCESS_TOKEN_TTL_SECONDS).toBe(24 * 60 * 60);
+  });
+
+  it('is sealed, not signed: the platform token is not readable from it', () => {
+    // The reason this is AES-GCM and a client id is an HMAC. A
+    // signed-but-readable bearer would publish a platform credential to anyone
+    // who saw the header — including our own logs.
+    const token = issueAccessToken(payload, key, T0);
+    const body = Buffer.from(token.slice('v1.'.length), 'base64url').toString('utf8');
+    expect(token).not.toContain(payload.platformAccessToken);
+    expect(body).not.toContain(payload.platformAccessToken);
+    expect(body).not.toContain('proj_1');
+  });
+
+  it('uses a key that is not the auth-state key', () => {
+    // Same secret, different label. A token sealed for one purpose must not
+    // open under another, or the domain separation is decorative.
+    const token = issueAccessToken(payload, key, T0);
+    expect(readAccessToken(token, deriveAuthStateKey('a-test-secret'), T0)).toBeNull();
+  });
+});
--- a/src/http/access-token.ts
+++ b/src/http/access-token.ts
@@ -1,0 +1,86 @@
+import { sealAuthState, openAuthState, InvalidAuthStateError } from './auth-state.js';
+
+/**
+ * The access token we hand an MCP client, as a value rather than a Redis key.
+ *
+ * Until now our bearer was literally a lookup key: `resolveProjectFromToken`
+ * took it, read `mcp:auth:binding:<token>`, and got back everything that gave
+ * it meaning. The token said nothing. Delete the store and every issued token
+ * became a meaningless string — not because OAuth was missing, but because we
+ * never put anything IN the token. That is why a restart signed everyone out,
+ * and it is the last thing keeping Redis alive.
+ *
+ * So carry it. Encrypted, not signed, and for a stronger reason than the auth
+ * state: this envelope holds the PLATFORM access token. A signed-but-readable
+ * bearer would publish a platform credential to anyone who saw the header.
+ *
+ * WHAT IS DELIBERATELY NOT IN HERE: the project's API key. It is a credential
+ * with its own lifetime, and sealing it would freeze it for the token's whole
+ * life — so a rotated or revoked project key would keep working. It is fetched
+ * per request instead, from a cache with a short fixed TTL. See
+ * project-key-cache.ts for why the TTL is fixed rather than bounded by `exp`.
+ *
+ * REVOCATION MOVES TO THE PLATFORM, and that is the trade to understand. The
+ * binding row could be deleted; a sealed token cannot. What replaces it is the
+ * platform's own revocation — `validateAccessToken` checks `revoked` on every
+ * call — which is where revocation was really enforced all along. The bound on
+ * how stale that can be is the project-key cache TTL, which is why it is 60
+ * seconds and not "until exp".
+ */
+
+export interface AccessTokenPayload {
+  /** Who this is, from the platform at login. */
+  userId: string;
+  userEmail: string;
+
+  /**
+   * The platform access token.
+   *
+   * This is what makes the MCP able to act as the user without storing
+   * anything — and what makes this envelope the most valuable thing we issue.
+   * It is also what lets platform-level tools (list/create/switch project)
+   * exist at all: they need this, not a project key.
+   */
+  platformAccessToken: string;
+
+  /** The project chosen at sign-in. A per-call or per-URL project replaces
+   *  this later; it is here now so nothing else has to change at once. */
+  projectId: string;
+  projectName: string;
+  organizationId: string;
+  accessHost: string;
+}
+
+/**
+ * How long an issued token is accepted.
+ *
+ * The binding it replaces used 30 days, refreshed on use. This is deliberately
+ * shorter: a value that cannot be revoked directly should not live for a month.
+ * The platform token inside has its own expiry too, and whichever runs out
+ * first ends the session — which is the safe direction.
+ */
+export const ACCESS_TOKEN_TTL_SECONDS = 24 * 60 * 60;
+
+export function issueAccessToken(
+  payload: AccessTokenPayload,
+  key: Buffer,
+  nowMs: number = Date.now()
+): string {
+  return sealAuthState(payload, key, nowMs, ACCESS_TOKEN_TTL_SECONDS);
+}
+
+/**
+ * Recover the payload, or null.
+ *
+ * Null rather than a throw because every caller treats "not our token" as an
+ * ordinary 401 — an expired sign-in is the common case here, not an attack,
+ * and a caller that could tell the two apart would be an oracle.
+ */
+export function readAccessToken(token: string, key: Buffer, nowMs: number = Date.now()): AccessTokenPayload | null {
+  try {
+    return openAuthState<AccessTokenPayload>(token, key, nowMs);
+  } catch (error) {
+    if (error instanceof InvalidAuthStateError) return null;
+    throw error;
+  }
+}
--- a/src/http/access-token.ts
+++ b/src/http/access-token.ts
@@ -28,10 +28,43 @@ import { sealAuthState, openAuthState, InvalidAuthStateError } from './auth-stat
  * seconds and not "until exp".
  */
 
+/**
+ * EVERY FIELD IN HERE IS PLATFORM-ISSUED AND BOUNDED. That is a property to
+ * preserve, not a coincidence, and it was not true when this file was first
+ * written.
+ *
+ * The first version also carried `projectName`, `organizationId`, `accessHost`
+ * and `userEmail`. Max found the hole in `projectName`: cloud-backend's
+ * `createProjectSchema` is `z.string().trim().min(2)` — a `.min` with no
+ * `.max` — so a project name is unbounded, and it rode inside every token
+ * issued for that project. A 10k-character name mints a token no
+ * `Authorization` header will carry, and the person who breaks it need not be
+ * the person who named it: one org member names the project, every member who
+ * selects it gets an unusable token.
+ *
+ * Dropping it cost nothing, which is the part worth recording. Those four
+ * fields were WRITTEN here and read almost nowhere:
+ *
+ *   projectName · organizationId · accessHost   re-fetched by
+ *                                               resolveProjectFromToken anyway,
+ *                                               which reads them off the
+ *                                               project-key cache, never off
+ *                                               the payload
+ *   userEmail                                   written at issue, read by
+ *                                               nothing at all
+ *
+ * So the seal was carrying four values for the benefit of one informational
+ * field in one endpoint's response body. Same reasoning as keeping the project
+ * API key out: display data that can be re-fetched does not belong in a
+ * credential.
+ *
+ * The rule for anyone adding a field: it must be platform-issued and bounded,
+ * or it does not go in. A caller-influenced string in here is a denial of
+ * service against everyone who shares the resource it names.
+ */
 export interface AccessTokenPayload {
   /** Who this is, from the platform at login. */
   userId: string;
-  userEmail: string;
 
   /**
    * The platform access token.
@@ -46,9 +79,6 @@ export interface AccessTokenPayload {
   /** The project chosen at sign-in. A per-call or per-URL project replaces
    *  this later; it is here now so nothing else has to change at once. */
   projectId: string;
-  projectName: string;
-  organizationId: string;
-  accessHost: string;
 }
 
 /**
@@ -61,12 +91,58 @@ export interface AccessTokenPayload {
  */
 export const ACCESS_TOKEN_TTL_SECONDS = 24 * 60 * 60;
 
+/**
+ * The longest token we will hand out — the belt to go with the braces above.
+ *
+ * The payload is bounded by construction now, so this should never fire. It
+ * exists because that sentence has been wrong three times today: a client id, a
+ * sealed state and this token were each "bounded" until a field turned out not
+ * to be. The one remaining variable-length value in here is the platform's own
+ * access token, which is not ours to bound.
+ *
+ * The number comes from measurements rather than from a round constant. Iris
+ * drove the real chain — Cloudflare, the Fly proxy, Node's 16KB default — with
+ * a synthetic bearer, and the rest is this seal measured directly:
+ *
+ *   15512 chars   accepted by the whole chain
+ *   15573 chars   431 Request Header Fields Too Large
+ *    1483 chars   what the old payload minted
+ *   ~1250 chars   what this payload mints (the four dropped fields cost 233)
+ *
+ * So 4096 is ~3.3x the real token, and it is reached when the PLATFORM's own
+ * token passes ~2950 characters — roughly 3.8x its size today. Far enough away
+ * that tripping this means something changed that someone should look at, and
+ * far enough below 15512 that the refusal happens here, where it is explained,
+ * rather than at a proxy that will not say why.
+ *
+ * It THROWS rather than truncating or warning, and the failure lands at
+ * sign-in: the callback's error handler turns it into the OAuth error page, so
+ * a person sees "something went wrong on our side" at the moment it went wrong.
+ * The alternative is worse in the way that is hard to debug — a token that
+ * every proxy in the chain silently refuses to carry, surfacing later as a
+ * client that cannot connect for reasons nothing logs.
+ */
+export const MAX_ACCESS_TOKEN_LENGTH = 4096;
+
 export function issueAccessToken(
   payload: AccessTokenPayload,
   key: Buffer,
   nowMs: number = Date.now()
 ): string {
-  return sealAuthState(payload, key, nowMs, ACCESS_TOKEN_TTL_SECONDS);
+  const token = sealAuthState(payload, key, nowMs, ACCESS_TOKEN_TTL_SECONDS);
+
+  if (token.length > MAX_ACCESS_TOKEN_LENGTH) {
+    // Deliberately does not print the token or the payload: this is the most
+    // valuable envelope we issue and a length is enough to act on.
+    throw new Error(
+      `Refusing to issue an access token of ${token.length} characters ` +
+        `(limit ${MAX_ACCESS_TOKEN_LENGTH}). A token this size will be rejected by proxies ` +
+        'in the request path, so issuing it would produce a client that cannot connect and ' +
+        'no explanation. Something has been added to the sealed payload that is not bounded.'
+    );
+  }
+
+  return token;
 }
 
 /**
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -201,6 +201,26 @@ export function authCodeKey(): Buffer {
   return deriveAuthCodeKey(INSFORGE_CONFIG.clientSecret);
 }
 
+/**
+ * And a fourth, for the access token we issue to MCP clients.
+ *
+ * Fourth purpose, fourth label, no relationship to the other three. This one
+ * seals the most valuable envelope we produce — it carries the platform access
+ * token — so it is the one that most needs its own key.
+ */
+const ACCESS_TOKEN_KEY_LABEL = 'mcp-access-token-v1';
+
+export function deriveAccessTokenKey(clientSecret: string): Buffer {
+  if (!clientSecret) {
+    throw new Error('INSFORGE_CLIENT_SECRET is required: the access token key is derived from it');
+  }
+  return createHmac('sha256', clientSecret).update(ACCESS_TOKEN_KEY_LABEL).digest();
+}
+
+export function accessTokenKey(): Buffer {
+  return deriveAccessTokenKey(INSFORGE_CONFIG.clientSecret);
+}
+
 // ============================================================================
 // Redis Configuration
 // ============================================================================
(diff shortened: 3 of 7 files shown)


jwfing (commented):
## Review — #94 "The token stops being a Redis key: the two pieces"

**Summary:** Two well-crafted, self-contained modules (`access-token.ts`, `project-key-cache.ts`) plus a fourth derived key — pure groundwork with no wiring, faithfully mirroring the existing `auth-state.ts` / `config.ts` patterns; no blocking issues.

### Requirements context
No matching spec/plan found — this repo has no `docs/superpowers/` (or any `docs/`) spec directory, so I assessed against the PR description alone. The PR is explicitly scoped as "two modules, no wiring yet"; I confirmed that by grep — `access-token`/`project-key-cache`/`accessTokenKey` are referenced only within the new files themselves, so nothing in the running OAuth path changes yet. That matches the stated intent, so scope discipline is clean.

### Findings

#### Critical
(none)

#### Suggestion

- **[software engineering] `src/http/access-token.ts` has no test file.** The cache module ships 6 tests, but `access-token.ts` (0 tests) has behavior worth pinning that isn't covered by the underlying `auth-state.test.ts`:
  - the `readAccessToken` contract that an `InvalidAuthStateError` becomes `null` while a *misconfiguration* error (the `assertKey` 32-byte throw) is re-raised (`access-token.ts:78-85`) — that branch is the whole reason the wrapper exists, and it's untested;
  - that `ACCESS_TOKEN_TTL_SECONDS = 24h` is actually applied (`access-token.ts:52-63`).
  A short round-trip test (`issueAccessToken` → `readAccessToken`, plus tamper→`null` and wrong-key-re-throw) would match the tests-as-design bar the rest of this PR sets.

- **[functionality/performance] `ProjectKeyCache` only expires entries on read, so a never-re-read entry is never freed** (`project-key-cache.ts:53-63`). The doc comment says such an entry "costs one Map slot for 60 seconds," but nothing sweeps it — if that exact `(user, project)` key is never `get()` again, the `Entry` lives until `forgetUser` or process restart, not 60s. With many distinct users/projects over a long-lived process the `Map` grows unbounded with dead entries. This is the same lazy-removal property the PR calls out as the binding's flaw ("the binding leaked because nothing ever removed it"), and the `size()`-based leak test (`project-key-cache.test.ts:54-61`) only proves removal *on the read that finds it stale*, never removal without a read. Blast radius is modest (short TTL, bounded active user set), so non-blocking — but consider either a periodic sweep or correcting the comment to state the eviction is read-triggered only.

- **[security] Cache key uses a space delimiter (`` `${userId} ${projectId}` ``, `project-key-cache.ts:35`), which is fragile for a tenant-isolation key.** The module's stated purpose is preventing cross-tenant key bleed, yet `("a b","c")` and `("a","b c")` collide to the same key, and `forgetUser`'s `` `${userId} ` `` prefix scan (`project-key-cache.ts:79-83`) would over- or under-match if an id contained a space. Platform ids are UUID/prefixed-slug shaped (no spaces) so this is not exploitable today — hence Suggestion, not Critical — but a delimiter that cannot appear in an id (or a nested `Map<userId, Map<projectId, Entry>>`) would make the isolation structural rather than dependent on id format.

#### Information

- **[security] The blast-radius change is sound and explicitly owned.** Carrying the platform access token in an *encrypted* (not signed) envelope, under a distinct 4th derived key (`config.ts:211-218`), with `exp` sealed inside the AES-256-GCM payload, is the correct construction and is consistent with the reviewed `auth-state.ts`. The "stolen token reaches every project the user can" trade is called out in the PR body as a deliberate product decision (Tony, on the record) — flagging only so it's visible as chosen, not something to relitigate in code review. No secrets are logged and `readAccessToken` returns `null` rather than leaking failure detail.

- **[software engineering] Conventions followed cleanly:** `.js` ESM import extensions, the derive-key label pattern (`mcp-access-token-v1`), lazy `accessTokenKey()` accessor, and the empty-secret guard all match the three existing keys verbatim (`config.ts:166-218`).

### Verdict
**approved** (informational — human approval is still a separate action). No Critical findings; the three Suggestions are hardening/coverage items the author can take or defer, and none block the "argue with the mechanism before wiring" goal of this PR. Posting as a COMMENT per the zero-Critical rule.

jwfing (commented):
## Review — #94 "The token stops being a Redis key: the two pieces"

**Summary:** Two well-reasoned, self-contained modules (a self-describing encrypted access token + a short-TTL per-user/project key cache) that lay the groundwork for removing `mcp:auth:binding:`; the code is clean, the crypto reuse is correct, and the design decisions are documented — I found no blocking issues, just a couple of things worth tightening before the wiring PR.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — no matching spec/plan found, so I assessed against the PR description and the surrounding code (`auth-state.ts`, `config.ts`, `oauth-manager.ts`) alone. The PR is explicitly groundwork with **no wiring**, which is respected: `issueAccessToken`/`readAccessToken`/`accessTokenKey`/`getProjectKeyCache` have no callers yet (verified via grep). That is in-scope and fine.

---

### Critical
(none)

### Suggestion

**Software engineering — `access-token.ts` ships without a test (`src/http/access-token.ts:64-86`).** `project-key-cache.ts` got six focused tests, but the module the PR itself calls *"the largest security change in the rewrite"* has none. It's a thin wrapper over the (presumably tested) `auth-state.ts` primitives, but it adds new branching worth pinning: the seal→open round-trip preserves every `AccessTokenPayload` field, an expired token returns `null` (not a throw), a tampered/wrong-key blob returns `null`, and — importantly — the `readAccessToken` catch re-throws anything that is *not* an `InvalidAuthStateError` (e.g. the plain `Error` from `assertKey` on a misconfigured key) rather than swallowing it to `null`. A three-or-four-case test would lock that contract in before wiring.

**Functionality — the "does not leak" property is narrower than the comment claims (`src/http/project-key-cache.ts:39-49`).** Eviction happens only on a *subsequent read* that finds an entry stale (or on `forgetUser`). The comment on `get` says an unread entry "costs one Map slot for 60 seconds," but nothing reclaims it at 60s — a `(user, project)` pair that is `set` once and never read again (a one-shot request from a user who doesn't return) leaves a dead entry indefinitely, so under churn across many distinct users the `Map` grows without an upper bound. The test *"does not leak entries once they expire"* (`src/http/project-key-cache.test.ts:54-61`) only passes because it performs a read after expiry. Given the module explicitly frames itself against the old binding that *"leaked because nothing ever removed it,"* a lightweight periodic sweep (or dropping expired siblings on `set`) would make the code match its stated intent. Low blast radius (tiny per-entry cost, unwired), hence Suggestion rather than blocking.

**Security/robustness — cache-key delimiter (`src/http/project-key-cache.ts:37` and `forgetUser` at `:79-83`).** `cacheKey` joins with a literal space (`` `${userId} ${projectId}` ``) and `forgetUser` matches on the prefix `` `${userId} ` ``. This is collision-safe and prefix-safe *only* if neither `userId` nor `projectId` can ever contain a space. Today they can't — they're platform IDs (`user.id` / `projectAccess.projectId`, per `oauth-manager.ts:427,453`) — so there is **no live bug**. But a delimiter that cannot appear in either component (a control char, or a structured key) would remove a latent cross-tenant footgun for zero cost, and it's cheap insurance given the module's whole reason for keying on the pair is to *prevent* cross-tenant key bleed.

### Information

- **Security — the headline trade is deliberate and the mechanism is sound.** The token envelope now carries the platform access token, so a stolen bearer reaches every project the user can (not one). The author flagged this explicitly as chosen. The implementation is correct: AES-256-GCM via `sealAuthState` (integrity + confidentiality from one primitive), a dedicated 4th derived key (`deriveAccessTokenKey`, HMAC-SHA256 of `clientSecret` with its own `mcp-access-token-v1` label — no key reuse across the four purposes), expiry carried *inside* the authenticated payload so it can't be extended, and a 24h TTL that is correctly shorter than the 30-day binding it replaces. No new security weakness is introduced by these two modules themselves. The revocation-staleness argument (bound = the 60s cache TTL, since `validateAccessToken` re-checks `revoked` per call) is coherent.
- **`config.ts` change is clean.** `deriveAccessTokenKey`/`accessTokenKey` mirror the existing `deriveAuthCodeKey`/`authCodeKey` sibling pattern exactly; `createHmac` was already imported. No concern.
- **Performance.** The ~60ms per-request platform round-trip is a real cost acknowledged in the PR, and the 60s cache is the mitigation shipping alongside it — the right call. Nothing blocking is added to a hot path here since the modules are unwired; the only perf-adjacent note is the unbounded-`Map`-growth point above.

---

### Verdict: **approved** (informational — a human still gives the explicit GitHub approval)

Zero Critical findings. Solid, carefully-documented groundwork. The two Suggestions (a test for `access-token.ts`, and making the cache's eviction actually match its "no leak" claim) are worth folding in before or alongside the wiring PR, but none of them block this merge.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: self-describing access token + project-key cache

**Summary:** A well-reasoned, well-commented move from a Redis-bound bearer to a sealed self-describing access token plus a short-lived project-key cache; the cryptography is sound and I found nothing blocking — a few non-blocking notes below.

### Requirements context
No matching spec/plan found — this repo has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the four inline module rationales, and the surrounding code in the workspace.

### Findings

#### Critical
(none)

#### Suggestion

- **[functionality/process] The description materially understates the scope — this PR *is* the switch-over, not just groundwork.** The body says *"Two modules, no wiring yet … so the mechanism can be argued with before the switch-over rather than after"* and *"Not in this PR: … the actual removal of `mcp:auth:binding:`."* But the diff fully removes the Redis binding path (`TOKEN_BINDING_PREFIX`, `getTokenBinding`/`touchBinding`/`revokeBinding`/`bindTokenToProject`, and the `getRedisClient` import) and rewires the live flow: `/oauth/token` now returns the sealed token (`src/http/server.ts:741-772`) and `resolveProjectFromToken` reads it (`src/http/oauth-manager.ts:349-397`). After merge, the OAuth path no longer uses Redis at all. This matters because the "platform access token now lives inside the bearer" trade-off ships **now**, so it should be evaluated as live rather than as a proposal to argue with later. Worth correcting the description so a reviewer doesn't under-scrutinize.

- **[software engineering / test coverage] No direct test for `access-token.ts`.** `src/http/access-token.ts:63-86` is new and security-critical, but only `project-key-cache.test.ts` was added. `auth-state.test.ts` covers the underlying GCM seal/open, yet there is no test asserting `readAccessToken` returns `null` on `InvalidAuthStateError` while *rethrowing* other errors (`access-token.ts:80-85`), nor that the 24h TTL is actually applied. The repo's TDD convention would expect a round-trip / tamper / expiry / wrong-key test for this wrapper.

- **[performance] `ProjectKeyCache` can grow unbounded for set-and-abandoned entries.** Eviction happens only lazily on `get` when an entry is found expired (`src/http/project-key-cache.ts:64-72`) or via `forgetUser`; there is no periodic sweep (contrast `session-manager.ts:546`, which does sweep). An entry that is `set` once and never read again (a one-shot user/project) stays in the `Map` until process restart. The comment at `project-key-cache.ts:60` — *"an entry nobody asks for again costs one Map slot for 60 seconds"* — is therefore inaccurate: the value is stale-bounded to 60s, but the memory is not reclaimed for 60s; it persists for the process lifetime. Blast radius is low (small entries, MCP restarts on deploy), and the "does not leak entries" test (`project-key-cache.test.ts:54-61`) only exercises the read-path eviction, not this case. Consider a small sweep or size cap, and fix the comment.

#### Information

- **[security] Platform access token inside the bearer — trade-off acknowledged and encryption is sound.** As the description requests, noting it as chosen: a stolen token now reaches every project the user can. Mitigations check out — AES-256-GCM with a random 96-bit nonce and `exp` carried *inside* the authenticated payload (`src/http/auth-state.ts:60-134`), sealed with a distinct HMAC-SHA256-derived 32-byte key (`src/http/config.ts:205-218`), and no key reuse across the four labels. 24h TTL (`access-token.ts:53`) vs. the old 30-day binding is a reasonable tightening.

- **[functionality] Revocation-authority claim verified.** On a cache miss, `resolveProjectFromToken` calls `getProjectAccess(platformAccessToken, …)` (`oauth-manager.ts:367`), which fetches project + API key from the platform (`insforge-api.ts:209-228`); a revoked/expired platform token yields a platform 401 → thrown → caught → `null` → 401 for the caller. So revocation is enforced by the platform and bounded by the 60s cache, exactly as documented.

- **[functionality] Stale hint to the bind endpoint.** The `authorization_required` challenges (`server.ts:974` and the SSE equivalent) and the startup banner still tell clients to `POST /api/projects/{projectId}/bind`, but `bind` can no longer re-point a token — it returns 409 for a different project and 401 for an expired/invalid one (`server.ts:865-905`). For the expired/refused case that this challenge fires on, the hint points at an endpoint that can't fix it; consider dropping the bind reference from that message.

- **[robustness] `readAccessToken` rethrows non-`InvalidAuthStateError`.** `access-token.ts:80-85` returns `null` only for `InvalidAuthStateError`; a misconfiguration (e.g. empty client secret → `assertKey` throws a plain `Error`) would propagate out of `resolveProjectFromToken` (`oauth-manager.ts:349`) as a 500 rather than a 401. Only reachable on server misconfiguration, so acceptable, but noting it.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings: the cryptographic design is correct, revocation semantics hold, and no callers were orphaned. The suggestions above (description accuracy, `access-token.ts` test coverage, cache growth/comment) are non-blocking.

jwfing (commented):
## Review — self-describing access token (`access-token.ts` + `project-key-cache.ts`)

**Summary:** A clean, well-argued move from a Redis-keyed bearer to a sealed self-describing token plus a short-TTL project-key cache; the cryptography and domain separation are sound, but a few behavioural consequences of going stateless (revoke, transient-error handling, advertised `expires_in`) deserve a second look before the switch-over.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the inline design commentary, and the surrounding OAuth code (`auth-state.ts`, `config.ts`, `insforge-api.ts`, `session-manager.ts`).

---

### Critical
(none) — the happy path is intact, key derivation is correctly domain-separated (`mcp-access-token-v1` vs the three existing labels, verified by the `deriveAuthStateKey` negative test), AES-256-GCM gives confidentiality+integrity, and no secret (platform token / apiKey / bearer) is written to logs.

### Suggestion

**1. Functionality/Security — `/oauth/revoke` no longer revokes the token** (`src/http/server.ts:858-882`)
Revoke now only calls `getProjectKeyCache().forgetUser(userId)`. Because the sealed token still carries a valid `platformAccessToken`, the very next request re-fetches the project key from the platform and succeeds — the token stays usable for its full 24h TTL. So an RFC 7009 call against a leaked bearer is effectively a no-op (it just forces one re-fetch). The PR frames this as "revocation moves to the platform," but the platform only revokes if the *platform* token is revoked, which this endpoint does not trigger. This is the consequence of removing all server-side state, and it's partially acknowledged — but the specific fact that "revoke does not invalidate the token" is worth stating explicitly (README / endpoint doc), and consider whether the platform token should be actively revoked here.

**2. Functionality/Resilience — transient platform errors are treated as revocation** (`src/http/oauth-manager.ts:363-380`)
On a cache miss, *any* rejection from `getProjectAccess` (network timeout, platform 5xx, rate-limit) is caught and turned into `return null` → `authorization_required` → the client is told to re-run the whole OAuth flow. The old binding held `apiKey` in Redis for 30 days, so a transient platform blip never signed users out. Now every user re-hits the platform each time the 60s cache lapses, and a brief platform hiccup forces a full reauth. Consider distinguishing 401/403 (genuine "no longer authorized" → return null) from 5xx/network (transient → surface a retryable 503 rather than reauth). The inline comment ("the platform saying this user may no longer reach this project") only holds for the authz-refusal case.

**3. Functionality — advertised `expires_in` doesn't match token lifetime** (`src/http/server.ts:801-810`, `src/http/access-token.ts:ACCESS_TOKEN_TTL_SECONDS`)
The PR body and cubic summary say `expires_in` is "24h or platform token expiry, whichever is sooner," but the code returns `ACCESS_TOKEN_TTL_SECONDS` (a flat 24h). The seal's `exp` is also a flat 24h and does not fold in the platform token's own expiry. So when the inner platform token expires first, the token dies (via a `getProjectAccess` 401) while the client still believes it has time left. Either compute the real minimum for `expires_in`, or adjust the description to match the code.

**4. Performance/Correctness — one-shot cache entries are never reclaimed** (`src/http/project-key-cache.ts:44-56`)
Expiry is only checked on `get`. The comment says an unread entry "costs one Map slot for 60 seconds," but without a sweeper an entry that is never read again is retained indefinitely, not 60s. For a user who authenticates once and never returns within the window, the slot leaks. Blast radius is tiny (small entries, and active clients re-read constantly), so this is minor — but the comment is inaccurate and the `does not leak entries` test only passes because it re-`get`s after expiry.

**5. Software engineering / test coverage — new wiring branches are untested**
The two new modules have excellent, intention-revealing tests. But the wiring changes don't: `resolveProjectFromToken`'s cache-miss-error→null path (`oauth-manager.ts:363-380`), and the `/oauth/revoke` cache-drop and `/api/projects/:id/bind` 200-vs-409 branches (`server.ts`) have no accompanying tests. These are exactly the branches where finding #1/#2 live; a test pinning "revoke drops the cache but the token still resolves" and "bind returns 409 for a different project" would document the intended behaviour.

### Information

- **PR body vs. diff mismatch:** the body says "Two modules, no wiring yet" and lists "the wiring … Not in this PR," but the diff fully wires the token endpoint, `resolveProjectFromToken`, `/oauth/revoke`, and `/api/projects/:id/bind`, and removes `getTokenBinding`/`touchBinding`/`revokeBinding`/`bindTokenToProject`. The cubic summary is the accurate one; consider reconciling the human-written description so reviewers scope correctly.
- **Removed direct raw-token path:** `resolveProjectFromToken` previously accepted a raw Insforge OAuth token for "direct API access"; it now only accepts the sealed token. Confirm no client relies on the raw-token path (not called out in the description).
- **Duplicated guidance text** in the `authorization_required` `error_description` ("… Authorize again. Complete the OAuth flow …") reads slightly redundantly (`server.ts:1027`, `server.ts:1213`).
- **Security posture, acknowledged:** a stolen token now reaches every project the user can, not one. The PR records this as an explicit, on-the-record decision — flagging only so it's visible in the review trail; no action requested.
- **No single-flight on cold cache:** concurrent first requests for the same `user+project` each issue the (2-call) `getProjectAccess`. Low impact given the 60s TTL; noting for completeness.

---

No Critical findings. The design is careful and the crypto is correct. The Suggestions above — chiefly the revoke-is-now-a-no-op semantics (#1) and transient-error→reauth conflation (#2) — are worth resolving or explicitly documenting before the wiring/switch-over PR lands, but none block this groundwork.

<sub>Static review only — `node_modules` absent in the review workspace, so the vitest/tsc/eslint gates were not executed here.</sub>

jwfing (commented):
## Review — self-describing access token + project-key cache

**Summary:** Two well-factored, thoroughly-tested modules that replace the `mcp:auth:binding:` Redis row with a sealed, self-describing access token plus a short-lived per-request project-key cache; the cryptography, key separation, and cache semantics are sound, and I found nothing blocking.

### Requirements context
No `docs/superpowers/` or `docs/specs/` directory exists in this repo (consistent with insforge-mcp being spec-less), so **no matching spec/plan was found — assessing against the PR description, the surrounding OAuth code, and the migration intent (removing the last OAuth-path Redis use) alone.** The PR explicitly scopes out the wiring/removal of `mcp:auth:binding:`, so this is groundwork; I reviewed the two modules and their partial hook-in (`resolveProjectFromToken`, `/oauth/revoke`, `/api/projects/:id/bind`, `/oauth/token`).

### Verification performed
- Confirmed the seal is AES‑256‑GCM with `exp` carried **inside** the authenticated payload and a `v1.` version guard (`auth-state.ts:60-134`) — tamper/expiry/wrong-key all collapse to the same `InvalidAuthStateError`, and `readAccessToken` correctly maps only that to `null` and rethrows anything else.
- Confirmed the fourth derived key uses a distinct label with no relationship to the other three (`config.ts:211-222`); the test asserts a token sealed for one purpose won't open under the auth-state key.
- Confirmed all removed methods (`bindTokenToProject`, `revokeBinding`, `touchBinding`, `getTokenBinding`, `getBindingByToken`) have **no remaining references** anywhere in `src/`; `hashToken` is still defined and used.
- Confirmed PKCE is still required at issue and verified (S256, `plain` rejected) at exchange (`oauth-manager.ts:243-248, 308-332`).
- Confirmed the `>4096` issue-time throw is caught by the `/oauth/select-project` handler and surfaced without leaking the token/payload (`server.ts:762-766`).

---

### Findings

#### Critical
(none)

#### Suggestion

**Functionality / robustness — a platform outage now presents as "sign in again"** (`oauth-manager.ts:363-387`, `insforge-api.ts:209-228`)
`resolveProjectFromToken` treats *every* `getProjectAccess` failure identically — cache-miss → `catch` → `return null` → 401 `authorization_required`. But `getProjectAccess` throws `InsforgeApiError` for 5xx **and** `node-fetch` throws raw network errors when the platform is unreachable. After the 60s cache lapses, a brief platform blip logs users out with "your sign-in is no longer valid — authorize again", and the re-auth they're told to perform will *also* fail while the platform is down. Because the project key is now fetched per request, every tool call depends on platform reachability (mitigated only by the 60s cache) where the 30-day binding previously did not. Consider branching on `InsforgeApiError.statusCode`: 401/403 → genuine revocation (401), 5xx/network → a retryable 503 rather than a re-auth prompt.

**Functionality / UX — the `/mcp` 401 still advises an endpoint that no longer binds** (`server.ts:1025-1031`)
The challenge says "Complete the OAuth flow **or call POST /api/projects/{projectId}/bind**", but after this PR `/bind` can no longer establish a binding — it only reports `200` (already scoped), `409` (different project), or `401` (invalid token). None of those resolve an `authorization_required` state, so the guidance points at a now-defunct remedy. The SSE challenge (`server.ts:1210-1216`) already dropped that clause, so the two are now inconsistent — align them.

**Security / design — `/oauth/revoke` no longer revokes the issued token** (`server.ts:858-882`)
Revoke now only calls `forgetUser` to flush the 60s cache; the sealed access token itself keeps resolving for up to 24h (or until the platform token/grant lapses). A client calling RFC 7009 revoke will reasonably believe its token is dead when it is not. This is the acknowledged consequence of stateless tokens ("revocation moves to the platform"), and the always-200 behavior is correctly preserved — flagging only so a reviewer confirms an *effectively advisory* revoke endpoint is the intended end state for this migration.

#### Information

- **`ProjectKeyCache.size()` isn't wired to `/health` yet** (`project-key-cache.ts:138-142`) — the comment says "For /health, so the cache is observable," but `size()` is currently referenced only by tests. Consistent with "no wiring yet"; the comment just slightly overstates the present state.
- **`expires_in` is always advertised as 24h** (`server.ts:801-810`) — the PR body says "24h or platform token expiry, whichever is sooner," but the response hard-codes `ACCESS_TOKEN_TTL_SECONDS`; the "whichever is sooner" is enforced *implicitly* (platform-token expiry surfaces later as a 401), not in the advertised value. Minor description/impl mismatch, harmless since `expires_in` is a hint.
- **No single-flight on cache miss** (`oauth-manager.ts:359-386`) — concurrent first requests for the same `(user, project)` each issue a `getProjectAccess` round-trip. Bounded and low-risk at a 60s TTL; noting only.

---

### Dimension coverage
- **Software engineering:** Excellent. Both new modules ship with focused test files exercising the exact edge cases (exact key set in the payload, length cap without leaking, wrong-key → null, TTL boundary, user+project isolation, the joined-string collision, prefix-`forgetUser`, leak eviction). Import style, error handling, and naming match the surrounding `http/` code.
- **Functionality:** Solves the stated goal (token becomes a value, not a Redis key). Gaps are the outage-as-401 conflation and the stale `/bind` guidance above — both non-blocking.
- **Security:** Strong. Correct AES‑256‑GCM sealing with `exp` under the tag, four-way key domain separation, platform token never readable from the bearer, payload bounded + length-capped against proxy DoS, PKCE intact, RFC 7009 200-always preserved, no secrets logged (revoke/refusal logs carry only a platform `userId`). The deliberately-broadened token scope ("reaches every project the user can") is explicitly documented and owner-approved.
- **Performance:** The per-request platform round-trip (~60ms) is the intended cost and is bounded by the 60s cache; eviction is O(n) but rate-limited to once/second on the write path with correctness held on the read path. No N+1, no event-loop blocking, no unbounded growth (the prior leak is fixed and tested).

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). No Critical findings; the three Suggestions are worth addressing (especially the outage→401 conflation) but none should block the merge of this groundwork.

jwfing (commented):
## Review: self-describing access token (groundwork)

**Summary.** A carefully-reasoned, well-tested move from Redis-key bearer tokens to a sealed, self-describing access token plus a short-TTL project-key cache; the crypto is sound, the two new pure modules are thoroughly covered, and I found no blocking issues.

### Requirements context
This repo has no `/docs/superpowers/` (or `docs/specs/`) directory — **no matching spec/plan found, so I assessed against the PR description, linked design notes in the code comments, and RFC 6749/7009/7591/9728 as cited.** The intent (remove the `mcp:auth:binding:` Redis dependency from the OAuth path; carry auth state in an encrypted, revocation-bounded token) is clearly stated in the PR body and the module headers, and the implementation matches it.

---

### Critical
(none)

No correctness, security, or data-loss defect found. Specifically verified:
- **Crypto (`auth-state.ts`, `access-token.ts`):** AES-256-GCM, fresh random 96-bit nonce per seal, `exp` sealed *inside* the authenticated payload (cannot be extended by a holder), 32-byte keys, `final()`/tag rejection on tamper. Four purpose-separated keys via distinct HMAC labels (`config.ts:129-222`). Tamper / rotation / cross-label rejection are all tested (`access-token.test.ts:88-131`).
- **No secret leakage:** the issue-refusal error deliberately omits token/payload (tested, `access-token.test.ts:58-68`); logs use fingerprints and `userId` only; the platform token is encrypted, not signed-readable (`access-token.test.ts:115-124`).
- **Wiring is complete and consistent:** `mcp:auth:binding:` is fully gone from the OAuth path (grep confirms no remaining reads/writes), and field mappings from `getProjectAccess` → `ProjectKeyCache` → `resolveProjectFromToken` line up (`insforge-api.ts:219-227`, `oauth-manager.ts:361-399`).
- **Cache correctness:** user+project nesting prevents cross-tenant serving and delimiter collisions; read-path expiry + rate-limited write-path eviction bound the previously-unbounded leak (`project-key-cache.ts:99-159`, tests `project-key-cache.test.ts:61-133`).

### Suggestion

**1. Software engineering — the wiring itself has no test coverage.** The two new *pure* modules are excellently tested, but the glue that consumes them — `server.ts` `/api/projects` unseal (`server.ts:907-915`), `/api/projects/:id/bind` (`server.ts:942-983`), and `oauth-manager.resolveProjectFromToken` cache-hit / miss / platform-refused branches (`oauth-manager.ts:345-400`) — has none (no `oauth-manager.test.ts` / `server.test.ts`). This matters because the one real bug this PR fixes (`/api/projects` forwarding the *sealed* bearer to the platform, `server.ts:897-910`) lived in exactly this untested area and was caught by reasoning rather than a test. A small unit test around `resolveProjectFromToken` (hit returns cached, miss fetches+caches, platform-refusal → `null`) and the `/api/projects` unseal path would guard the regression that already happened once.

**2. Security/robustness — the state cookie carries the same unbounded platform token but has no length guard.** `issueAccessToken` gets a belt-and-braces `MAX_ACCESS_TOKEN_LENGTH` throw (`access-token.ts:125-146`) precisely because the platform token inside is "not ours to bound." But after the callback, `attachPlatformToken` folds that *same* platform token into the sealed **auth-state** and re-seals it into the `__Host-` cookie (`oauth-manager.ts:119-121`, `server.ts:650-654`) with no equivalent guard. The cookie-size budget table in `auth-state-cookie.ts:23-26` (`no client_id, client state bounded to 512 → 1159, fits`) was measured *before* the platform token was folded in, so it understates the real post-callback cookie by ~1.25 KB. Today (~1250-char platform token) it's comfortably under 4096; but the exact growth scenario the access-token guard anticipates (platform token > ~2950 chars) would push this cookie over the 4096-byte browser limit, and the browser drops it **silently** → the user lands on `select-project` with "Session expired" and no explanation — the same silent-failure class the access-token guard was written to prevent, one hop earlier and unguarded. Consider a symmetric size assertion (or a note in the budget table) on the post-callback state.

### Information

**1. Functionality/docs — the PR description's "no wiring yet" is stale.** The prose intro says *"Two modules, no wiring yet"* and *"Not in this PR: The wiring, and therefore the actual removal of `mcp:auth:binding:`."* The actual diff **does** fully wire both modules and removes `mcp:auth:binding:` from the OAuth path (confirmed: `server.ts` and `oauth-manager.ts` consume the new tokens; no binding references remain). The cubic and CodeRabbit auto-summaries correctly describe the wired state. Worth reconciling the human-written intro so reviewers/changelog aren't misled about scope.

**2. Security — blast-radius expansion is real and (per the description) chosen.** Sealing the platform access token means a stolen MCP bearer reaches every project the user can, and grants platform-level authority (list/create/switch-project), not one project. The PR body records this as an explicit product decision ("they can have more permission"); flagging here for reviewer visibility, not as an objection. Relatedly, the 60s cache-hit path (`oauth-manager.ts:362-363`) skips the platform's per-call `revoked` check, so a revocation can be up to 60s stale — a documented, bounded trade.

### Performance
No regressions in the changed code. The added per-request cost is the platform round-trip on a cache miss (~60ms, measured and documented), amortized by the 60s cache; eviction is O(n) but rate-limited to ≤1/s on the write path, and `size()` is O(n) only on `/health`. The N+1 in `getAllUserProjects` is pre-existing and untouched.

---

### Verdict: **approved** (informational — human approval via the separate approve flow)
Zero Critical findings. The suggestions are worth addressing before this becomes the default path (especially #1, given the wiring is where the sole real bug hid), but none block merge.

*Note: `npm`/`vitest` could not be executed in the review sandbox (no `node_modules`); this is a static review. The PR self-reports 221 passing tests.*

jwfing (approved):
LGTM - approved.

PR #95, opened 2026-07-31
A stale token gets a challenge, not a stack trace

tonychang04:
Iris measured this on the live slug, and it is not hypothetical.

```
no Authorization header   401 + WWW-Authenticate      correct
Bearer <stale/garbage>    500, raw Express stack,     wrong
                          MaxRetriesPerRequestError, NO challenge
```

`resolveProjectFromToken` reached Redis, threw, and the throw escaped to Express's default handler.

**"Not a 500" is not the same as "a 401".** The `WWW-Authenticate` header is the only thing that tells an MCP client to re-run OAuth — a 500 reads as *the server is broken, give up*, so a client holding an old token never recovers on its own.

**This is live now.** The token format changed three times today (#82, #87/#89, #94) and each invalidated everything issued before it, so every client holding an earlier one lands on this path — including Tony's, and including every existing user at the cutover. This branch is what decides between "it re-authenticated itself" and "it broke".

## Fixed at the wrapper, not the call sites

```
no header              401 + challenge
stale/garbage bearer   401 + challenge
not even our prefix    401 + challenge
```

Every route already goes through that one function, so a future caller cannot reintroduce the escape.

## The honest cost

A genuine infrastructure fault is now also reported as an authorization failure, which mislabels it. That is the right trade **only** because the client's available action is identical either way — re-authorize — and because a silent 500 gives it no action at all. The fault is logged at error level so it stays visible to us even though the client is told something softer. That reasoning is in the code, not just here.

215 tests.

---
## Summary by cubic
Ensure stale or invalid bearer tokens return 401 with a WWW-Authenticate challenge instead of 500, so MCP clients re-authenticate automatically and don’t see stack traces.

- **Bug Fixes**
  - Wrapped `resolveProjectFromToken` in try/catch; on error, return null to trigger 401 + `WWW-Authenticate`.
  - Log failures at error level; no errors escape to Express’s default handler.
  - Consistent behavior for no header, stale/garbage tokens, and non-matching prefixes.

<sup>Written for commit bbbf9ba1382773676c10f309199e2a8bade26a52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/95?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -144,7 +144,32 @@ async function resolveProjectFromToken(token: string): Promise<{
   oauthTokenHash: string;
 } | null> {
   const oauthManager = getOAuthManager();
-  return oauthManager.resolveProjectFromToken(token);
+  try {
+    return await oauthManager.resolveProjectFromToken(token);
+  } catch (error) {
+    // NOTHING here may escape, and that is the whole point of this wrapper.
+    //
+    // Measured on the live slug: a stale bearer — one issued before today's
+    // token format changed — reached Redis, threw MaxRetriesPerRequestError,
+    // and Express answered with a 500 and a raw stack. No WWW-Authenticate.
+    //
+    // "Not a 500" is not the same as "a 401". The challenge header is the ONLY
+    // thing that tells an MCP client to re-run OAuth; a 500 reads as "the
+    // server is broken, give up", so a client holding an old token never
+    // recovers on its own. Iris found this, and it is not hypothetical: the
+    // token format changed three times today and every client holding an
+    // earlier one lands here, as will every existing user at the cutover.
+    //
+    // The honest cost, stated rather than hidden: a genuine infrastructure
+    // fault is now also reported as an authorization failure, which
+    // mislabels it. That is the right trade only because the client's
+    // available action is identical either way — re-authorize — and because a
+    // silent 500 gives it no action at all. The fault is logged at error level
+    // so it stays visible to us even though the client is told something
+    // softer.
+    console.error('[OAuth] Could not resolve a token; answering with a challenge:', error);
+    return null;
+  }
 }
 
 /**


jwfing (commented):
## Review: A stale token gets a challenge, not a stack trace

**Summary:** Correct, well-scoped defensive fix — wrapping `resolveProjectFromToken` so a throw (e.g. `MaxRetriesPerRequestError` from Redis) becomes `null` instead of escaping to Express's 500 handler, which routes stale/garbage tokens through the existing 401 + `WWW-Authenticate` challenge path.

**Requirements context:** No matching spec/plan found — this repo has no `docs/superpowers/` (or any spec dir); assessing against the PR description alone. Intent is clear from the body and the inline comment.

I verified the mechanism end to end: at `src/http/server.ts:946` / `:1133` a `null` return lands in the `if (!projectInfo)` block, and for an OAuth token with no legacy headers reaches `sendUnauthorized(...)`, which sets `WWW-Authenticate` and returns 401 (`src/http/auth-challenge.ts:71-78`). So the throw→null→challenge chain the PR claims is real. `getOAuthManager()` is a trivial singleton getter (`oauth-manager.ts:514-519`), so the one call left outside the `try` won't reintroduce the escape in practice.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — no regression test for the fixed behavior** (`src/http/server.ts:147-172`). The PR body cites "215 tests," but that's the existing suite passing; no test exercises the new catch. There is no `server.test.ts`, and the only reference to `server.ts` in tests (`auth-challenge.test.ts:245`) is an unrelated static-grep guard for `MCP_SERVER_URL`. The repo's stated convention is TDD/regression tests for bug fixes, and this is exactly the kind of close-call a test should pin: a future refactor of the route wiring could silently reintroduce the 500-escape and nothing would fail. Standing up a route-level test (supertest against `app` with a mocked `oauthManager.resolveProjectFromToken` that throws) — or at minimum a unit test asserting the wrapper swallows a throw and returns `null` — would guard it. Non-blocking given no existing harness, but worth adding.

#### Information
- **Functionality — challenge message mislabels an invalid token as valid.** A thrown/stale token now flows into the `oauthToken && !legacyApiBaseUrl` branch (`server.ts:957-964`, `:1143-1148`) whose body says *"OAuth token is valid but not bound to a project."* For a token that is actually invalid (or a token whose lookup transiently failed), that wording is slightly misleading. It's harmless — the `WWW-Authenticate` header is what actually drives a spec-compliant client to re-run OAuth, and this is pre-existing behavior for any token that resolves to `null` rather than throwing — so no change is required. Noting only in case a distinct "could not verify token" message would aid debugging.
- **Software engineering — `getOAuthManager()` at `server.ts:146` sits outside the `try`.** If it ever threw, the escape the wrapper exists to prevent would recur. It's a no-throw singleton getter today, so effectively zero risk; mentioned only for completeness.

**Security:** No concerns. The catch logs the `error` object (not the token) at error level, no secret/PII is added to any response, and the change strengthens rather than weakens auth. No new dependencies.

**Performance:** No concerns. The `try/catch` adds no work on the success path; no new loops, queries, or allocations. (Aside: with Redis down, every request resolves to `null` and every client is pushed to re-auth, but the OAuth path also needs Redis — this is the "honest cost" the PR already documents, and 401 remains strictly better than a 500.)

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the two items above are non-blocking. Nicely reasoned change with the rationale captured in-code.

jwfing (commented):
## Review: A stale token gets a challenge, not a stack trace

**Summary:** A tight, correct fix — wrapping `resolveProjectFromToken` in a try/catch that returns `null` converts an escaping `MaxRetriesPerRequestError` (raw 500) into the existing 401 + `WWW-Authenticate` path, which is exactly what lets an MCP client re-run OAuth. No blocking issues.

**Requirements context:** No matching spec/plan found — `docs/superpowers/` and `docs/specs/` are both empty in this repo. Assessed against the PR description, the surrounding code, and the referenced token-format changes (#82, #87/#89, #94).

I verified the claimed behavior end-to-end:
- `await` is correctly placed *inside* the `try` (`src/http/server.ts:148`); without it the rejection would escape the catch. This is the easy-to-get-wrong detail and it's right.
- Both call sites — `POST /mcp` (`server.ts:946`) and `GET /sse` (`server.ts:1133`) — funnel a `null` result into `sendUnauthorized(...)`, which sets `WWW-Authenticate` and returns 401 (`src/http/auth-challenge.ts:76-77`). So `null` genuinely produces the challenge.
- The wrapper is the *only* path to `oauthManager.resolveProjectFromToken` (grep confirms callers are just the two routes via the wrapper), so the PR's "a future caller cannot reintroduce the escape" claim holds.
- No regression to the legacy path: a stale OAuth token now falls through to legacy header handling instead of 500ing, consistent with the existing "both provided → prefer legacy" logic (`server.ts:972-995`).

### Critical
(none)

### Suggestion

**Software engineering — no regression test for the new behavior** (`src/http/server.ts:147-172`). The body cites "215 tests," but none exercise this path: there is no `server.test.ts`, and no test references `resolveProjectFromToken`, `project_binding_required`, or the catch (`auth-challenge.test.ts` only covers the challenge *builder*, not the route). This is precisely the kind of close-call bug (missing `await`, wrong error-to-status mapping) a test would pin. Since `resolveProjectFromToken` is a private, non-exported function and the route layer is currently untested, a focused test would need either a small app/supertest harness or exporting the wrapper — a real but bounded lift. Non-blocking given the routes have never had coverage, but worth adding so the "throw → null → 401" contract can't silently regress.

**Functionality — stale/garbage tokens are labeled `project_binding_required`** (`server.ts:957-963`, `1143-1147`). Because the catch collapses "infra fault", "stale token", and "unknown/unbound token" all to `null`, a stale bearer now takes the `oauthToken && !legacyApiBaseUrl` branch and receives `error: "project_binding_required"` with `"OAuth token is valid but not bound to a project… call POST /api/projects/{projectId}/bind"`. The 401 + challenge is delivered (so a spec-compliant client recovers via re-auth — the PR's core goal is met), but the JSON body mislabels an invalid token as valid, and a client that reads the body could be nudged toward `/bind`, which would also fail with the stale token. RFC 6750 would suggest `invalid_token` here. Non-blocking — the recovery action is identical — but the message accuracy is a small wart.

### Information

- **Security — no concerns.** The catch logs only the `error` object (an ioredis error), never the token, and the token is fingerprinted elsewhere (`tokenFingerprint`). The client receives only the generic challenge — no stack, no secret, no PII. Auth is strengthened, not weakened.
- **Performance — no concerns.** A try/catch wrapper on an already-async call; no new queries, allocations, or hot-path work.
- **Honest-cost note (already acknowledged in the code comment):** during a *full* Redis outage this returns 401 to every client including valid ones, and the OAuth re-auth flow also needs Redis (`createAuthorizationState`), so recovery still can't complete until Redis is back. That's no worse than the prior 500, and the comment states the trade explicitly — flagging only for completeness, not as a change request.

**Verdict:** `approved` (informational; a human still gives the explicit GitHub approval). No Critical findings — the fix is correct, minimal, and well-reasoned. The Suggestions (a regression test, and an `invalid_token` label for stale tokens) would round it out but are not blocking.

jwfing (approved):
LGTM - approved.

PR #96, opened 2026-07-31
Sessions live in this process, and redis.ts is deleted

tonychang04:
The last Redis use. Stacked on #94 — **merge that first**; this branch targets it rather than master so the diff is only the session change.

## Why a session could not become a value

Registrations, auth state, the platform token, auth codes and the token binding all became sealed or signed values. A session cannot, and it is worth being exact about why rather than calling the server stateless at the end of this:

a session owns an `McpServer` and a **live transport** — for a streaming client, an open TCP connection held by one process. That is not information, so no amount of cryptography moves it to another machine.

Redis was never making sessions stateless either. It stored a copy of the data *beside* the connection, and `restoreSession` rebuilt a new server and transport around a **reused session id**. That works only for a client that reconnects with a POST — which is precisely the client that could have re-initialized instead. What it bought:

| | |
|---|---|
| a session id surviving a restart | real, but only for POSTing clients |
| sharing sessions across instances | never used: one instance, and a transport is not shareable |
| an expiry clock | replaced here by the idle sweep |

against a hard dependency in the request path of every tool call.

## The part a client actually depends on is the status code

A session id we do not hold now answers **404**, not 400.

That is the one thing in here a client's recovery is keyed on: a client receiving 404 for a request carrying `Mcp-Session-Id` MUST start a new session (Streamable HTTP 2025-03-26), and the SDK's own server does the same — *"Requests with invalid session IDs are rejected with 404 Not Found"*. The 400 this used to fall through to is the same class of mistake as the 500 in #95: not the code recovery is keyed on, so a restart leaves the client stuck instead of reconnecting.

The branch order is extracted into `routeForSessionRequest` and pinned by tests, because the case that decides it is easy to get wrong: **an initialize that still carries a stale session id** is exactly what a reconnecting client sends, and routing that to 404 answers "not found" to the request that repairs the situation.

## Measured against a running server, no Redis anywhere

```
initialize (legacy creds)          200 + Mcp-Session-Id
reuse that id                      200, tools/list served
restart, reuse that id             404 + what to do next
initialize carrying the dead id    200 + a new Mcp-Session-Id
unknown id on GET / DELETE         404
no id, not an initialize           400   (unchanged: nothing to recover)
```

## The sweep loses logic rather than gaining it

The `exists` term is deleted, not replaced. Every subtlety it carried came from **two clocks that could disagree** — Redis wiping records early made every resident session read as an orphan at once — and one clock cannot disagree with itself. What remains is the condition that was doing the real work: nobody has talked to this session for the TTL.

The open-stream guard stays and is **not** redundant: a client holding `GET /mcp` open sends no requests, so nothing stamps its clock while it is plainly connected.

Two things that follow from there:

- `startIdleSweep` is now **unconditional**. Gated on Redis it was correct — Redis owned the lifetime and this only reclaimed instances left behind. Now it *is* session expiry, so a server that skips it leaks every session it creates: ~252 kB each (Iris's measurement) against a ~493 MB heap limit, so roughly 2,000 before the process dies.
- `deleteSession` drops the entry **outside** the try. A close that throws must not strand a session, because the sweep only reaps what it can close — that entry would never be collected again.

## verify-deploy: a check that cannot fail is worse than no check

Both `/health` session counts now come from one store, so the resident-vs-active ratio can never disagree and that check can never fail — while still reading as coverage in a green run. It is replaced by heap percentage, which catches a stalled sweep, a leak from any other cause, and plain undersizing.

`ioredis` is dropped from `package.json`. `REDIS_URL`/`REDIS_HOST` are now read by nothing; the note in `config.ts` says so, because the danger is assuming a Redis is there because the variable is.

## The cost, stated rather than discovered

**A deploy drops every live session.** Clients are told to re-initialize, and the server now says so in the way the protocol defines — but *nobody has watched a real editor take this 404 and reconnect*. That is the acceptance for this change rather than a detail after it: @qa-bot's real-client run is what closes it.

Also true and not fixed here (Iris's framing, which is right): this makes us **Redis-free, not stateless**. One machine only; a second replica needs sticky sessions or the mcp-use port. That is the ticket this thread was opened for, and it is separate work.

217 tests pass. `redis.ts` and `redis.test.ts` are deleted.


jwfing (commented):
## Review: Sessions live in-process, `redis.ts` deleted

**Summary:** Removes the last Redis dependency by making the in-memory session map the sole store, replaces `restoreSession` with a spec-correct 404 for unknown session ids, and makes the idle sweep unconditional so it becomes the real expiry mechanism — a well-reasoned, well-tested refactor with no blocking issues.

### Requirements context
No `/docs/superpowers/` (or any `docs/`) directory exists in this repo — **no matching spec/plan found, so I assessed against the PR description, the stacked-branch context (targets `feat/self-describing-access-token`, #94), and the MCP Streamable HTTP contract.** I verified the PR's central claim — that an unknown/stale `Mcp-Session-Id` must be answered with **404**, not 400 — against the upstream `@modelcontextprotocol/sdk` docs via context7: the SDK's own `validateSession` returns `404 / -32001 "Session not found"` for a session id it does not hold and `400 / -32000` for a missing header on a non-initialize request. `routeForSessionRequest` (`src/http/session-manager.ts:192-201`) implements exactly that contract, and the branch order (initialize-with-stale-id → `create`, not `not-found`) matches the reconnect path. The reasoning is accurate, not hallucinated.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — the one behavior a client depends on has only unit coverage.** `routeForSessionRequest` is thoroughly pinned in `session-manager.test.ts:63-108`, but the actual Express wiring at `src/http/server.ts:1002-1030` (POST `/mcp` returning the 404 body) and `server.ts:1142-1149` / `1176-1182` (GET/DELETE) has no request-level test. The PR itself is candid that the end-to-end reconnect is unverified ("*nobody has watched a real editor take this 404 and re-initialize*", `server.ts:1020-1023`) and defers that to a real-client run. Given the repo has no supertest/integration harness for the app today, this stays a Suggestion rather than a blocker — but a thin supertest over the three handlers, or explicit tracking of the qa-bot acceptance as the merge gate, would close the gap the unit test can't reach.

#### Information
- **Software engineering — newly-dead public methods.** After this change `getStreamableSession` is the only accessor server.ts uses; `getSSESession` (`session-manager.ts:308-314`) and `getRuntimeSession` (`session-manager.ts:288-290`) have no remaining call sites in the repo (the SSE path reads its own `sseTransports` map in `server.ts:1199+`). Consider removing them, or leave a note if they're kept as intentional API surface.
- **Software engineering — redundant double-stamp on GET `/mcp`.** `openStream(sessionId)` already sets `lastSeenAt` (`session-manager.ts:389-394`), and the handler calls `touchSession(sessionId)` immediately after (`server.ts:1154-1156`). Harmless, just two writes where one would do.
- **Functionality — stale operator hint in verify-deploy.** `scripts/verify-deploy.mjs:341` still explains a 500 at `/oauth/authorize` as "*the flow still needs backing state it cannot reach (remaining Redis uses)*". With this stack removing Redis entirely, that parenthetical is now misleading to whoever reads a failing deploy gate; worth updating to the current cause.

#### Security
No security-relevant regressions. The change removes an external dependency (net attack-surface reduction), adds no new user input to any sink, logs no secrets (token logging still goes through `tokenFingerprint`), and does not touch the auth/authorization path — session creation still resolves and validates the OAuth token before a session exists. The 404 body leaks nothing sensitive.

#### Performance
Net improvement. `getSessionData` becomes a synchronous map lookup instead of a Redis round-trip on the request path of every tool call; the sweep is O(n) over resident sessions every 5 min (default) and filtered to streamable transports; `heapStats()` on `/health` is two cheap V8 calls. No new N+1, blocking I/O, or per-request allocation. The unconditional `startIdleSweep` (`server.ts:1384`) is correctly now the only thing bounding heap — the accompanying `/health` heap-percentage check in `verify-deploy.mjs:204-209` is the right replacement for the now-meaningless resident-vs-active ratio.

### Verdict
**approved** (informational — no Critical findings; posted as a comment, human approval via the normal flow). The Redis removal is coherent end to end: the lockfile and `package.json` drop `ioredis` consistently, `SESSION_CONFIG`/`redis.ts`/`redis.test.ts` are cleanly removed with no dangling imports, and the 404 routing is verified against the upstream SDK contract. The two things worth doing before/after merge are the operational acceptance (a real client taking the 404 and reconnecting) and the minor cleanups above.

jwfing (commented):
## Code review — Sessions in-process, `redis.ts` deleted

**Summary:** Deletes the last Redis dependency by holding MCP sessions entirely in-process, replacing the Redis TTL with an unconditional idle sweep and turning an unknown session id into a spec-correct 404; the change is well-scoped, thoroughly unit-tested, and I found no blocking issues.

### Requirements context
This repo has no `/docs/superpowers/` or `/docs/specs/` directory (confirmed — MCP server repo, gated by vitest + tsc + eslint + tsup only). No matching spec/plan found — assessed against the PR description, the linked stacking context (#94/#95), and the Streamable HTTP 2025-03-26 protocol semantics the PR cites. Note this PR targets `feat/self-describing-access-token` (#94), not `master`; the diff was reviewed against that base so only the session change is in scope.

---

### Critical
(none)

### Suggestion

**Functionality / testing — the 404→reconnect path is not verified with a real client.** `src/http/server.ts:1002-1030` returns 404 for an unheld session id, which is the correct protocol answer and is unit-pinned by `routeForSessionRequest` tests (`session-manager.test.ts:70-107`). But as the PR body and the inline comment (`server.ts:1020-1023`) both state, *no real editor has been observed taking this 404 and re-initializing*. Since every deploy now drops all live sessions and the entire recovery story rides on clients honoring 404, I'd treat the pending `@qa-bot` real-client run as a merge/deploy gate rather than a follow-up. Non-blocking because the routing is spec-correct and the SDK server behaves identically.

**Software engineering — routing is tested as a pure function, not at the route layer.** The branch contract is well covered at the `routeForSessionRequest` unit level, but the *wiring* in `POST /mcp` is untested — e.g. that a session id belonging to an SSE session makes `getStreamableSession()` return null (`session-manager.ts:296-302`) and therefore falls through to `create`/`not-found` rather than `use-existing`. A thin supertest-style integration test over the three status outcomes (200 reuse / 404 unknown / 200 re-init-with-stale-id) would guard the glue that the pure-function tests can't reach.

### Information

- **`tsup.config.ts:13` still lists `'ioredis'` in `external`.** Now that the dependency is dropped from `package.json` nothing imports it, so the entry is inert — but it's stale and worth removing while the Redis cleanup is fresh.
- **verify-deploy heap check is a point-in-time smoke test.** `scripts/verify-deploy.mjs:200-208` grades `heapUsedPct < 80` right after deploy, when heap is low — so it won't catch a *slow* session leak at deploy time; its real value is continuous monitoring, which the PR acknowledges ("that number belongs in whatever ends up watching this endpoint", `session-manager.ts:571-572`). Still a strict improvement over the old resident-vs-active ratio, which the PR correctly identifies as a check that can no longer fail. The `/health` `memory` block it reads is really emitted (`server.ts:208-216, 235`) and uses correct V8 API fields (`used_heap_size`, `heap_size_limit`), so there is no dead-else silent-pass risk here.
- **`scripts/verify-deploy.mjs:341`** still references "remaining Redis uses" in an error hint (line not modified by this PR); phrasing is slightly stale now that Redis is fully gone.

---

### Dimension coverage

- **Software engineering** — Strong. 217 tests pass; `session-manager.test.ts` covers the routing contract, the TTL boundary, the open-stream guard, overlapping-stream reconnects, the touch-inside-sweep re-check, close-that-throws still dropping the entry, and sweep-timer idempotency/survival. Conventions (ESM `.js` imports, error handling, monotonic-vs-wallclock clock split) are consistent with the surrounding code. Deleted `redis.test.ts` was appropriately removed with its subject.
- **Functionality** — Solid. `startIdleSweep` is correctly made unconditional (`server.ts:1384`), which is the load-bearing fix: gated on Redis it would now leak every session. `deleteSession` drops the map entry *outside* the try (`session-manager.ts:449`) so a throwing close can't strand a session past the sweep. The SSE sweep exclusion (`session-manager.ts:465-466`) and the "initialize carrying a stale id → create, not 404" ordering (`session-manager.ts:197-200`) are both correct and tested. No dangling references to the removed Redis surface remain (grepped `redis`/`ioredis`/`SESSION_CONFIG`/`restoreSession` — only comments left).
- **Security** — No new concerns. No SQL/shell/user-input surface added; the Redis removal drops `REDIS_URL`/`REDIS_HOST`/`REDIS_PASSWORD` handling (config note is accurate). Session reuse keyed on the session id is unchanged pre-existing MCP design, not weakened here. Tokens are still fingerprinted rather than logged; session ids logged in full is pre-existing.
- **Performance** — Net positive. `getSessionData`/`hasSession` become synchronous Map lookups instead of a Redis round trip on the request path of every tool call. The sweep is O(n) every 5 min, bounded; `heapStats()` on `/health` is cheap. No N+1, unbounded loops, or new blocking I/O on the event loop.

### Verdict
**approved** (informational — zero Critical findings; the two Suggestions are non-blocking and the author already flags the real-client verification as the acceptance gate). A human still gives the explicit GitHub approval via the approve flow.

jwfing (approved):
LGTM - approved.

PR #97, opened 2026-07-31
Revoke actually revokes, and a platform blip is not a dead sign-in

tonychang04:
Four findings — the review round on #94, plus Iris's two — all the same shape: **an answer that was easy to produce rather than true.** Stacked on #96; merge order is #94 → #96 → this.

## 1. `/oauth/revoke` did nothing, and returned 200 while doing it

With the binding row gone it only dropped the cached project key. The sealed bearer still carried a live platform token, so the very next request re-fetched and succeeded — revoking a leaked credential bought one extra round trip and left it usable for a full 24 hours. **That is worse than having no revoke at all**, because the person who called it stops looking for the leak.

It now unseals the bearer and revokes the platform token upstream. Iris confirmed the endpoint is deployed and needs only our client id — which is what makes this the option that *doesn't undo the evening's work*. The alternative (a local revocation marker) is a durable per-token record consulted on every request: `mcp:auth:binding:` under another name.

Measured against a stub platform:

```
our token, platform healthy    200, upstream received the platform token
not our token                  200, upstream NOT called      (RFC 7009 §2.2)
upstream unreachable           503 + Retry-After
upstream refuses our client    500, named as our misconfiguration
```

RFC 7009's blanket 200 is about not leaking which tokens exist. It is **not** licence to report a failed revocation as success — that's the same false assurance this endpoint just stopped giving, one layer down. The cost is written next to the code: a failure distinguishes "was ours" from "was not", which only happens when the platform is already failing and isn't attacker-triggerable.

## 2. A transient platform error was reported as a revoked grant

Any rejection from `getProjectAccess` became *"your sign-in is no longer valid"*. A platform blip threw away a working session and sent a person through a browser login to fix something that would have cleared on retry. The old 30-day binding never did this; making the lookup per-request is what introduced it.

**This is my own #95 trade, corrected.** I chose 401-for-everything because the client's available action was identical either way — and Iris is right that *it stops being identical the moment a retry would have worked*. Her rule, adopted verbatim: **401 when the credential is the problem, 5xx when we could not tell.**

```
platform 403 (grant gone)    401 + WWW-Authenticate
platform 500 (cannot tell)   503 + Retry-After, NO challenge
token does not open          401 + WWW-Authenticate
```

`429` is deliberately *not* a refusal despite sitting in the 4xx range — it's the platform declining to answer right now, which is exactly the retry case, and a careless `>= 400 && < 500` test would swallow it. An unrecognised failure defaults to retryable: the expensive mistake is the one that signs a user out.

## 3. `expires_in` advertised 24h whatever the platform token said

The PR body claimed "24h or the platform token's expiry, whichever is sooner"; the code returned a flat 24h. When the inner token dies first the client believes it has hours left and keeps retrying a credential we already know is dead.

Now the real minimum, read from the platform JWT's `exp`. Deliberately **unverified** — it isn't ours to verify, and a forged `exp` can only *shorten* what we advertise, which is the harmless direction. Anything unparseable falls back to our own TTL.

## 4. The branches where 1–3 lived had no tests

They do now, including the classifier's boundaries (401/403 refusal; 429, 5xx, network, unknown → not).

232 tests pass.

---
## Summary by cubic
Make revoke actually revoke the platform token and stop logging users out on transient platform errors. Adds a 10s timeout to all platform calls, fixes the OAuth revoke path, corrects `expires_in`, and updates error mapping to clearly signal temporary outages.

- Bug Fixes
  - `/oauth/revoke` now unseals the bearer and revokes the platform token upstream; still 200 for unknown tokens (RFC 7009), 503 with Retry-After when unreachable, and 500 when our client is refused. Local key cache is dropped unconditionally. Fixed path to `/api/oauth/v1/revoke`.
  - Error handling: 401/403 return 401 with WWW-Authenticate; 5xx/network/timeout map to 503 with Retry-After and no challenge; 429 passes through as 429. The status mapper gained an explicit 503 branch. The OAuth callback now follows the same rules; `invalid_grant` stays 400 and no longer leaks raw “fetch failed”.
  - Bound all platform calls to 10s via a shared wrapper and moved the token exchange into it.
  - `expires_in` now reports the minimum of 24h and the platform JWT `exp`; falls back to our TTL if unreadable and never goes negative.
  - Merged the `/api/projects` bearer unseal and added tests for revoke, lifetime calculation, and status mapping.

<sup>Written for commit 48c0ce5a859d22adfe9a333f81cd5ddfa0c153ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/97?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jwfing (commented):
## Review — #97 "Revoke actually revokes, and a platform blip is not a dead sign-in"

**Summary:** A careful, well-documented correctness/security fix that makes `/oauth/revoke` actually revoke upstream, splits "credential refused" from "platform unreachable" (401 vs 503), and advertises the true `expires_in` — all three are real improvements and the diff is tightly scoped. No blocking issues found.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory, so I assessed against the PR description, the four stated findings, and the surrounding code. This is a stacked PR (base `feat/sessions-in-memory`, merge order #94 → #96 → this); I reviewed only the delta this PR introduces.

I verified the touched API surface against the workspace at the head SHA:
- `AccessTokenPayload.platformAccessToken` exists (`access-token.ts:77`) — the JWT the lifetime/revoke logic reads.
- `INSFORGE_CONFIG.clientId` (`config.ts:78`), `INSFORGE_API_BASE`, `InsforgeApiError(msg, statusCode)` all present and used correctly.
- `getProjectAccess` → `getProject`/`getProjectApiKey` both `throw new InsforgeApiError(text, response.status)`, so a platform **403 → `isAuthorizationRefusal` true** (challenge) and a **5xx / node-fetch network error → false** (retryable). The classifier's branch table matches reality.
- The new `throw` in `oauth-manager.resolveProjectFromToken` (`oauth-manager.ts:407`) is reached only through `resolveProject` (`server.ts:194`), whose `try/catch` maps it to `outcome:'unavailable'`. Both call sites (`server.ts:1155` mcp-create, `server.ts:1341` sse) handle `unavailable` before falling through to the challenge. **No unwrapped throw escapes to Express** — the wrapper's invariant is preserved.

---

### Findings

#### Critical
(none)

#### Suggestion
- **[software-engineering / test coverage]** The new *unit* helpers are well tested (`accessTokenLifetimeSeconds` boundaries in `access-token.test.ts`; the 401/403-vs-429/5xx/network classifier in `error-status.test.ts`). What has **no test** is the endpoint wiring that consumes them: the `/oauth/revoke` status mapping (`server.ts:955-985` — 200 on success, 500 on refusal/misconfig, 503+Retry-After on unavailable) and the `resolveProject` → `sendUnavailable` 503 path (`server.ts:1157`, `1343`). A regression that swapped the 500/503 mapping, or dropped the `unavailable` early-return so a platform blip fell through to the challenge, would pass the current suite. I note the repo has no existing `server.test.ts`/supertest harness, so this isn't an established pattern — but given the whole point of the PR is *which status a failure becomes*, a handler-level test (with `revokePlatformToken` / `getProjectAccess` mocked) would guard the exact behavior the PR is correcting.
- **[functionality / security]** `revokePlatformToken` failure classification (`server.ts:965-976`) maps **any upstream 401/403 → 500 "server misconfiguration"**. That's right for `invalid_client`, but RFC 7009 §2.2 recommends the revoke endpoint answer **200 even for an unknown/expired *token***. If the platform ever returns 401/403 for a token reason (rather than a client-auth reason), a benign revoke surfaces to the caller as an alarming "this is a server misconfiguration, not something retrying will fix." This rests entirely on the upstream contract Iris confirmed (`200 {success:true}` for our client, `401 invalid_client` only for a bad client_id). Worth a one-line guard or comment pinning that assumption so a future platform change doesn't silently mislabel token-level rejections.

#### Information
- **[security]** On upstream failure the revoke endpoint distinguishes "was ours" (500/503, because `!payload` early-returns 200 at `server.ts:930`) from "was not ours" (200) — a response oracle the blanket-200 rule exists to hide. The PR documents this explicitly: it only manifests when the platform is already failing and is not attacker-triggerable. I agree the trade is acceptable; recording it for the reviewer.
- **[functionality]** `accessTokenLifetimeSeconds` can return `0` when the platform token is already expired at issue time (floored, and tested at `access-token.test.ts`). Harmless here since tokens are minted right after login with a fresh platform token; a note only because a strict OAuth client could dislike a non-positive `expires_in`.
- **[security]** Reading the platform JWT's `exp` **unverified** (`access-token.ts:jwtExpirySeconds`) is safe as written: `Math.max(0, Math.min(ACCESS_TOKEN_TTL_SECONDS, remaining))` caps at our own TTL, so a forged `exp` can only *shorten* what we advertise. Good defensive framing; no action.
- **[performance]** `server.ts` seals the access token then immediately `readAccessToken`s it back to compute `expires_in` (token endpoint). One extra decrypt on the login path only — negligible; flagging only as a micro-note, not worth changing.

---

### Verdict
**`approved`** (informational — the GitHub green check remains a separate human action). Zero Critical findings; the two Suggestions are non-blocking. This is exactly the kind of "make the answer *true*, not just easy" fix the PR set out to be, and the reasoning is unusually well captured in the comments. Posted as a COMMENT per the ≥1-Critical rule.

jwfing (commented):
## Review: "Revoke actually revokes, and a platform blip is not a dead sign-in"

**Summary:** A well-reasoned correctness/security fix that makes `/oauth/revoke` revoke the sealed platform token upstream, splits transient-platform-failure from credential-refusal (401 vs 503), and advertises the real `expires_in` — I found no blocking issues.

### Requirements context
No `docs/superpowers/` (or other spec/plan dir) exists in `insforge-mcp` — assessing against the PR description, its four numbered findings, and the surrounding code alone. This is a stacked PR (base `feat/sessions-in-memory`, merge order #94 → #96 → this); review is scoped to this diff.

The four items are internally consistent with the code: `#1` upstream revoke (`insforge-api.ts:211`, `server.ts:944-984`), `#2` refusal-vs-unavailable classifier (`error-status.ts:28-32`, `oauth-manager.ts:377-408`, `server.ts:182-224`), `#3` real lifetime (`access-token.ts:135-165`, `server.ts:855-858`), `#4` tests for the pure logic.

---

### Critical
(none) — no correctness, security, data-loss, or breaking-behavior defects found. The unverified-`exp` read only ever *shortens* the advertised lifetime (`Math.min` against the ceiling in `access-token.ts:145`), so it cannot be abused to extend a session; the throw newly introduced in `oauth-manager.resolveProjectFromToken` has exactly one caller (`server.ts:197`, the `resolveProject` wrapper) which converts it to `unavailable`, so nothing leaks an unhandled 500.

### Suggestion

**Software engineering — the headline feature (#1) has no automated coverage.**
`revokePlatformToken` (`src/http/insforge-api.ts:211-232`) and the revoke endpoint's branch mapping — 200 on success, 500 on `invalid_client`, 503 on unreachable (`src/http/server.ts:944-984`) — are untested. This is *consistent* with repo conventions (`insforge-api.ts` has no test file and there are no endpoint/supertest tests), and the reused decision logic (`isAuthorizationRefusal`) is well covered in `error-status.test.ts`. But the entire security value of #1 rests on that upstream `fetch` actually firing with the right shape. A mocked-`fetch` unit test for `revokePlatformToken` (2xx → resolves; non-2xx → throws `InsforgeApiError` carrying `response.status`) would lock the contract cheaply.

**Functionality — the revoke wire format is asserted nowhere.**
`revokePlatformToken` posts `application/json` with `{ token, token_type_hint, client_id }` and no client secret (`src/http/insforge-api.ts:212-220`). RFC 7009 §2.1 specifies `application/x-www-form-urlencoded`; you note Iris confirmed the InsForge endpoint accepts this JSON shape, which is fine *if that contract holds* — but if the platform ever rejects the body, the feature degrades silently to "revoke always 500/503, nothing actually revoked," which is precisely the false-assurance failure #1 set out to kill. Worth one integration check against the live endpoint before shipping, since no test pins the format.

**Functionality — `unavailable` bypasses a present legacy API key (low blast radius).**
In the mcp `create` path (`src/http/server.ts:1155-1158`) and the SSE path (`server.ts:1341-1342`), an `unavailable` outcome early-returns 503 even when the same request also carries a legacy API key that needs no platform round-trip. Previously a transient failure fell through and the legacy key could still serve the request. Narrow (requires dual credentials *and* a platform blip), but if the legacy key is meant to be a hard fallback, gate the early return on `!legacyApiKey`. `unauthorized` already falls through correctly, so only the `unavailable` branch has this.

### Information

- **`error-status.ts:3-8`** — the original `statusForHttpError` doc comment ("Preserve client-error statuses… Treat upstream 5xx…") is now stranded above the newly-inserted `isAuthorizationRefusal`, leaving `statusForHttpError` (line 34) undocumented and giving `isAuthorizationRefusal` two comment blocks. Move that comment back down to its function.
- **`server.ts:976`** — a platform `400` from the revoke endpoint maps to `503 temporarily_unavailable`, inviting a retry that won't help. Only 401/403 are special-cased; other 4xx aren't really "temporary." Extremely unlikely (we control the request), hence informational.
- **`server.ts:855`** — `expires_in` is computed by re-decrypting the token just issued two lines earlier. Having `exchangeCode` return the payload it already built would avoid the second decrypt. Cosmetic.

### Dimension notes
- **Security:** no auth/authorization weakened; the unverified JWT `exp` can only shorten the advertised lifetime; the platform token is not logged (only the platform's error *text* is, on failure). Refusing to return a blanket 200 on a failed revoke is the correct RFC-7009 reading. No new dependencies.
- **Performance:** no DB in this PR, so no index/N+1 concerns. The added cost is one `fetch` on revoke (rare) and one cheap JWT parse per token issue. The per-request platform lookup pre-dates this PR (#95/#96).

### Verdict
**approved** (informational — human still gives the GitHub green check via the approve flow). Zero Critical findings; the Suggestions above are worth a look, especially adding a `revokePlatformToken` unit test and confirming the revoke wire format against the live platform, but none block merge.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: "Revoke actually revokes, and a platform blip is not a dead sign-in"

**Summary:** A well-reasoned set of four correctness fixes to the OAuth revoke path, upstream error classification, `expires_in`, and platform-call timeouts; the logic is sound and the pure functions are tested, but the HTTP endpoints where the bugs actually lived still have no automated coverage.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/`, `/docs/specs/`, or `.claude/skills/` directory. Assessed against the PR title/body, the referenced findings (#94/#95/#96), and repo conventions (TDD, vitest+tsc+eslint gates). A scope note: this PR is stacked on #96, which is not yet merged into `master`, so the diff-vs-`master` conflates the two. Commit `707fc08` ("Sessions live in this process, and redis.ts is deleted") is #96's session/Redis rewrite (`session-manager.ts`, `redis.ts`, `config.ts`, `verify-deploy.mjs`, `package.json`); the revoke/error/lifetime/timeout work in commits `0fe346f`, `83592de`, `40289b8`, `48c0ce5` is this PR's own. I reviewed the whole diff but weighed blocking judgment against this PR's commits, and treat the session rewrite as belonging to #96's review.

---

### Findings

#### Critical
(none)

#### Suggestion

**1. Software engineering — the endpoints this PR fixes have no automated test (`src/http/insforge-api.ts:269-287`, `src/http/server.ts:940-1021`, `src/http/server.ts:1048-1056`, `src/http/server.ts:661-698`).**
The pure classifiers are covered well: `error-status.test.ts` pins the 401/403-vs-5xx/429/network boundaries, and `access-token.test.ts` pins the lifetime math. But the bugs this PR fixes live one layer up, at the route/wiring level, and that layer is untested:
- `revokePlatformToken` and the `/oauth/revoke` route's own branch logic (unknown-token→200, refusal→500, unreachable→503+Retry-After) — the four outcomes in the PR body read as manual measurements against a stub, not committed tests. The PR body's "Adds tests for revoke" maps to lifetime + status-mapping tests; the revoke *endpoint* itself is not exercised.
- The `/api/projects` unseal fix (`server.ts:1048`) — the description candidly says it "was missed because nothing calls this endpoint in tests," and that gap is unchanged by this PR.
- The callback's thrown-vs-parsed-error split (`server.ts:661-698`) that turns an unroutable platform into 503 instead of a browser-visible `fetch failed`.

The repo has no `server.test.ts`/supertest harness, so I understand why this is nontrivial, and the code reads as correct — hence Suggestion, not blocking. But given the repo's TDD convention and that three of these were "answers that were easy to produce rather than true," a regression test at the route level (e.g. injecting a stub `revokePlatformToken`/`exchangePlatformCode`) is exactly what would keep the next such regression from shipping green.

**2. Security — `console.error('[OAuth] Token exchange error:', tokens)` logs the full platform response (`src/http/server.ts:701`).**
This branch is entered on `tokens.error || !tokens.access_token`. In the ordinary case `access_token` is absent, so nothing sensitive is logged. But if the platform ever returns an `error` *alongside* an `access_token`, the whole object — including that platform credential — lands in logs. Low likelihood, but the surrounding code is otherwise scrupulous about never logging the platform token (`access-token.ts:174-181` refuses to print it by length alone), so logging a specific field or a redacted shape would keep that discipline.

#### Information

- **Functionality / performance:** No security regressions or new N+1s — removing the per-request Redis round trip (`getSessionData` is now synchronous) is a net improvement, and the 10s `AbortSignal.timeout` wrapper (`insforge-api.ts:125-127`) correctly makes the 503 path reachable where undici's 300s default headers-timeout previously hid it. The `expires_in` ceiling is measured from the token endpoint while the seal's 24h starts at code issue, so it can over-advertise by at most the 5-minute code TTL when the platform token outlives 24h — immaterial.
- **Design trade (from #96, riding along):** Dropping Redis means a restart drops every live session and the server is single-instance-only. This is documented exhaustively (`session-manager.ts:9-48`) and deliberate; flagging it only so it's on the record for #96's review, not as a defect here. The `verify-deploy.mjs` change from a session-ratio check to a heap-percentage check is the right call given the ratio can no longer diverge.
- **Doc nit:** The comment at `insforge-api.ts:262` writes `POST /oauth/v1/revoke` while the actual (and correct) call is `/api/oauth/v1/revoke` — mildly confusing given the surrounding comment's whole point is that the `/api` prefix is what avoids a 404.

---

Zero Critical findings. The core logic — revoke now revokes upstream, transient platform errors map to 503+Retry-After instead of a spurious 401, `expires_in` reports the real minimum, and every platform call is bounded — is correct and carefully justified. The two Suggestions (route-level test coverage for the fixed endpoints; tightening the token-exchange error log) are worth addressing but non-blocking.

jwfing (approved):
LGTM - approved.

PR #98, opened 2026-07-31
Session ids are credentials, so stop writing them into the logs

tonychang04:
Stacked on `fix/revoke-actually-revokes` so #96 does not need a re-review.

## Why

An existing session is served without a token check, so `Mcp-Session-Id` alone yields the stored project key and full tool access. Measured, not argued — one session created *with* credentials, then reused with none:

```
POST /mcp  tools/list   session id ONLY  -> 200, full tool list
POST /mcp  tools/call   session id ONLY  -> 200
     and the backend received:  x-api-key = <the victim's key>
```

We already fingerprint the OAuth token. We printed the session id in the clear beside it, in the same string. Iris pulled live session ids out of Manufact's runtime logs using an API key, which is what makes this a credential at rest rather than a style point.

## What

21 sites, not the six that a grep for `` `Session: ${sessionId}` `` finds. A session id reaches a log through three expression shapes across two files:

| shape | where | visible to a grep for the first shape? |
|---|---|---|
| `` `${sessionId}` `` / `` `${sessionId \|\| 'none'}` `` | server.ts | yes |
| `console.log('...', sessionId)` | server.ts, incl. the new 404 branch | **no** |
| `` `${transport.sessionId}` `` | the SSE cluster | **no** |

`sessionFingerprint()` lives beside the session code and is the same construction as `tokenFingerprint` — two equivalent credentials, one treatment. Deterministic, so the same session prints the same value at every site and a support report still correlates.

## Verification

Runtime, on **both** transports, because a grep only covers the paths it exercises:

```
streamable: initialize -> tools/list -> GET stream -> dead-id probe (404 branch) -> DELETE
SSE:        GET /sse -> POST /messages -> close

grep for any 36-char UUID in the whole server log:  0 matches
what the log carries instead:  Session created: b37d69ec
                               Session initialized: b37d69ec
                               Session deleted: b37d69ec
                               [SSE] Session created: 4465d50d
```

The test reads the **source**, not the behaviour, for exactly that reason: the SSE sites are invisible to a runtime grep unless someone opens an SSE session. It fails on an expression shape nobody has thought of yet, and carries a meta-test proving it is not vacuous. Both mutations confirmed failing:

```
reintroduce `${sessionId}`                 -> 1 test fails, names server.ts:1316
reintroduce `, newSessionId)` positionally -> 1 test fails
```

238 tests, tsc clean, lint clean.

## Scope

This stops us writing the credential down. It does **not** make a stolen session id useless — binding it to the session's `oauthTokenHash` is separate and still worth doing.

---
## Summary by cubic
Replace raw session IDs in all logs with an 8‑char SHA‑256 fingerprint to prevent credential leaks. Also add a CI-safe path check to catch bad platform routes before deploy.

- **Bug Fixes**
  - Added `sessionFingerprint()` (SHA‑256 → 8‑char hex, returns `none` when missing).
  - Replaced all session-id logging in `server.ts` and `session-manager.ts` (Streamable HTTP, SSE, errors, shutdown, 404) with the fingerprint, including positional args and `${transport.sessionId}`.
  - Added `session-fingerprint.test.ts` to scan source and fail on any console line that prints a raw session ID; includes a meta-test to prove the scan works.
  - Added `scripts/verify-platform-paths.mjs` and `npm run verify:paths` to assert platform routes exist (401/400/403 count as present; 404 fails).

<sup>Written for commit 6e7edca2268f670c1b14d0dee46ed14ffbd02327. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/98?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a command to verify that configured platform API routes are available and reachable.
* **Security**
  * Session identifiers are now masked in server and connection logs using short fingerprints instead of raw values.
* **Tests**
  * Added coverage to confirm fingerprint stability, uniqueness, fixed length, missing-value handling, and prevention of accidental identifier exposure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "test:coverage": "vitest run --coverage",
     "lint": "eslint .",
     "format": "prettier --write .",
-    "verify:deploy": "node scripts/verify-deploy.mjs"
+    "verify:deploy": "node scripts/verify-deploy.mjs",
+    "verify:paths": "node scripts/verify-platform-paths.mjs"
   },
   "keywords": [
     "mcp",
--- a/scripts/verify-platform-paths.mjs
+++ b/scripts/verify-platform-paths.mjs
@@ -1,0 +1,80 @@
+#!/usr/bin/env node
+/**
+ * Does every platform path we call actually exist?
+ *
+ * The revoke bug was a 404: the code called `/oauth/v1/revoke` where the
+ * platform serves `/api/oauth/v1/revoke`. It would have failed one hundred
+ * percent of the time in production, and nothing caught it — a stub answers
+ * whatever path the stub was written for, unit tests never leave the process,
+ * and the production probes that did run happened to exercise other routes.
+ * It was found by hand, while moving the call for an unrelated reason.
+ *
+ * So this asks the only question a stub cannot: is this path a route the
+ * platform knows? It does not need credentials and does not need a login,
+ * because the platform distinguishes the two cases by status:
+ *
+ *   401 / 400 / 403   the route exists and refused us. Correct.
+ *   404               the route does not exist. A typo that ships.
+ *
+ * Deliberately NOT a health check — it says nothing about whether the platform
+ * is well, only whether we are calling names it has. Nine seconds, no secrets,
+ * safe to run in CI on every push.
+ *
+ *   node scripts/verify-platform-paths.mjs [apiBase]
+ */
+
+const BASE = (process.argv[2] || process.env.INSFORGE_API_BASE || 'https://api.insforge.dev').replace(
+  /\/+$/,
+  ''
+);
+
+// Every platform path the server calls. A `:id` segment is filled with a
+// well-formed value that no tenant owns — the point is the route, not the row.
+const NOBODY = '00000000-0000-0000-0000-000000000000';
+const PATHS = [
+  { method: 'GET', path: '/auth/v1/profile', from: 'insforge-api.ts validateToken' },
+  { method: 'GET', path: '/organizations/v1', from: 'insforge-api.ts getAllUserProjects' },
+  { method: 'GET', path: `/organizations/v1/${NOBODY}/projects`, from: 'insforge-api.ts' },
+  { method: 'GET', path: `/projects/v1/${NOBODY}`, from: 'insforge-api.ts getProject' },
+  { method: 'GET', path: `/projects/v1/${NOBODY}/access-api-key`, from: 'insforge-api.ts' },
+  { method: 'GET', path: '/api/oauth/v1/authorize', from: 'server.ts /oauth/authorize' },
+  { method: 'POST', path: '/api/oauth/v1/token', from: 'insforge-api.ts exchangeCode' },
+  { method: 'POST', path: '/api/oauth/v1/revoke', from: 'insforge-api.ts revoke' },
+];
+
+const EXISTS = new Set([200, 201, 204, 302, 400, 401, 403, 405, 422]);
+
+const results = await Promise.all(
+  PATHS.map(async (p) => {
+    const started = Date.now();
+    try {
+      const res = await fetch(`${BASE}${p.path}`, {
+        method: p.method,
+        redirect: 'manual',
+        headers: p.method === 'POST' ? { 'content-type': 'application/json' } : {},
+        body: p.method === 'POST' ? '{}' : undefined,
+        signal: AbortSignal.timeout(10_000),
+      });
+      return { ...p, status: res.status, ms: Date.now() - started };
+    } catch (error) {
+      return { ...p, status: 0, error: String(error).slice(0, 60), ms: Date.now() - started };
+    }
+  })
+);
+
+let bad = 0;
+for (const r of results) {
+  const ok = EXISTS.has(r.status);
+  if (!ok) bad++;
+  console.log(
+    `${ok ? 'ok  ' : 'FAIL'}  ${String(r.status).padEnd(3)} ${r.method.padEnd(4)} ${r.path.padEnd(48)} ${r.from}${
+      r.error ? `  ${r.error}` : ''
+    }`
+  );
+}
+
+console.log(
+  `\n${results.length - bad}/${results.length} platform paths exist on ${BASE}` +
+    (bad ? `\n${bad} route(s) the platform does not serve — a request we send will always fail.` : '')
+);
+process.exit(bad ? 1 : 0);
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -8,7 +8,7 @@ import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/
 import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
 
 // Local imports
-import { getSessionManager, routeForSessionRequest } from './session-manager.js';
+import { getSessionManager, routeForSessionRequest, sessionFingerprint } from './session-manager.js';
 
 import { getOAuthManager } from './oauth-manager.js';
 import {
@@ -1087,7 +1087,7 @@ app.post(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
   const oauthToken = extractOAuthToken(req);
   const { apiKey: legacyApiKey, apiBaseUrl: legacyApiBaseUrl } = extractLegacyHeaders(req);
 
-  console.log(`[${new Date().toISOString()}] POST ${STREAMABLE_HTTP_ENDPOINTS.mcp} - Session: ${sessionId || 'none'}, Token: ${oauthToken ? tokenFingerprint(oauthToken) : 'none'}`);
+  console.log(`[${new Date().toISOString()}] POST ${STREAMABLE_HTTP_ENDPOINTS.mcp} - Session: ${sessionFingerprint(sessionId)}, Token: ${oauthToken ? tokenFingerprint(oauthToken) : 'none'}`);
 
   let transport: StreamableHTTPServerTransport;
 
@@ -1117,7 +1117,7 @@ app.post(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
 
   if (route === 'use-existing') {
     transport = existingRuntime!.transport;
-    console.log('[Streamable HTTP] Using existing transport for session:', sessionId);
+    console.log('[Streamable HTTP] Using existing transport for session:', sessionFingerprint(sessionId));
     sessionManager.touchSession(sessionId);
   } else if (route === 'not-found') {
     // A session id we do not hold. 404 IS THE ANSWER, and it is the one part of
@@ -1141,7 +1141,7 @@ app.post(STREAMABLE_HTTP_ENDPOINTS.mcp, async (req: Request, res: Response) => {
     // and re-initialize. The status is what the spec and the SDK server agree
     // on; the client's behaviour is Quinn's test, and it is the acceptance for
     // this change rather than a detail after it.
-    console.log('[Streamable HTTP] Unknown session, asking the client to start a new one:', sessionId);
+    console.log('[Streamable HTTP] Unknown session, asking the client to start a new one:', sessionFingerprint(sessionId));
     return res.status(404).json({
       error: 'Session not found',
       error_description:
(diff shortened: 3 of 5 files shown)


jwfing (commented):
## Review: Session ids are credentials, so stop writing them into the logs

**Summary:** A focused, well-tested security-hardening change that routes every session-id log site through a non-reversible `sessionFingerprint()`, matching the existing `tokenFingerprint` treatment — I found no blocking issues.

### Requirements context
No `docs/superpowers/` (or `docs/specs/`) directory exists in this repo, so there is no matching spec/plan — assessed against the PR description, the stated threat model (session id served without a token check ⇒ bearer credential; runtime logs readable over an API), and the surrounding code. Verified the change against the actual source in a workspace checkout of head `303fe5b`.

### Findings

#### Critical
(none) — this is the good kind of security PR: it removes a credential-at-rest without changing any routing or session behavior.

**Coverage I verified:**
- **Completeness (functionality):** grep for `console.(log|error|warn|info)` lines referencing `sessionId | newSessionId | initializedSessionId | transport.sessionId` across all of `src/` (not just the two changed files) returns **zero** raw sites remaining. The only leftover match is `session-manager.ts:565` (`${sessionIds.length}` — a count, not an id). ~21 `sessionFingerprint(` call sites, matching the PR's claim.
- **No other leak channel (security):** no request-logging middleware (morgan/pino-http), no `req.headers` / full-header dumps, and no other code path logs the `Mcp-Session-Id` header value. No error-response body (`error_description`/`res.json`) echoes the raw id back to the caller.
- **Construction parity (security):** `sessionFingerprint` (`session-manager.ts:24`) is `createHash('sha256').update(id).digest('hex').substring(0,8)` — identical to `tokenFingerprint` (`server.ts:134`), with the added falsy→`'none'` guard that preserves the prior `${sessionId || 'none'}` behavior. Non-reversible; consistent treatment of two equivalent credentials.
- **No scoping bug:** the shutdown loop at `server.ts:1589` uses the `for (const [sessionId, transport] of sseTransports)` loop key — correctly in scope.
- **Performance:** one extra sha256 over a ~36-char string per log line, on an I/O path that already writes to the log. Negligible; no hot-path concern.

#### Suggestion
- **software engineering — `session-fingerprint.test.ts:42` regex has real blind spots.** The source-reading guard is a nice idea (and the anti-vacuity meta-test at line 61 is the right instinct), but the `RAW` regex matches exactly three enumerated shapes: `${…sessionId…}` / `${…sessionId ||`, and a trailing positional `, sessionId)`. It does **not** catch string concatenation (`console.log('sid ' + sessionId)`), a *middle* positional arg (`console.log('a', sessionId, 'b')` — no closing `)` right after), or property access (`${sessionId.slice(0,4)}`). The PR body says it "fails on a shape nobody has thought of yet" — strictly it fails only on the three known shapes. None of those unhandled shapes exist in the code today, so blast radius is low, but a future `+ req.headers['mcp-session-id']` would slip past a green test. Consider broadening the regex (or asserting on the raw identifier token near any `console.*(` call) so the guard is closer to the "any raw id" claim it makes.

#### Information
- **security — 32-bit fingerprint ⇒ possible collisions.** 8 hex chars = 32 bits, so at high concurrent-session counts two distinct sessions can print the same fingerprint, marginally weakening support correlation. This is inherited from `tokenFingerprint` by design (consistency is the point), so no change requested — just noting the trade-off.
- **scope — stolen id still usable.** As the PR explicitly states, this stops writing the credential down but does not make a captured session id inert; binding it to `oauthTokenHash` is separate follow-up work. Correctly out of scope here.
- **conventions:** new test is colocated as `src/http/session-fingerprint.test.ts`, matching the repo's `src/**/*.test.ts` include glob and neighboring test files; import uses the `.js` extension per convention. (Could not execute the suite — deps aren't installed in the shallow review clone — but tsc/lint/238-tests are reported clean by the author and the change is static/self-evident.)

### Verdict
**approved** (informational — human approval still required via the GitHub approve flow). Zero Critical findings; the one Suggestion and the Information notes are non-blocking. Clean, correct, and complete for its stated scope.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: Session ids are credentials, so stop writing them into the logs

**Summary:** A tight, correct security fix that replaces raw session ids with a deterministic 8‑char SHA‑256 fingerprint at every log site, backed by a source‑scanning regression test; no Critical findings.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the linked threat model (session id = bearer credential yielding stored project key + full tool access without a token check), and repo conventions. The core claim checks out: `sessionFingerprint()` uses the **identical** construction to the existing `tokenFingerprint()` (`src/http/server.ts:134-136`) — `sha256 → hex → substring(0,8)` — so the "two equivalent credentials, one treatment" rationale holds.

### Findings

#### Critical
(none) — the change only rewrites log *strings*; every logic path (`touchSession`/`createSession`/`deleteSession`, `sseTransports` map keys, analytics payloads) still receives the raw id (`server.ts:1172,1273,1366,1434,1460`). No functional or routing behavior changes, and no existing behavior is broken.

#### Suggestion
- **[software-engineering / scope] `scripts/verify-platform-paths.mjs` is a separate concern and is not actually wired into CI.** The PR's stated intent for this script is "safe to run in CI on every push" and "catch bad platform routes before deploy," but `.github/workflows/ci.yml` runs only `lint` / `test` / `build`, and nothing invokes `npm run verify:paths` (`package.json:27`). As shipped it's a manual tool that will *not* automatically catch the next route typo — the exact failure mode it was written to prevent. Either add a CI step that runs it (note it makes live network calls to `api.insforge.dev`, so it needs egress and belongs in a scheduled/deploy job rather than the unit-test job), or state in the header that it's intentionally manual. Separately, this path‑verifier belongs conceptually to the revoke‑404 work (#96), not the session‑id‑logging change — bundling it here widens the PR's scope beyond its title/"Why."
- **[software-engineering / test robustness] The source‑scan guard has blind spots a future regression could slip through.** `session-fingerprint.test.ts` (`RAW` regex + `files = ['server.ts','session-manager.ts']`) only matches the three hardcoded identifiers `sessionId|newSessionId|initializedSessionId` and only scans those two files. A new log site that names the variable differently (e.g. `sseSessionId`, `id`) or lives in a third file (e.g. a sweeper) would evade the guard while still leaking a raw id. Consider matching on any `…[sS]essionId` identifier and deriving the file list from a glob.

#### Information
- **[functionality] Positional‑arg regex only catches the id as the *final* argument.** `,\s*…sessionId…\s*\)` matches `console.log('x', sessionId)` but not `console.error('x', sessionId, err)` (id not last). None of the current sites hit this, but it's a gap worth noting alongside the point above.
- **[functionality] `verify-platform-paths.mjs` `EXISTS` set omits 5xx.** A route that exists but returns 500 on the empty `{}` POST body would be reported as a non‑existent route (false positive). Harmless while the script is manual, but would cause flakiness if wired into CI as suggested.
- **[software-engineering] Nice touches:** the meta‑test proving the scan is non‑vacuous, and the explicit `none` handling for `undefined`/`null`/`''`, are the right instincts. 8 hex chars (32 bits) is more than enough entropy for support‑log correlation with negligible collision risk.

#### Security
No security regressions — this is a net security improvement (removes a bearer credential from logs). The verifier script sends no secrets and only unauthenticated probes. ✅

#### Performance
One extra SHA‑256 over a ~36‑char UUID per logged request; negligible and identical in cost to the pre‑existing `tokenFingerprint` already on the same hot path. No concern. ✅

### Verdict
**approved** (informational; human approval still required via the GitHub approve flow). Zero Critical findings — the security fix is correct, well‑tested, and cleanly scoped on the logging change itself. The suggestions above (wire the verifier into CI or document it as manual; broaden the guard regex) are non‑blocking and can be addressed in a follow‑up.

jwfing (commented):
## Review: Session ids are credentials — stop writing them into the logs

**Summary:** A clean, well-scoped security fix that replaces raw `Mcp-Session-Id` logging with a deterministic 8-char SHA-256 fingerprint at every site across both transports, backed by a source-scanning guard test; no blocking issues.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory. Assessed against the PR description, the linked revoke bug (#96), and repo conventions (gates = lint + vitest + build per `.github/workflows/ci.yml`).

---

### Critical
(none)

### Suggestion

**1 — Software engineering / functionality: `verify-platform-paths.mjs` is added but never wired into CI.** `scripts/verify-platform-paths.mjs:1-80` + `package.json:26` add a `verify:paths` script whose entire stated purpose is to catch route-typo 404s "in CI on every push" (the exact failure mode that shipped the revoke bug). But `.github/workflows/ci.yml:10-37` only runs `lint` / `test` / `build`, and nothing references `verify:paths` or `verify-platform-paths` anywhere under `.github/`. As shipped, the script won't actually prevent recurrence — it's a manual tool until someone invokes it. Note the likely reason it *isn't* in CI: it fetches a live external service (`https://api.insforge.dev`, `scripts/verify-platform-paths.mjs:22`), so adding it to the PR gate couples CI green to platform uptime. Worth either (a) wiring it into a non-blocking/scheduled job, or (b) stating in the script header that it's an on-demand check, so the "safe to run in CI on every push" comment doesn't read as "already runs in CI."

**2 — Software engineering: the guard test is line-based and has a blind spot.** `src/http/session-fingerprint.test.ts:53-58` flags a raw id only when a `console.*` token and the raw-id expression land on the *same* source line. A multi-line call such as `console.error(\n  \`...${transport.sessionId}\`,\n  error)`, or an object-shorthand `console.log('x', { sessionId })`, would slip past both the `console\.(log|error|warn|info)` filter and the `RAW` regex. No such call exists today (verified — all 21 logging sites are single-line), so this is not a live leak, just a durability gap in the guard that catches "a shape nobody has thought of yet." A one-line comment acknowledging the single-line assumption would be enough.

### Information

**3 — Doc/code mismatch in the path checker.** `scripts/verify-platform-paths.mjs:47` treats `{200,201,204,302,400,401,403,405,422}` as "route exists," but the header comment (`:37-39`) only enumerates `401 / 400 / 403`. Harmless — the wider set is more correct — but the comment and code drifted.

**4 — The checker sends live POSTs to prod OAuth endpoints.** `scripts/verify-platform-paths.mjs:38-42` POSTs `{}` to `/api/oauth/v1/token` and `/api/oauth/v1/revoke`. Confirmed side-effect-free (empty body → 400/401, no token revoked), so this is fine to run — noting it only because it's easy to overlook that a "path existence" probe is hitting mutation routes.

---

### Dimension coverage
- **Software engineering:** Solid. New `session-fingerprint.test.ts` covers stability, non-reversibility, fixed length, and the `none`/`null`/`''` branch, plus a non-vacuous meta-test proving the scanner actually fires (`:60-70`). The guard-test approach (read source, not behaviour) is the right call given SSE sites are invisible to a runtime grep. Import style, error handling, and naming all match the surrounding code. `sessionFingerprint` is byte-identical in construction to the existing `tokenFingerprint` (`server.ts:134-136`) — good consistency.
- **Functionality:** The fix does what it claims. Grep of `src/**` confirms every `console.*` site touching a session id in `server.ts` and `session-manager.ts` now routes through `sessionFingerprint`, and no other source file logs a session id. Correctly handles the empty/`undefined`/`null` id (streamable GET/DELETE/messages paths where the header may be absent) by printing `none` rather than crashing or printing `undefined`. Analytics (`trackSessionCreated`) does not carry the raw id, so no external-sink leak remains.
- **Security:** This is the point of the PR and it lands. 8 hex chars = 32-bit space — negligible collision risk for log correlation, and non-reversible. No secrets newly introduced; no auth path touched; no new dependency (`crypto.createHash` is stdlib). As the PR itself notes in scope, this stops writing the credential down but does not bind a stolen id to `oauthTokenHash` — correctly called out as separate follow-up, not a gap in this change.
- **Performance:** One SHA-256 over a ~36-char string per log line — trivial, and only on already-logging code paths. No new queries, loops, or blocking I/O. No concern.

---

### Verdict
**approved** (informational — human approval via the GitHub approve flow is separate). Zero Critical findings; the two Suggestions are non-blocking follow-ups and the Information items are minor. Nice, tightly-scoped fix with a genuinely thoughtful test.

jwfing (approved):
LGTM - approved.

PR #99, opened 2026-07-31
A session id alone is not enough, on all three verbs

tonychang04:
The last thing before the deploy. Stacked on #98.

Fingerprinting the id out of the logs (#98) stopped us **writing the credential down**. This stops a copy that escaped anyway from **being used**: a session opened with an OAuth token now requires the same token to use it.

## Three verbs, one function

Quinn's first version bound POST. A session is reachable by three — a POST-shaped fix for a session-shaped problem. They demonstrated the gap rather than arguing it:

```
DELETE /mcp with just the session id   -> 200
the victim's next valid request        -> 404
```

One request, no credential of any kind, and someone else is logged out. GET is quieter and worse: it opens the server→client stream on the id alone and receives everything pushed for that session, forging nothing.

`refuseMismatchedCredential` is called by all three handlers, always **after** the handler has established the session exists — so the 404-before-authentication property is untouched.

## Two things I got wrong, found only by driving a real re-authorization

**1. The refusal was a 401.** That is the instruction "re-run OAuth" — and the client that just did holds a *new* token and retries with the session id it still has. Credential valid, session real, different sign-ins. A 401 sends it round OAuth again, to return with another new token and the same old id: an infinite loop caused by signing in again.

It is now **404** — the same answer routing already gives, meaning "start a new session". It also leaks less: "not found" and "not yours" become one answer.

**2. Worse, and invisible to my own test.** Routing prefers a session we hold, so a re-authorized client sending `initialize` *while its old session is still alive* routes to `use-existing` and was refused — and `initialize` is the very request that repairs the situation, so it was refused for as long as the stale session lived. A loop with a 24-hour exit. My escape-hatch test asserted `hasRuntime: false` and passed happily.

The binding now refuses **use** of a session, never **creation** of one: an initialize with a mismatched credential falls through to `create`. Reusing the session would be worse anyway — a different credential may be a different user or project.

## Verified at the route level, on the built artifact

Against a stubbed platform with a real OAuth-bound session:

```
owner, right token           POST    200
attacker, id only            POST    404
attacker, id only            GET     404
attacker, id only            DELETE  404
victim after failed DELETE   POST    200   <- the DoS is closed
re-authorized, new token     POST    404
re-authorized, initialize    POST    200, new session
dead id, no credential       POST    404   <- recovery path untouched
```

## Credit and carried-over reasoning

The `legacy`/`''` passthrough and the design notes are Quinn's, from the version they retracted — rejecting those would log out every x-api-key client on deploy. The full-hash-vs-8-char trap is theirs too and now has a test: `SessionData` holds a 64-char sha256 while `tokenFingerprint` is the first 8 chars, and comparing the two forms rejects every OAuth session on its second request.

Safe against mid-session logout because the SDK sends both together — measured with a real client, every POST carried the token.

262 tests, tsc clean.

---
## Summary by cubic
Bind session usage to the OAuth token that created it across POST/GET/DELETE, returning 404 on mismatches to block leaked-id abuse and avoid re-auth loops. Also map upstream/network failures to 503, enforce platform URL family via a source guard, and pin session ID generation on create to keep the initialize carve-out safe.

- **Bug Fixes**
  - Enforced credential binding after the session is confirmed alive; mismatches now return 404. GET and DELETE are checked too, closing the stream-snoop and one-request DoS gaps.
  - `initialize` with a mismatched credential now creates a new session instead of being refused, preventing re-auth loops and preserving recovery.
  - Legacy sessions (`x-api-key`) and paths that store no hash remain unaffected; comparisons use the full 64-char SHA-256, not the 8-char fingerprint.
  - Added upstream unavailability detection and return 503 for timeouts, DNS/connection errors, and platform 5xx. 429 still passes through; plain errors remain 500.
  - Added a source-level guard that enforces platform URL construction (`OAUTH_API_BASE` for OAuth paths, `INSFORGE_API_BASE` for others) and ensures only `platformFetch` calls `fetch`, preventing the revoke path family bug from recurring.
  - Pinned session creation to always generate a new `sessionId` (`randomUUID`) and ignore any caller-supplied id, preventing initialize-on-victim hijacks and keeping the initialize exemption safe.

<sup>Written for commit 2968584151a6447b5f7a04cf96c16e63c6eb61fc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/99?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/oauth-manager.ts
+++ b/src/http/oauth-manager.ts
@@ -93,7 +93,7 @@ const AUTH_CODE_TTL = 5 * 60; // 5 minutes
 /**
  * Generate a hash of the token for storage
  */
-function hashToken(token: string): string {
+export function hashToken(token: string): string {
   return createHash('sha256').update(token).digest('hex');
 }
 
--- a/src/http/platform-paths.test.ts
+++ b/src/http/platform-paths.test.ts
@@ -81,3 +81,34 @@ describe('platform URL construction', () => {
     expect(bare, 'expected only platformFetch to call fetch directly').toBe(1);
   });
 });
+
+/**
+ * A property that only became load-bearing when the binding landed.
+ *
+ * #99 exempts `initialize` from the credential check so a re-authorized client
+ * can recover. That is safe for exactly one reason: the session id created by
+ * the create path is generated, never taken from the request. If it were ever
+ * derived from the incoming `Mcp-Session-Id`, an attacker holding their own
+ * valid credentials could initialize ONTO a victim's session id, overwrite the
+ * entry in the session map and take the session over — the exemption handing
+ * them the very thing the binding prevents.
+ *
+ * Quinn measured it closed (asked for 11111111-…, got a server-generated id).
+ * This pins it, because "we happen to generate it" is not a property anyone
+ * would notice losing in review.
+ */
+describe('the create path never adopts a caller-supplied session id', () => {
+  const server = readFileSync(join(__dirname, 'server.ts'), 'utf8');
+
+  it('generates the new session id with randomUUID', () => {
+    expect(server).toMatch(/const newSessionId = randomUUID\(\);/);
+  });
+
+  it('feeds the transport that generated id and nothing from the request', () => {
+    const generators = [...server.matchAll(/sessionIdGenerator:\s*([^,\n]+)/g)].map((m) => m[1].trim());
+    // The meta-check: if this stops matching, the assertion below is vacuous.
+    expect(generators.length).toBeGreaterThanOrEqual(1);
+    // `() => sessionId` — the request's id — is the mutation this forbids.
+    expect(generators).toEqual(generators.map(() => '() => newSessionId'));
+  });
+});
--- a/src/http/platform-paths.test.ts
+++ b/src/http/platform-paths.test.ts
@@ -1,0 +1,83 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+
+/**
+ * The check that would have caught the revoke bug, and the one nothing had.
+ *
+ * `revokePlatformToken` called `/oauth/v1/revoke`. The platform serves that at
+ * `/api/oauth/v1/revoke` and answers 404 at the other, so revoke would have
+ * failed every time in production — while passing every test, because the stub
+ * matched `url.includes('/oauth/v1/revoke')`, which is true of both. Nothing
+ * else could catch it either: the reachable production probes
+ * (not-our-token -> 200) return before the upstream call, and CI never touches
+ * the platform.
+ *
+ * Iris proposed a live probe against api.insforge.dev. That verifies more than
+ * this does — it confirms the constants are CORRECT, not merely consistent —
+ * and it is worth having. But it couples CI to the platform being reachable,
+ * which makes a green build depend on someone else's uptime. This is the half
+ * that needs no network and no uptime, so it can run on every push:
+ *
+ *   the platform has two path families, and every URL must come from the
+ *   constant that names its family rather than from a hand-typed prefix.
+ *
+ * That is exactly the mistake that was made. `${INSFORGE_API_BASE}/oauth/...`
+ * type-checks, reads correctly, and is wrong.
+ */
+
+const source = readFileSync(join(__dirname, 'insforge-api.ts'), 'utf8');
+
+/** Every template-literal URL passed to platformFetch. */
+function platformUrls(): string[] {
+  return [...source.matchAll(/platformFetch\(\s*`([^`]+)`/g)].map((m) => m[1]);
+}
+
+describe('platform URL construction', () => {
+  it('finds the calls at all, so this test cannot pass by matching nothing', () => {
+    // The meta-check. A regex that silently stops matching turns this file into
+    // a green light that asserts nothing — the same vacuous-check trap as a
+    // ratio that can never disagree.
+    expect(platformUrls().length).toBeGreaterThanOrEqual(7);
+  });
+
+  it('builds every OAuth URL from OAUTH_API_BASE, never by hand', () => {
+    // `${INSFORGE_API_BASE}/oauth/v1/revoke` is a 404 on the real platform.
+    // It compiles, it reads correctly, and it is wrong.
+    const handBuiltOAuth = platformUrls().filter(
+      (u) => /oauth/i.test(u) && !u.startsWith('${OAUTH_API_BASE}')
+    );
+    expect(handBuiltOAuth).toEqual([]);
+  });
+
+  it('builds every non-OAuth URL from INSFORGE_API_BASE at the root', () => {
+    // The other family: /auth/v1, /organizations/v1, /projects/v1 — all at the
+    // root, and all 404 under /api. Measured, not assumed.
+    const wrong = platformUrls().filter(
+      (u) => !/oauth/i.test(u) && !u.startsWith('${INSFORGE_API_BASE}/')
+    );
+    expect(wrong).toEqual([]);
+  });
+
+  it('never puts an /api prefix on a root-family URL', () => {
+    // /api/auth/v1/profile is a 404. This is the inverse of the revoke bug and
+    // just as invisible to a reader.
+    const prefixed = platformUrls().filter((u) => u.startsWith('${INSFORGE_API_BASE}/api/'));
+    expect(prefixed).toEqual([]);
+  });
+
+  it('routes every outbound platform call through platformFetch', () => {
+    // The timeout lives in that wrapper, so a bare `fetch(` here is a call with
+    // no bound — which is how the callback exchange sat unclassified in
+    // server.ts until it was moved into this file.
+    //
+    // EXACTLY ONE is correct: platformFetch's own call, the line that applies
+    // the AbortSignal. A second one is someone adding a platform call that
+    // skips the timeout, which is the failure this asserts against — and if
+    // this ever reads 0, the wrapper itself has been refactored away and the
+    // bound with it.
+    const bare = [...source.matchAll(/(?<!platformF|F)etch\(/g)]
+      .length;
+    expect(bare, 'expected only platformFetch to call fetch directly').toBe(1);
+  });
+});
--- a/src/http/server.ts
+++ b/src/http/server.ts
@@ -8,9 +8,9 @@ import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/
 import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
 
 // Local imports
-import { getSessionManager, routeForSessionRequest, sessionFingerprint } from './session-manager.js';
+import { getSessionManager, routeForSessionRequest, sessionAcceptsCredential, sessionFingerprint } from './session-manager.js';
 
-import { getOAuthManager } from './oauth-manager.js';
+import { getOAuthManager, hashToken } from './oauth-manager.js';
 import {
   SERVER_CONFIG,
   INSFORGE_CONFIG,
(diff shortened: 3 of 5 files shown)


jwfing (commented):
## Review: "A session id alone is not enough, on all three verbs"

**Summary:** The core fix — binding a session's `Mcp-Session-Id` to the OAuth token it was opened with, and enforcing it on POST, GET, and DELETE — is correct, well-reasoned, and closes the described one-request DoS (DELETE), silent stream-hijack (GET), and session takeover (POST); the residual gaps below are non-blocking.

**Requirements context:** No matching spec/plan found — this repo has no `/docs/superpowers/` or `/docs/specs/` directory. Assessed against the PR description, the linked stacking on #98, and the surrounding code. The intent (session id is a bearer credential; a leaked copy must not be *usable*) is clear from the body and matches the implementation.

I verified the mechanism end-to-end in the workspace:
- `resolveProjectFromToken` stores `oauthTokenHash: hashToken(token)` — the full 64-char sha256 (`oauth-manager.ts:420`).
- `credentialMatchesSession` compares against `hashToken(presentedToken)` — the same full form, never `tokenFingerprint`'s 8 chars (`server.ts:257-263`). The full-vs-truncated trap the body calls out is genuinely avoided.
- 404-before-binding ordering holds on all three verbs: GET/DELETE resolve `getStreamableSession` → 404 first, then `refuseMismatchedCredential` (`server.ts:1413-1424`, `1451-1461`); POST reaches the check only inside `route === 'use-existing'` (`server.ts:1259-1261`). An id-only attacker (`presented === undefined`) is correctly refused with 404, and the recovery path for a dead id is untouched.
- The `initialize`-on-a-live-mismatched-session escape hatch (`server.ts:1255-1257`) correctly reroutes to `create` and re-resolves with the new token, avoiding the 24-hour loop described.

### Critical
(none)

### Suggestion

**1. Software engineering — the new tests don't exercise the handler wiring, and the per-verb parametrization is decorative.** `session-binding.test.ts` only tests the pure helpers `sessionAcceptsCredential` / `routeForSessionRequest`. The `it.each(VERBS)` blocks (`session-binding.test.ts:32-66`) never use the verb parameter — every row calls `sessionAcceptsCredential(STORED, …)` with identical arguments regardless of `'POST' | 'GET' | 'DELETE'`. The comment claims "adding a fourth [verb] without binding it fails a test rather than passing quietly," but that is not true: nothing here asserts that the GET/DELETE/POST *handlers* actually call `refuseMismatchedCredential`. Given the PR's own narrative is that the bug slipped past tests twice at exactly the handler layer (POST-only binding; the live-session `initialize` loop), an integration-level test that drives the three routes and asserts 404-on-mismatch / 200-on-match would be the test that actually guards this regression class. The manual route-level table in the description is good evidence but isn't captured in CI.

**2. Security (functionality gap) — the same session-id-alone attack remains open on the SSE `/messages` path.** `GET /sse` will resolve an OAuth token and create an SSE session carrying a real `oauthTokenHash` (`server.ts:1492-1495`, `1576`), but `POST /messages?sessionId=…` looks the transport up by the query-param id and forwards the message with no credential check (`server.ts:1624-1648`) — `refuseMismatchedCredential` is wired only into the three streamable `/mcp` verbs. So an OAuth-authenticated SSE session is still usable by anyone holding the id (which here travels in the URL query string, arguably more leak-prone than a header). This is pre-existing and outside the PR's stated "three verbs" scope, so it is not a blocker for this PR — but the PR's mission ("a session id alone is not enough") is only partly realized while `/messages` is left bound-free, and `sessionAcceptsCredential` is now exactly the tool to close it. Worth a tracked follow-up.

**3. Scope / documentation — the `error-status.ts` 503 change is unrelated to session binding and unmentioned in the PR body.** The diff adds `isUpstreamUnavailable` and a 503 retryable-class branch to `statusForHttpError` (`error-status.ts:46-98`, 114 lines incl. tests), but the PR description is entirely about credential binding and never references it. The change itself looks correct and is well-tested (4xx passthrough incl. 429, 5xx/timeout/DNS → 503, unattributable → 500). But per the repo's scope-discipline convention (one PR = one concern), this reads as a second concern bundled in silently — a reviewer working from the body wouldn't expect it. Recommend either calling it out in the description or splitting it into its own PR.

### Information

- **Test clarity:** the `'routeForSessionRequest still takes no credential at all'` test (`session-binding.test.ts:88-101`) is convoluted — `args` is derived from a hardcoded object literal, not from the function under test, so the only substantive assertion is the arity check `routeForSessionRequest.length === 1`. It passes, but it doesn't verify what its name/comment claim. A direct comment or a simpler arity assertion would be clearer.
- **Timing:** `sessionAcceptsCredential` uses `===` on the two hex hashes rather than a constant-time compare. Not exploitable here — the stored value is a sha256 an attacker can't preimage into a usable token, and matching requires presenting a token that hashes to it — so no change is needed; noting only for completeness.
- **Minor duplication:** on an `initialize` that *matches* a live session, `credentialMatchesSession` runs twice (the fallthrough guard at `server.ts:1255` and again inside `refuseMismatchedCredential` at `1261`), i.e. a second sha256 on a rare path. Negligible.

*Not run:* I did not execute the vitest/tsc gates in the review workspace; the body's "262 tests, tsc clean" claim is taken at face value.

### Verdict

**approved** (informational — no Critical findings; human approval is still a separate action). The security fix is sound and the reasoning is unusually well-documented. The suggestions above (handler-level test coverage, the SSE `/messages` follow-up, and surfacing the bundled error-status change) are all non-blocking.

jwfing (commented):
## Review — "A session id alone is not enough, on all three verbs"

**Summary:** Binds an OAuth-created streamable session to the token that opened it across POST/GET/DELETE (404 on mismatch), lets a re-authorized `initialize` fall through to session creation, and reclassifies upstream-unavailable failures as 503 — the core logic is correct, well-reasoned, and I found no blocking issues.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessed against the PR description, the stacked-PR chain (#95/#97/#98), and the linked design reasoning in the code comments. This matches the known shape of `insforge-mcp` (MCP server; gates are vitest + tsc + eslint + tsup; no spec dir).

### Verification performed
- Traced the hash contract end-to-end: `SessionData.oauthTokenHash` is set from `oauth-manager.hashToken(token)` (full 64-char sha256) for OAuth sessions (`server.ts:1576` ← `resolveProject`), `'legacy'` for x-api-key (`server.ts:1352`), `''` for the SSE legacy fallback (`server.ts:1526`). `credentialMatchesSession` hashes the presented Bearer with the *same* `hashToken` (`server.ts:262`), so both sides are the full form — the truncation trap the PR describes is genuinely avoided.
- Confirmed `getSessionData` is a **synchronous** in-memory map read (`session-manager.ts:351-353`), so calling `credentialMatchesSession` without `await` is correct — no silent `undefined→accept` bug.
- Confirmed the binding is placed **after** the session-existence check in all three handlers (`server.ts:1424` GET, `server.ts:1461` DELETE, and inside the `use-existing` branch for POST), preserving the 404-before-auth recovery property.
- Verified via context7 against the MCP TypeScript SDK that `_commonHeaders()` attaches the Bearer token before **every** request, including the standalone SSE GET stream (`_startOrAuthSse`) and `terminateSession` — so the binding will not break legitimate SDK clients on GET/DELETE.

## Findings

### Critical
(none)

### Suggestion

**1. Software engineering — the committed tests don't actually exercise the wiring they claim to guard** (`src/http/session-binding.test.ts:30-52`)
The `it.each(VERBS)` blocks run the *same* `sessionAcceptsCredential(...)` assertion three times; they never touch the POST/GET/DELETE handlers. So the file's stated guarantee — "adding a fourth [verb] without binding it fails a test rather than passing quietly" — is not enforced: a regression that dropped `refuseMismatchedCredential(...)` from, say, the GET handler (`server.ts:1424`) would pass every committed test. The route-level check the PR body describes ("Verified at the route level, on the built artifact") appears to have been manual and was not committed. The repo currently has no `server.ts` integration harness (no test imports `./server`), so this would be new ground — but given the whole point of the PR is per-verb coverage, a supertest-style test hitting the three handlers with (owner-token / id-only / wrong-token) would be the test that actually protects the property.

**2. Software engineering — the `routeForSessionRequest` structural-pin test is largely tautological** (`src/http/session-binding.test.ts:87-104`)
`Object.keys(... ? { hasRuntime: 0, sessionId: 0, isInitialize: 0 } : {})` derives its keys from a hardcoded literal, not from the function under test, so it asserts a literal against itself. Only `routeForSessionRequest.length === 1` carries real signal (guards against threading a token in as a positional arg). Consider simplifying — the intent ("routing takes no credential") is worth pinning, but this construction doesn't pin it.

### Information

**3. Security (scope) — the deprecated SSE `/messages` path still treats the id alone as a bearer credential** (`src/http/server.ts:1624-1648`)
`POST /messages?sessionId=…` reaches a live session on the query param alone with no credential binding — the same shape this PR closes for the streamable `/mcp` verbs. It's out of scope (deprecated 2024-11-05 transport) and I'm not asking for it here, but flagging so "id-alone is closed" isn't assumed to hold everywhere.

**4. Security — non-constant-time hash comparison** (`src/http/session-manager.ts:213-224`)
`sessionAcceptsCredential` compares with `===` while the repo already uses `timingSafeEqual` for credential checks (`src/http/client-id.ts:259`). Not exploitable here: the compared value is a server-computed sha256 of an *attacker-supplied* token, so exploiting a timing side-channel would require a sha256 preimage attack. `===` is acceptable; noted only for consistency.

**5. Functionality — custom/non-SDK clients that omit the token on GET/DELETE** 
Per the SDK verification above, official-SDK clients are safe. A hand-rolled client that sends the token on POST but not on the GET reconnect or DELETE would now get 404 (stream won't open; session lingers until TTL instead of terminating). The PR acknowledges this trade-off and it's the correct default; noting the blast radius.

**6. Performance** — no concerns. `getSessionData` is an O(1) map lookup and the binding adds at most one sha256 per request; no new loops, queries, or blocking I/O.

**7. `error-status.ts`** — the 500→503 reclassification (`isUpstreamUnavailable` at `error-status.ts:34-56`; 429 preserved by the 4xx-passthrough ordering in `statusForHttpError`) is correct and has thorough new unit coverage in `error-status.test.ts`. Clean.

## Verdict
**approved** (informational — no Critical findings; the two Suggestions are non-blocking, chiefly a test-coverage gap for the route wiring). Human approval via the normal GitHub flow still required.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: bind session usage to the OAuth token that created it

**Summary:** A correct, well-reasoned security fix that closes the leaked-session-id abuse across POST/GET/DELETE and repairs the re-auth loop; the core logic holds up under scrutiny, but the security-critical *handler wiring* is not actually guarded by an automated test.

### Requirements context
No `docs/superpowers/` (or `docs/specs/`) directory exists in this repo — no matching spec/plan found, so this is assessed against the PR description, the stacked #97/#98 context, and the surrounding code. The three PR-owned commits under review are `c5f0543` (session binding), `e51fe62` (platform-path source guard), and `e3c70d1` (pin generated session id); `insforge-api.ts`'s 503/`platformFetch` work arrives via the merged #97/#98 tips and is out of scope here.

### Verification performed
- Traced the binding end-to-end: stored `oauthTokenHash` is `hashToken(token)` (`oauth-manager.ts:420`, full 64-char sha256), and `credentialMatchesSession` presents `hashToken(extractOAuthToken(req))` (`server.ts:258-263`) — same derivation, so the comparison is sound. The full-hash-vs-8-char-fingerprint trap is genuinely avoided.
- Confirmed `refuseMismatchedCredential` is placed *after* the `getStreamableSession` existence 404 in all three handlers (`server.ts:1261`, `:1440`, `:1476`), so the 404-before-authentication recovery property is preserved.
- Confirmed the initialize carve-out is safe: mismatched `initialize` falls through to `create` (`server.ts:1255-1257`), which always mints `randomUUID()` (`server.ts:1372`) — an attacker holding a *valid* token cannot `initialize` onto a victim's id; they get their own fresh session and the victim's is untouched. Attacker with id-only + `initialize` → 401 challenge, no hijack.
- Confirmed `ACCESS_TOKEN_TTL_SECONDS` (24h, `access-token.ts:92`) equals `SESSION_TTL` (24h, `session-manager.ts:119`), so token rotation does not spuriously 404 clients mid-session.
- Ran the four `platform-paths.test.ts` assertions against the real `insforge-api.ts`: 7 `platformFetch` template URLs, OAuth via `OAUTH_API_BASE`, root family via `INSFORGE_API_BASE/`, exactly 1 bare `fetch(` — all hold.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering / test coverage — the handler wiring the fix depends on is not tested** (`session-binding.test.ts:27-54`, `server.ts:1440`, `server.ts:1476`). No test imports `server.ts` or exercises the Express handlers; every assertion targets the pure `sessionAcceptsCredential`. The `it.each(VERBS)` rows are cosmetic — the verb parameter is never referenced, so all three rows run the *identical* `sessionAcceptsCredential(STORED, …)` call. The file's own comment claims "the per-verb rows exist so that adding a fourth without binding it fails a test" (`session-binding.test.ts:16-17`), but nothing here touches a handler: deleting the `refuseMismatchedCredential(...)` line from the GET or DELETE handler — i.e. re-introducing the *exact* POST-only bug this PR set out to fix — would pass every test green. The PR body's route-level table (`owner→200`, `attacker id-only→404` ×3, `victim-after-DELETE→200`, `re-auth initialize→200 new session`) is precisely the regression guard that should be committed, e.g. a `supertest` run against `app` with a stubbed session manager. Given the author's own note that the first two bugs were "invisible to my own test," encoding that end-to-end drive is high-value. Not blocking because I verified the live code is correct by reading all three handlers — but the safety net is illusory.

#### Information
- **Software engineering — source-text regex tests are brittle** (`platform-paths.test.ts:29-33, 62-64`). The guards scrape `insforge-api.ts`/`server.ts` as strings. They pass today, but a URL built by concatenation or a variable (rather than an inline template literal) would silently drop out of `platformUrls()`; the `>= 7` meta-check catches "matches nothing" but not "matches most." Reasonable as a zero-network CI guard — just be aware it degrades to partial coverage without failing.
- **Software engineering — the "routing takes no credential" test is close to tautological** (`session-binding.test.ts:86-97`). It asserts a hand-written key list and `routeForSessionRequest.length === 1`. A credential threaded in as a *new field on the existing input object* (the most likely regression) changes neither, so the stated property ("if someone later threads a token into routing … this test fails") isn't really enforced.
- **Functionality — `credentialMatchesSession` runs twice on an `initialize`-to-live-session POST** (`server.ts:1255` then `:1261`): two sha256 hashes. Negligible; noting only for completeness.

#### Security / Performance
No security regressions: id-only attackers get 404 on all three verbs, `initialize` cannot hijack an id, 404 (vs 401) also collapses "not found" and "not yours" into one answer, reducing enumeration signal. Only token fingerprints are logged — no secret leakage. No new user input reaches SQL/shell, no new dependencies. Performance impact is a single `Map` lookup + one sha256 per request — no N+1, loops, or blocking I/O introduced.

### Verdict
**approved** (informational — human approval via the separate approve flow). Zero Critical findings; the implementation is correct and the reasoning in the comments is unusually thorough. The one Suggestion worth acting on before deploy is committing an integration test that drives the three verbs, so the POST-only class of bug can't silently return.

jwfing (approved):
LGTM - approved.

PR #100, opened 2026-07-31
A missing test file is a failure, not a pass

tonychang04:
Two lines, and it closes a gate that could silently stop working.

Iris adopted a file-scoped gate before the deploy — run one test file, let a non-zero exit stop the deploy. It works, and it works **only** because she invoked vitest directly:

```
npx vitest run <missing file>   exit 1   <- the gate
npm test  -- <missing file>     exit 0   <- the same gate, vacuous
```

`--passWithNoTests` lives in `test` and `test:unit`, so the invocation a CI author reaches for first is exactly the one that turns "the file is not there" back into green. That's two ways to lose the same gate — hers was the shell (a pipe masking vitest's exit code), this one was already committed.

The flag is vestigial: it presumably predates the suite existing. There are 269 tests now, and "no test files found" is worth failing on in every remaining case — a renamed file, a bad path in CI, a glob that stopped matching.

```
after:  npm test -- <missing>   exit 1
        npm test                20 files, 269 tests, pass
```

Its own PR rather than folded into the tool work, because it changes what CI treats as success — that should be visible in a diff of its own rather than buried under a feature.

---
## Summary by cubic
Treat missing test files as failures by removing `--passWithNoTests` from the `test` and `test:unit` scripts. CI now exits non-zero when a specified file or glob matches no tests, preventing silent passes.

<sup>Written for commit 87a77c2980b5e6a5ef2d8b21b6928b9f95468ec1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Tests**
  * Updated test commands to require at least one test, improving test-run reliability.


jwfing (commented):
## Review: "A missing test file is a failure, not a pass"

**Summary:** Dropping `--passWithNoTests` from the `test` and `test:unit` scripts correctly restores a non-zero exit when a test target matches no files — a small, well-scoped, and correct change.

### Requirements context
No `/docs/superpowers/` (or any `docs/`) directory exists in this repo, so there is no matching spec/plan — assessing against the PR description and the repo's CI wiring alone. The PR body is itself a clear statement of intent: make "no test files found" fail rather than pass silently.

### Verification of the core claim
Confirmed against the Vitest v4.1.6 docs and source (the repo pins `vitest ^4.1.0`):
- `passWithNoTests` **defaults to `false`**. `docs/config/passwithnotests.md`: *"prevents Vitest from failing if no tests are found."*
- The exit-code path (`test-run.ts` `hasFailed`, `logger.ts`): with no test modules, `passWithNoTests: true` → *"No test files found, exiting with code 0"*; otherwise → *"exiting with code 1"*.

So after this change, `npm test -- <missing-file>` exits 1, matching the PR's stated goal, while a full `npm test` still finds the suite and passes.

### Findings

**Critical** — (none)

**Suggestion** — (none)

**Information**
- *Functionality / no regression:* I traced every consumer of the changed scripts. `ci.yml` runs `npm run test`, which still matches `src/**/*.test.ts` (20 files) and stays green. `test:integration:real` (`.github/workflows/integration-real-project.yml`) targets a specific file and **never carried the flag**, so it is unaffected. `test:coverage` and `test:watch` are likewise unchanged. `package.json:19-23`. The only new failure mode is the intended one: an empty/mismatched test glob now fails loudly. ✅
- *Software engineering / test coverage:* No automated test accompanies the change, which is appropriate — this is a package-manager script/gate line with no runtime surface a unit test could exercise, and the repo has no harness for asserting on npm-script exit codes. The change is self-verifying via CI's own `npm run test` step. Not a gap.
- *Software engineering (pre-existing, out of scope):* `test` and `test:unit` are now byte-identical (`vitest run`). That redundancy predates this PR; flagging only so it's on the radar, not as something to fix here.
- *Behavioral note for future CI authors:* Vitest auto-sets `passWithNoTests = true` when `--changed` is passed (`resolveConfig.ts`). So a future `--changed`-based workflow would still tolerate an empty match set by design — worth knowing, but not affected by or in scope for this PR.
- *Security / Performance:* No security- or performance-relevant surface in this change — it only alters CI/test exit semantics.

### Verdict
**approved** (informational; posted as a COMMENT — the GitHub green-checkmark approval remains a separate human action). Zero Critical findings. The diff is two lines, the reasoning in the PR body is sound, the vitest behavior is verified, and no existing consumer regresses.

jwfing (commented):
## Review: Remove `--passWithNoTests` from `test` / `test:unit`

**Summary:** A correct, minimal, well-scoped change that makes a run with no matching test files exit non-zero, closing a silent-pass gate — I verified the behavior against vitest v4 and traced every in-repo invocation path; no breakage found.

### Requirements context
No `docs/superpowers/` (or other spec/plan directory) exists in this repo — assessing against the PR description alone. The PR's stated intent is that `npm test -- <missing file>` should exit `1` rather than pass vacuously, and that "no test files found" is worth failing on in every remaining case.

### Findings

#### Critical
(none)

#### Suggestion
(none)

#### Information

- **Functionality — behavior verified against vitest v4 (correct).** `package.json:19-20` drops `--passWithNoTests` from `test` and `test:unit`. In vitest v4.1.x (`vitest ^4.1.0`, `package.json:92`) `passWithNoTests` defaults to `false`; with it false, vitest's logger emits *"No test files found, exiting with code 1"* and `hasFailed()` returns `true` for an empty module set. So the removal restores fail-on-no-tests exactly as the PR describes — `npm test -- <missing>` will now exit `1`.
- **Functionality — no CI path is broken by the stricter gate.** The only automatic gates that call these scripts run the *full* suite, which matches 269 tests and therefore never hits the empty-set case: `ci.yml:31` (`npm run test`) and `publish-mcp.yml:29` (`npm run test --if-present`). `integration-real-project.yml:33` uses `test:integration:real`, which never carried the flag. No in-repo caller invokes `test`/`test:unit` with a filter that could legitimately match zero files (only `README.md:98` documents `npm run test`).
- **Software engineering — consistency & scope.** The change is appropriately isolated to `package.json` and leaves the sibling scripts self-consistent: `test:coverage` and `test:integration:real` never had `--passWithNoTests`, so all `vitest run` variants now share the same fail-on-no-tests semantics. No unit test is warranted (an npm-script string is not meaningfully testable), and giving CI-gate changes their own PR, as done here, is reasonable.
- **Security:** No security-relevant changes — no new user input, dependencies, secrets, or auth surface.
- **Performance:** No performance impact — a build-tooling flag change only.
- Minor: `test:unit` is now byte-identical to `test` and is not referenced anywhere in-repo (workflows and README use `test`). Not a defect and out of scope for this PR, but if it's genuinely unused it could be dropped or made an alias in a future cleanup.

### Verdict
**approved** (informational; the human still gives the explicit GitHub approval). Zero Critical findings — the change is correct, verified, and cleanly scoped.

jwfing (approved):
LGTM - approved.

PR #101, opened 2026-07-31
The repair page clears both scopes, not one

tonychang04:
Blair found that #90's repair instructions — live on the error page right now — half-perform.

A server can be defined in two scopes at once, and from add-mcp@2.0.0's own source:

```
program.command("remove <query>")
  .option("-g, --global", "Remove from global configs INSTEAD OF project-level")
```

**"instead of" is the defect.** Measured on a machine with both:

```
npx add-mcp remove <url> -y
  -> "Removed 1 server from 1 agent"    the project entry
  -> the user-level entry is STILL THERE
```

So the page printed one command, the command reported success, and the stuck user re-added and stayed stuck — with a cheerful confirmation that the repair had worked. That is exactly the silent-partial-success shape this page exists to end, committed by the page itself.

## Why not `-g`

Blair suggested `-g` or an all-scopes sweep. I checked the flag rather than taking it: **`-g` *instead* would only move which scope is missed**, and there is no all-scopes flag on `remove`. Both commands is the honest repair, and running either against a scope that holds nothing is harmless — add-mcp reports zero removed and exits 0.

Four commands now rather than three. That's the cost of the page being true, and a stuck user following four correct steps beats three that report success and leave them stuck.

## Tests

Updated at all three sites that pinned the old triple, plus new coverage that **both** scopes get a remove, and that the removes come **before** the re-add — re-adding first would rebuild the entry the removal exists to clear.

271 tests pass.

---
## Summary by cubic
Fixes the OAuth error page repair instructions to clear the MCP server from both project and global scopes, and tells users to run them from the project directory. This prevents partial “success” from scope or CWD mismatches.

- **Bug Fixes**
  - Reconnect flow runs both removals: `npx add-mcp remove <url> -y` and `npx add-mcp remove <url> -g -y`, then re-add; reason: `add-mcp@2.0.0`’s `-g` removes global instead of project, so both are required.
  - Error text now says to run the commands from the project directory where InsForge is used, since project removal is CWD-relative.
  - Tests ensure two removes appear before the add and that the directory guidance is present.

<sup>Written for commit 5d5dd582578c761aa71ff94778dd27842a0ff101. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OAuth reconnection by clearing both project-level and global server configurations before reconnecting.
  - Ensured cleanup commands run before the server is re-added, including when URLs are derived from the host.
  - Clarified that reconnection commands should be run from the project directory.

### DIFF
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -193,3 +193,27 @@ describe('the repair clears every scope it can live in', () => {
     expect(add).toBeGreaterThan(lastRemove);
   });
 });
+
+describe('the page says where to run the commands', () => {
+  /**
+   * Project scope is CWD-relative, so both removes miss a project entry
+   * written somewhere else. Blair measured it:
+   *
+   *   entry in ~/projA/.mcp.json, run from ~/projB
+   *     remove <url> -y      "No matching servers found"
+   *     remove <url> -g -y   "No matching servers found"
+   *     projA entry          STILL THERE
+   *
+   * Both results are honest — there is nothing to remove where they looked —
+   * and the user is still broken. That is the same silent-partial-success one
+   * directory over, and the only fix available to a page is to say where to
+   * stand. A sentence, not a command.
+   */
+
+  it('tells the reader which directory to run them from', () => {
+    for (const error of ['invalid_client', 'invalid_request']) {
+      const { message } = humanFormOf({ error }, MCP_URL);
+      expect(message).toMatch(/project directory/);
+    }
+  });
+});
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -67,6 +67,7 @@ describe('every browser-visible failure says what to do', () => {
       expect(action).toEqual([
         'codex mcp logout insforge',
         `npx add-mcp remove ${MCP_URL} -y`,
+        `npx add-mcp remove ${MCP_URL} -g -y`,
         'npx add-mcp https://mcp.insforge.dev/mcp',
       ]);
       expect(action.join(' ')).not.toContain('@insforge/install');
@@ -84,9 +85,14 @@ describe('every browser-visible failure says what to do', () => {
     // rebuilds the SAME mcpOAuth key and picks the dead registration back up,
     // so the commands alone repair nothing. Clearing the client's saved
     // authorisation is the step that works, and it comes first.
-    const [logout, remove, add] = reconnectCommand(MCP_URL);
+    const [logout, removeProject, removeGlobal, add] = reconnectCommand(MCP_URL);
     expect(logout).toContain('logout');
-    expect(remove).toContain('remove');
+    expect(removeProject).toContain('remove');
+    // TWO removes: add-mcp's -g means "instead of project-level", not "as
+    // well as", so one command clears one scope and reports success while the
+    // other copy survives. Blair measured that on a machine with both.
+    expect(removeGlobal).toContain('remove');
+    expect(removeGlobal).toContain('-g');
     expect(add).toContain(MCP_URL);
   });
 
@@ -114,6 +120,7 @@ describe('every browser-visible failure says what to do', () => {
     expect(reconnectCommand(slug)).toEqual([
       'codex mcp logout insforge',
       `npx add-mcp remove ${slug} -y`,
+      `npx add-mcp remove ${slug} -g -y`,
       `npx add-mcp ${slug}`,
     ]);
     // The two add-mcp commands are host-derived; the codex one names the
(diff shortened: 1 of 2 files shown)


jwfing (commented):
## Review: The repair page clears both scopes, not one

**Summary:** A tight, well-tested fix that appends a second `add-mcp remove ... -g -y` to the OAuth error page's reconnect instructions so both the project- and global-scope entries are cleared; the core factual claim checks out and I found no blocking issues.

### Requirements context
No matching spec/plan found — this repo has no `/docs/superpowers/` (or `docs/`) directory. Assessed against the PR description, the linked context (#90 / #74), and the surrounding source. The central factual claim was independently verified: I resolved `add-mcp` (neon-solutions/add-mcp) via context7 and confirmed the `remove` command exposes `-g, --global` ("Remove from global configs"), that the default invocation targets project scope, and that there is **no** all-scopes flag on `remove`. That matches the author's rationale exactly — two invocations really are the only way to clear both scopes, and the `-g INSTEAD` framing is consistent with the published option semantics.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — the prose still narrates only "first" and "last", now skipping two middle commands.** `src/http/oauth-error-response.ts:189-190` and `:205` tell the Codex user to "run the first command below. Then reconnect with the last one." With the list now four items long, the two `remove` commands in the middle are the ones this PR exists to make the user run — yet the prose doesn't call them out, and a user reading literally could run first + last and skip the removes, reproducing the exact silent-partial-repair this page is meant to end. This ambiguity pre-existed (there was always one un-narrated middle `remove`), so it's not a regression and non-blocking — but since the PR's whole thesis is "the page should be true," tightening the wording to "run the commands below in order" would close the gap. Downgrading from blocking precisely because it predates this change.

#### Information
- **Software engineering — test coverage is thorough and correctly targeted.** All three sites that pinned the old three-command triple are updated (`oauth-error-response.test.ts:69-71`, `:88`, `:120-124`), and the new `describe` block adds behavioral coverage for both required properties: both scopes get a remove (`:180-184`) and the removes precede the re-add (`:188-193`). The ordering test's add-detection (`findIndex(c => c.includes('add-mcp') && !c.includes('remove'))`) is robust for the fixture URL used. Good TDD discipline.
- **Functionality — the 4th command renders cleanly.** `src/http/templates/oauth-error.ts` maps each action entry to its own `<code>` block with `.map()`, so there's no hardcoded assumption of three commands; the extra line renders as expected.
- **Security — no new surface.** `mcpUrl` is derived from `SERVER_CONFIG.publicUrl` (canonical host, deliberately not the attacker-controllable `Host` header, per the extensive comment at `:104-116`), and the value is HTML-escaped by the template before rendering. The added line reuses that same value — no new injection, secret-logging, or auth change. No security-relevant concerns.
- **Performance — none.** `reconnectCommand` is a pure function returning a small string array; adding one element has no meaningful performance impact and there's no I/O or query involved.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings. The change is correctly scoped to two files, the behavior is verified against upstream `add-mcp` docs, and test coverage is solid. The one Suggestion about the prose wording is optional and pre-existing.

jwfing (commented):
## Summary

Adds a second `add-mcp remove … -g -y` step to the OAuth error page's repair instructions so both the project-level and global scopes are cleared, closing a silent-partial-success gap where `#90`'s three-command repair reported success while a duplicate global entry survived.

## Requirements context

No matching spec/plan found — this repo has no `/docs/superpowers/` (or `docs/specs/`) directory, so I assessed against the PR description, the linked prior work (`#74`, `#90`), and the surrounding code. The PR body and inline comments are unusually explicit about intent, which made the intended behavior clear.

I independently verified the PR's central external-behavior claim against `add-mcp@2.0.0`'s published source (`unpkg.com/add-mcp@2.0.0/dist/index.js`) rather than trusting the narrative:
- `-g, --global` → *"Remove from global configs **instead of** project-level"* — confirmed; the flag switches which scope `listInstalledServers({ global })` queries, it does not union both. ✔
- A `remove` against a scope holding no match logs *"No matching servers found"* and returns with exit 0 — confirmed harmless. ✔
- `-y, --yes` is a valid option on the `remove` subcommand. ✔

So the "instead of" premise, the "run both, either is harmless" claim, and the flags are all factually correct.

## Findings

### Critical
(none)

### Suggestion
(none)

### Information

- **Functionality / dependency coupling — `src/http/oauth-error-response.ts:165-170`.** The printed commands use unpinned `npx add-mcp`, so a user runs whatever version is current, while the correctness reasoning is pinned to `add-mcp@2.0.0`'s specific `-g = "instead of"` semantics and its two-scope (project/global) model. If a future add-mcp changes that flag's meaning, adds a third scope, or introduces an all-scopes flag, these instructions could become redundant or under-clear again — the same failure class this PR fixes. This coupling is pre-existing (the `#90` commands already depend on unversioned add-mcp behavior), not introduced here, and is **non-blocking**. Worth a mention only so the maintainers are aware the page's correctness tracks an external CLI's latest behavior. No change requested.

### Notes by dimension

- **Software engineering** — Good TDD discipline. All three existing sites that pinned the old three-command triple were updated to the new four-command form, and a focused new `describe` block adds behavior coverage: both scopes removed, the project-vs-global distinction (`-g` present on exactly one), and the removes-before-re-add ordering (`oauth-error-response.test.ts:165-195`). The ordering test using `lastIndexOf` / `findIndex(c => includes('add-mcp') && !includes('remove'))` is robust against the extra remove line. Style matches the file's established heavily-narrated idiom.
- **Functionality** — Solves the stated defect; verified above. Command order (logout → remove project → remove global → re-add) is correct, and every error branch that surfaces `reconnectCommand` (`invalid_client`, `invalid_request`, default) picks up the fix consistently.
- **Security** — No security-relevant concerns. `mcpUrl` derives from `SERVER_CONFIG.publicUrl` (canonical server config, deliberately not the attacker-controlled Host header — see the comment at `:104-116`), and every command line is HTML-escaped via `escapeHtml` in `templates/oauth-error.ts:100-102` before rendering. No new user input, secrets, auth changes, or dependencies.
- **Performance** — Not applicable; the change is pure in-memory string-array construction with no loops, I/O, or DB access.

## Verdict

**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the one Information note is a pre-existing coupling, not a defect in this change. Clean, tightly-scoped, well-tested, and the external claim it rests on holds up against the real package source.

jwfing (approved):
LGTM - approved.

PR #103, opened 2026-08-01
The repair page's last branch still recommended the no-op

tonychang04:
## TL;DR

The repair page's `default` branch still tells people to remove the MCP server and add it back — the one remedy this file already documents as repairing nothing. #101 fixed that wording in `invalid_client` and `invalid_request` and did not reach the third branch. One shared constant now, plus the property test that would have caught it.

Not a cutover blocker: the flip cohort hits `invalid_client`, which is correct today. This is the branch behind it.

## Why it matters

`/oauth/callback` forwards the **platform's** error code verbatim when there is no registered redirect to bounce it back to (`server.ts`), so the codes landing on `default` are ones we do not choose, and a browser is the only thing that ever sees them.

Quinn traced the shipped Claude Code v2.1.220 binary and settled what actually repairs a stuck client:

```
menu "Clear authentication" -> clearServerTokensFromLocalStorage(...)
                            -> preserveClientRegistration falsy -> deletes the whole
                               mcpOAuth record, clientId included        <- a real repair

claude mcp remove           -> calls the same function behind a catch that swallows
                               the failure; both records survive        <- no-op
npx add-mcp remove          -> writes ~/.mcp.json; the record is in ~/.claude.json
                                                                        <- cannot help
```

One of the three remedies a person would try clears anything, and it is the one that cannot be pasted from a code block. The `default` branch was recommending one of the other two.

## The change

- `CLEARING_INSTRUCTION` — one constant, used by every branch that prints commands. The sentence has been corrected three times (wrong installer, reconnect-repairs-nothing, the second remove for the scope `-g` misses) and each correction had to find every copy. The comment carries Quinn's trace as the reason it must not be "simplified" back to remove-and-re-add, which is the friendlier-reading instruction that silently stops repairing anything.
- The upstream `error_description` is **kept and led with** — it is the only account of what actually failed. What changes is that it no longer arrives with the reconnect commands and no way to make them work.
- The wording test asserted the right property over the wrong set: it named the two branches someone remembered. It is now a property over every code, with and without a description.

## Verification

- Reverted against the old text, the new property fails **7 ways** (every default-branch case, with and without a description). A green test I have not watched fail is not evidence.
- Full suite green: 280 passed, 20 files. `npm run build` clean, `npm run lint` 0 errors.

## The trade-off, stated

For a code like `access_denied` (someone clicked Deny), "clear your saved authorisation" is more than they need — the honest remedy is "try again". I did not gate the instruction by code, because the branch exists precisely for codes we cannot classify, and the branch **already printed the reconnect commands for all of them**. This adds the step that makes those commands work rather than adding advice where there was none. If we want per-code triage there, that is a separate change with a list of codes behind it.

---
## Summary by cubic
Fixes the repair page so it never recommends the remove‑and‑re‑add no‑op, adds per‑code copy (incl. PKCE), removes commands from the catch‑all, and hardens the HTML escaper to prevent stack‑trace leaks from non‑string values.

- **Bug Fixes**
  - Shared `CLEARING_INSTRUCTION` used only for stale-registration cases: `invalid_client` and redirect-URI `invalid_request`; excluded for `access_denied`, `server_error`, and all unknown codes.
  - New branches: `access_denied` (“This sign-in was not approved”) and `unsupported_response_type` (client asked for an unsupported flow). Both avoid clearing and commands.
  - PKCE: added `PKCE_REQUIRED_PAGE` and a server override for `invalid_request` without `code_challenge`; explains PKCE and shows no clearing.
  - `default` no longer prints commands or clearing; leads with the platform description (trims blanks, normalizes punctuation via `endsSentence`) and tells the user to restart sign-in.
  - Tests cover all codes: assert no-clearing where it’s wrong, forbid any “add it back” phrasing, pin new branch headings, and verify PKCE/blank-description/punctuation seams.
  - Normalize non-string `error_description` values (arrays, numbers, objects, null) to avoid a crash and stack-trace leak; fall back to the generic message with a 400 response.
  - Renderer hardening: `escapeHtml` now coerces non-string inputs to safe strings to prevent unauthenticated stack-trace disclosure; tests cover arrays, numbers, null, and HTML escaping.

<sup>Written for commit 18f09a4fd03125ecaa66d43e19ab63e56db6e9c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/103?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth error messages with clearer authentication-clearing and reconnection guidance.
  * Added specific messaging for cancelled sign-ins and unsupported response types.
  * Preserved valid platform-provided descriptions while normalizing punctuation and handling malformed or blank text.
  * Removed ineffective recommendations to remove and re-add connections.
  * Added dedicated guidance when required PKCE security information is missing.
  * Expanded OAuth error handling with consistent, actionable recovery guidance across known and unknown failures.
  * Prevented OAuth error pages from failing when unexpected heading values are provided.

### DIFF
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, vi } from 'vitest';
-import { humanFormOf, prefersHtml, reconnectCommand, sendOAuthError } from './oauth-error-response.js';
+import { humanFormOf, prefersHtml, reconnectCommand, sendOAuthError, PKCE_REQUIRED_PAGE } from './oauth-error-response.js';
 
 const MCP_URL = 'https://mcp.insforge.dev/mcp';
 
@@ -125,13 +125,13 @@ describe('every browser-visible failure says what to do', () => {
   // required it. That is this same bug pointed the other way — a test naming one
   // remedy stops distinguishing the cases where that remedy is not the remedy.
   // Blair caught it against the rendered page.
+  //
+  // The table is now exactly the codes that mean a stale registration, which is
+  // the only thing clearing repairs. Everything else moved to the table below —
+  // see the comment there for why that direction is the load-bearing one.
   it.each([
     ['invalid_client', undefined],
     ['invalid_request', undefined],
-    ['unsupported_response_type', 'Only response_type=code is supported'],
-    ['token_exchange_failed', undefined],
-    ['weird_new_code', undefined],
-    ['weird_new_code', 'the disk melted'],
   ])('%s (description: %s) never prints the commands without the step that makes them work', (error, error_description) => {
     const { message, action } = humanFormOf({ error, error_description }, MCP_URL);
     // Asserted, not guarded. This read `if (!action) return` — meant for
@@ -152,15 +152,58 @@ describe('every browser-visible failure says what to do', () => {
   // clearing must not be handed the clearing instruction. Without this the two
   // tables are just a list and nothing stops `access_denied` drifting back into
   // the default branch, which is how it got the wrong advice in the first place.
+  //
+  // The unknown codes belong here rather than above, and that is the decision
+  // this table encodes. What can actually reach the catch-all is enumerable:
+  // the platform answers every rejection it makes at its own authorize endpoint
+  // with a 400 JSON body rather than a redirect, so those never re-enter this
+  // server; the single code it puts on our callback is `access_denied`. Nothing
+  // arriving here is a stale registration, so nothing arriving here should be
+  // told to clear one — least of all a stranger during a cutover.
   it.each([
     ['server_error', undefined],
     ['access_denied', undefined],
     ['access_denied', 'The user denied the request'],
+    ['unsupported_response_type', 'Only response_type=code is supported'],
+    ['login_required', undefined],
+    ['invalid_scope', 'Client not authorized for scopes: mcp:write'],
+    ['temporarily_unavailable', undefined],
+    ['weird_new_code', undefined],
+    ['weird_new_code', 'the disk melted'],
   ])('%s (description: %s) is not told to clear a registration that works', (error, error_description) => {
     const { message, action } = humanFormOf({ error, error_description }, MCP_URL);
     expect(action).toBeUndefined();
     expect(message).not.toMatch(/Clear authentication/);
     expect(message).not.toMatch(/add(ing)? it back/i);
+  });
+
+  // The table above cannot tell a dedicated branch from the catch-all, because
+  // the catch-all is now command-free too and satisfies every negative in it.
+  // Deleting either branch left the suite green — so these pin the thing that
+  // makes them branches at all: they name what happened in their own words.
+  it.each([
+    ['access_denied', 'This sign-in was not approved'],
+    ['unsupported_response_type', 'Your client asked for a sign-in this server does not support'],
+  ])('%s says what happened in its own words, not the catch-all\'s', (error, heading) => {
+    const human = humanFormOf({ error }, MCP_URL);
+    expect(human.heading).toBe(heading);
+    expect(human.heading).not.toBe(humanFormOf({ error: 'a_code_with_no_branch' }, MCP_URL).heading);
+  });
+
+  it('tells the PKCE case its own cause, not the port one', () => {
+    // `invalid_request` is emitted from several sites with different causes.
+    // The branch copy is right for the redirect_uri mismatch and false here —
+    // a request without code_challenge has nothing to do with a restarted port,
+    // and clearing a registration does not add PKCE to a client. This is the
+    // site override, kept in this file so it is testable at all; the overrides
+    // written inline at their call sites are covered by nothing.
+    expect(PKCE_REQUIRED_PAGE.action).toBeUndefined();
+    expect(PKCE_REQUIRED_PAGE.message).toMatch(/PKCE/);
+    expect(PKCE_REQUIRED_PAGE.message).toMatch(/code_challenge/);
+    expect(PKCE_REQUIRED_PAGE.message).not.toMatch(/different port/);
+    expect(PKCE_REQUIRED_PAGE.message).not.toMatch(/Clear authentication/);
+    // The branch it overrides asserts the cause it must not inherit.
+    expect(humanFormOf({ error: 'invalid_request' }, MCP_URL).message).toMatch(/different port/);
   });
 
   it('treats a blank description as no description', () => {
@@ -183,11 +226,11 @@ describe('every browser-visible failure says what to do', () => {
       { error: 'weird_new_code', error_description: 'the platform said no' },
       MCP_URL
     );
-    expect(message).toContain('the platform said no. In Claude Code');
+    expect(message).toContain('the platform said no. Start it again');
     // A description that already punctuates itself is not given a second stop.
     expect(
       humanFormOf({ error: 'weird_new_code', error_description: 'it broke.' }, MCP_URL).message
-    ).toContain('it broke. In Claude Code');
+    ).toContain('it broke. Start it again');
   });
 
   it('names the host the person is actually talking to, in BOTH commands', () => {
@@ -229,21 +272,22 @@ describe('every browser-visible failure says what to do', () => {
     // Deliberately no longer the description ALONE. It is the only account of
     // what actually failed, so it is kept and it comes first — but it used to
     // arrive with the reconnect commands and no way to make them work, which
-    // is the same trap as the fallback sentence it sits beside.
+    // is the same trap as the fallback sentence it sits beside. It now closes
+    // with the one instruction that is true for every unknown code.
     const { message } = humanFormOf(
       { error: 'weird_new_code', error_description: 'the disk melted' },
       MCP_URL
     );
     expect(message).toMatch(/^the disk melted/);
-    expect(message).toMatch(/Clear authentication/);
+    expect(message).toMatch(/Start it again from your editor or client.$/);
   });
 
   it('still says something when there is no description either', () => {
-    // The lead sentence and the repair, both — a length check alone passed
+    // The lead sentence and a next step, both — a length check alone passed
     // happily on the old text, whose 40+ characters were the no-op advice.
     const { message } = humanFormOf({ error: 'weird_new_code' }, MCP_URL);
     expect(message).toMatch(/^The sign-in did not complete\./);
-    expect(message).toMatch(/Clear authentication/);
+    expect(message).toMatch(/Start it again from your editor or client\.$/);
   });
 });
 
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -119,12 +119,16 @@ describe('every browser-visible failure says what to do', () => {
   // and `npx add-mcp remove` writes a different file entirely, so commands
   // printed without the menu step leave the user in the loop the page exists
   // to end.
+  // The table is codes whose remedy IS clearing. `access_denied` was in it and
+  // should never have been: nothing is stale when someone cancels, so asserting
+  // /Clear authentication/ over it did not merely allow the wrong advice, it
+  // required it. That is this same bug pointed the other way — a test naming one
+  // remedy stops distinguishing the cases where that remedy is not the remedy.
+  // Blair caught it against the rendered page.
   it.each([
     ['invalid_client', undefined],
     ['invalid_request', undefined],
     ['unsupported_response_type', 'Only response_type=code is supported'],
-    ['access_denied', undefined],
-    ['access_denied', 'The user denied the request'],
     ['token_exchange_failed', undefined],
     ['weird_new_code', undefined],
     ['weird_new_code', 'the disk melted'],
@@ -142,6 +146,37 @@ describe('every browser-visible failure says what to do', () => {
     // phrasing — it was selling remove-and-re-add as the remedy, which reads as
     // the friendlier instruction in any wording someone reaches for next.
     expect(message).not.toMatch(/add(ing)? it back/i);
+  });
+
+  // And the other half of the same property: a code whose remedy is NOT
+  // clearing must not be handed the clearing instruction. Without this the two
+  // tables are just a list and nothing stops `access_denied` drifting back into
+  // the default branch, which is how it got the wrong advice in the first place.
+  it.each([
+    ['server_error', undefined],
+    ['access_denied', undefined],
+    ['access_denied', 'The user denied the request'],
+  ])('%s (description: %s) is not told to clear a registration that works', (error, error_description) => {
+    const { message, action } = humanFormOf({ error, error_description }, MCP_URL);
+    expect(action).toBeUndefined();
+    expect(message).not.toMatch(/Clear authentication/);
+    expect(message).not.toMatch(/add(ing)? it back/i);
+  });
+
+  it('does not run the platform\'s sentence into ours', () => {
+    // The forwarded description is not ours and carries no punctuation
+    // guarantee. Joined with a bare space, "The user denied the request" ran
+    // straight into "In Claude Code: run /mcp" as one sentence. Our own literal
+    // fallbacks all end in a full stop, so nothing in the fixtures showed it.
+    const { message } = humanFormOf(
+      { error: 'weird_new_code', error_description: 'the platform said no' },
+      MCP_URL
+    );
+    expect(message).toContain('the platform said no. In Claude Code');
+    // A description that already punctuates itself is not given a second stop.
+    expect(
+      humanFormOf({ error: 'weird_new_code', error_description: 'it broke.' }, MCP_URL).message
+    ).toContain('it broke. In Claude Code');
   });
 
   it('names the host the person is actually talking to, in BOTH commands', () => {
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -163,6 +163,17 @@ describe('every browser-visible failure says what to do', () => {
     expect(message).not.toMatch(/add(ing)? it back/i);
   });
 
+  it('treats a blank description as no description', () => {
+    // A present-but-blank `error_description` is truthy, so it used to skip the
+    // fallback and render ". In Claude Code: ..." — a sentence opening with a
+    // full stop. Reachable by hand: ?error=x&error_description=%20.
+    for (const blank of [' ', '   ', '\t']) {
+      const { message } = humanFormOf({ error: 'weird_new_code', error_description: blank }, MCP_URL);
+      expect(message).toMatch(/^The sign-in did not complete\./);
+      expect(message).not.toMatch(/^\s*\./);
+    }
+  });
+
   it('does not run the platform\'s sentence into ours', () => {
     // The forwarded description is not ours and carries no punctuation
     // guarantee. Joined with a bare space, "The user denied the request" ran
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -107,6 +107,35 @@ describe('every browser-visible failure says what to do', () => {
     }
   });
 
+  // The property, over every code rather than over the two branches someone
+  // remembered. The test above asserted the right thing about the wrong set:
+  // it named `invalid_client` and `invalid_request` explicitly, so the default
+  // branch went on telling people to remove the server and add it back — the
+  // remedy this same file documents as a no-op — for as long as nobody looked.
+  // A test that lists the cases it imagines stops covering the one it does not.
+  //
+  // Quinn traced why the distinction matters: only the Claude Code TUI action
+  // deletes the mcpOAuth record. `claude mcp remove` swallows its own failure
+  // and `npx add-mcp remove` writes a different file entirely, so commands
+  // printed without the menu step leave the user in the loop the page exists
+  // to end.
+  it.each([
+    ['invalid_client', undefined],
+    ['invalid_request', undefined],
+    ['unsupported_response_type', 'Only response_type=code is supported'],
+    ['access_denied', undefined],
+    ['access_denied', 'The user denied the request'],
+    ['token_exchange_failed', undefined],
+    ['weird_new_code', undefined],
+    ['weird_new_code', 'the disk melted'],
+  ])('%s (description: %s) never prints the commands without the step that makes them work', (error, error_description) => {
+    const { message, action } = humanFormOf({ error, error_description }, MCP_URL);
+    if (!action) return; // server_error offers no commands, so it needs no clearing step
+    expect(message).toMatch(/Clear authentication/);
+    // And it must not recommend the substitute that repairs nothing.
+    expect(message).not.toMatch(/adding it back is the usual remedy/);
+  });
+
   it('names the host the person is actually talking to, in BOTH commands', () => {
     // On the Manufact slug it must say the slug, or someone follows the
     // instruction and reconnects to the box we are migrating off.
@@ -142,9 +171,17 @@ describe('every browser-visible failure says what to do', () => {
     expect(human.message).toMatch(/different port/);
   });
 
-  it('falls back to the description for a code it has never seen', () => {
-    expect(humanFormOf({ error: 'weird_new_code', error_description: 'the disk melted' }, MCP_URL).message)
-      .toBe('the disk melted');
+  it('leads with the description for a code it has never seen', () => {
+    // Deliberately no longer the description ALONE. It is the only account of
+    // what actually failed, so it is kept and it comes first — but it used to
+    // arrive with the reconnect commands and no way to make them work, which
+    // is the same trap as the fallback sentence it sits beside.
+    const { message } = humanFormOf(
+      { error: 'weird_new_code', error_description: 'the disk melted' },
+      MCP_URL
+    );
+    expect(message).toMatch(/^the disk melted/);
+    expect(message).toMatch(/Clear authentication/);
   });
 
   it('still says something when there is no description either', () => {
--- a/src/http/oauth-error-response.test.ts
+++ b/src/http/oauth-error-response.test.ts
@@ -130,10 +130,18 @@ describe('every browser-visible failure says what to do', () => {
     ['weird_new_code', 'the disk melted'],
   ])('%s (description: %s) never prints the commands without the step that makes them work', (error, error_description) => {
     const { message, action } = humanFormOf({ error, error_description }, MCP_URL);
-    if (!action) return; // server_error offers no commands, so it needs no clearing step
+    // Asserted, not guarded. This read `if (!action) return` — meant for
+    // server_error, which is not in this table and so never reached it. What it
+    // would actually have done is let a branch that silently LOST its commands
+    // pass the test written to protect them: exactly the it-cannot-fail shape
+    // that let the default branch drift in the first place. Caught in review by
+    // john-bot and coderabbit independently.
+    expect(action).toBeDefined();
     expect(message).toMatch(/Clear authentication/);
-    // And it must not recommend the substitute that repairs nothing.
-    expect(message).not.toMatch(/adding it back is the usual remedy/);
+    // Broader than the one sentence that was removed. The defect was never that
+    // phrasing — it was selling remove-and-re-add as the remedy, which reads as
+    // the friendlier instruction in any wording someone reaches for next.
+    expect(message).not.toMatch(/add(ing)? it back/i);
   });
 
   it('names the host the person is actually talking to, in BOTH commands', () => {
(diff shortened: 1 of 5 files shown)


jwfing (commented):
## Review — #103 "The repair page's last branch still recommended the no-op"

**Summary:** A tight, well-tested fix that routes the OAuth error page's `default` branch through a shared `CLEARING_INSTRUCTION` constant, killing the last copy of the "remove and re-add" no-op that #101 didn't reach — no defects found.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` or `docs/specs/` directory, so this was assessed against the PR description, the linked #101/#74 context, and the surrounding code. Verified locally at head `0e53516`: `npx vitest run src/http/oauth-error-response.test.ts` → **29 passed**, `eslint` on both changed files → **0 errors**.

### Findings

#### Critical
(none)

#### Suggestion
(none)

#### Information

- **Software engineering — literal-string negative assertion is redundant belt-and-suspenders** (`src/http/oauth-error-response.test.ts:136`). The property test guards with `expect(message).not.toMatch(/adding it back is the usual remedy/)`, matching the exact removed copy. If the no-op advice were ever reintroduced with slightly different phrasing this negative would pass silently — but the positive `expect(message).toMatch(/Clear authentication/)` on line 134 is the load-bearing assertion and fully covers the actual requirement, so the negative is a harmless extra tripwire. No change needed; noting for awareness.

- **Functionality — default-branch message can read as a run-on** (`src/http/oauth-error-response.ts:258`). `` `${body.error_description || 'The sign-in did not complete.'} ${CLEARING_INSTRUCTION}` `` concatenates the platform's `error_description` (which we don't control and may not end in punctuation) directly before the instruction, e.g. "the disk melted In Claude Code: run /mcp…". Purely cosmetic — the escaping and content are correct — but a period/normalisation between the two segments would read cleaner. Author's call.

- **Narrative nit (no code impact):** the description says the reverted property "fails **7 ways**"; the `it.each` has 6 default-branch cases (the other two, `invalid_client`/`invalid_request`, were already fixed by #101 and pass against the old text). Doesn't affect the change.

### Dimension coverage

- **Software engineering ✅** — Correct DRY extraction: all three command-bearing branches (`invalid_client`, `invalid_request`, `default`) now share `CLEARING_INSTRUCTION`; `server_error` correctly keeps no commands. The enumerated-branch test is converted into a genuine property over every code including unknowns, which is exactly the regression that would have caught the original miss. Matches house conventions (`.js` import specifiers, the file's heavy-narration comment style). TDD/red-green claim verified by running it.
- **Functionality ✅** — Solves the stated problem: the `default` branch no longer recommends the measured no-op and instead leads with `error_description` (the only account of the real failure) followed by the working clearing step. Confirmed the branch already carried `action: reconnectCommand(mcpUrl)`, so this adds the step that makes those already-printed commands actually work rather than changing the command set. The `access_denied` over-instruction trade-off is explicitly documented and reasonable.
- **Security ✅** — No new attack surface. `error_description` is platform-forwarded and could be untrusted, but `renderOAuthErrorPage` passes `message` through `escapeHtml` (`src/http/templates/oauth-error.ts:99`), so there's no HTML/XSS injection. No secrets, tokens, or PII newly logged or returned; no auth/authorization logic touched; no new dependencies.
- **Performance ✅** — Pure synchronous string composition on a cold browser-error path. No queries, loops, allocations, or blocking I/O concerns.

### Verdict
**approved** (informational — a human still gives the explicit GitHub approval). Zero Critical findings; the two Information notes are optional polish.

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: The repair page's last branch still recommended the no-op

**Summary:** A tight, well-tested fix that hoists the OAuth "clearing" guidance into one `CLEARING_INSTRUCTION` constant and applies it to the `default` branch #101 missed, so the browser-visible repair page no longer sells remove-and-re-add (a documented no-op) as the remedy.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the linked history (#74 → #101 → this), and the in-file design comments, which are unusually thorough. Intent is clear and the change stays within it.

### Findings

#### Critical
(none)

#### Suggestion
(none)

#### Information
- **Functionality / cosmetics — `src/http/oauth-error-response.ts:258`.** On the `default` branch the message is `` `${error_description || 'The sign-in did not complete.'} ${CLEARING_INSTRUCTION}` ``. When the platform's `error_description` arrives without trailing punctuation (the test's own `'the disk melted'` case), the rendered sentence runs together: *"the disk melted In Claude Code: run /mcp…"*. Purely cosmetic and only on the unknown-code path with an unpunctuated upstream string; the built-in fallback (`'The sign-in did not complete.'`) already ends in a period. Author may ignore.

### Dimension notes
- **Software engineering — strong.** The property test now runs over *every* code (with and without a description) instead of the two branches someone remembered, which is exactly the gap that let the `default` branch drift for two PRs. The follow-up commit (0b91ada) also fixed a real hole in the *new* test — the `if (!action) return` early-out would have silently passed a branch that lost its `action`; it now `expect(action).toBeDefined()`. Mutation-verified (restore old text → 8 fail; drop `action` → 6 fail). I re-ran the suite locally: **280 passed / 20 files**, `lint` 0 errors, `build` clean — all claims hold.
- **Functionality — solves the stated problem.** Verified the `default` branch is genuinely reachable: `/oauth/callback` forwards the platform's `error` code verbatim to `sendOAuthError` when there is no registered redirect to bounce it back to (`src/http/server.ts:657-660`), so arbitrary upstream codes land here. The DRY extraction is safe — every branch that uses `CLEARING_INSTRUCTION` ("run the first command below … reconnect with the last one") also sets `action: reconnectCommand(mcpUrl)`, so the sentence's references to the command list are always satisfied; `server_error` correctly uses neither.
- **Security — no new surface.** The `default` branch reflects the platform-supplied `error_description` (influenceable via the callback query string), but the template escapes `heading`/`message`/every `action` line through `escapeHtml` (`src/http/templates/oauth-error.ts:98-114`), so there is no XSS. This reflection is also pre-existing, not introduced here. No secrets/PII logged, no auth/authorization change, no new dependencies.
- **Performance — n/a.** Pure string construction; no queries, loops, or I/O.

### Verdict
**approved** (informational — human approval via the separate approve flow). Zero Critical findings; the single Information note is optional cosmetics.

jwfing (commented):
## Review — #103 (repair page's `default` branch no-op)

**Summary:** A tight, well-tested copy/logic fix that stops the `default` OAuth-error branch from recommending the remove-and-re-add no-op, adds a dedicated `access_denied` branch, and de-duplicates the repair text into one `CLEARING_INSTRUCTION` constant — I found no blocking issues.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in `insforge-mcp` — assessing against the PR description, the referenced #101/#74, and the surrounding code alone. The PR's stated intent (extend the #101 fix to the third branch, guard it with a property test) matches the diff exactly and stays in scope (2 files, source + test).

I verified the load-bearing claims against the checked-out head:
- **Reachability of `default`/`access_denied` with platform-chosen codes** — confirmed: `src/http/server.ts:657-660` forwards the platform's raw `error` verbatim into `humanFormOf` when there is no registered `redirect_uri` to bounce to (`server.ts:642-651` handles the redirect case). So the codes hitting these branches really are ones we don't choose.
- **`error_description` is not an injection vector** — the forwarded (now led-with) description flows into `renderOAuthErrorPage`, which passes both `message` and each `action` line through `escapeHtml` (`src/http/templates/oauth-error.ts`). No XSS.
- **`mcpUrl`** is built from `SERVER_CONFIG.publicUrl` (`oauth-error-response.ts:324`), not the attacker-controlled Host header — no host-header spoofing into the printed commands.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality —** `oauth-error-response.ts:279-295` (default branch) & `server.ts:531-535`. `unsupported_response_type` is a genuine *client request* error (the client sent `response_type≠code`), and it is reachable to this page via `sendOAuthError` at `server.ts:532`. Falling to the `default` branch, it now receives the full "Clear authentication + reconnect" remedy, which won't fix a malformed `response_type`. This is the exact "trade-off, stated" the PR body calls out and explicitly defers ("per-code triage … is a separate change"), and the branch's whole justification is codes we can't classify — so this is non-blocking. Flagging only so the deferred item is on record: if per-code triage lands later, `unsupported_response_type` is a concrete candidate alongside `access_denied`/`server_error`.

#### Information
- **Functionality —** `oauth-error-response.ts:211-214` (`endsSentence`). A whitespace-only `error_description` (e.g. `" "`) is truthy, so it bypasses the `|| 'The sign-in did not complete.'` fallback, trims to `""`, and yields a leading lone period: `". In Claude Code: …"`. Cosmetic and essentially unreachable in practice (the platform would have to forward a blank-but-present description), so not worth guarding — noted only for completeness.
- **Software engineering —** the extraction of `invalid_client`'s inline text into `CLEARING_INSTRUCTION` is byte-identical for that branch; for `invalid_request` it normalizes "reconnect with the last" → "reconnect with the last one". Harmless, and arguably an improvement (consistent wording). No test asserts the old phrasing, so nothing regresses.

### Dimension coverage
- **Software engineering:** Strong. TDD is evident (PR reports the new property fails 7 ways when reverted against the old text), the fix converts a case-listing test into a property over every code (with/without description), and it adds the negative half (`access_denied`/`server_error` must *not* be told to clear). ESM `.js` import extensions and named-export conventions match the file. No concerns.
- **Functionality:** Solves the stated bug. `default` now leads with the upstream description, appends the real repair, and `endsSentence` prevents the run-on into `CLEARING_INSTRUCTION`. `access_denied` correctly returns no `action` and no clearing text. See the two non-blocking notes above.
- **Security:** No new risk. Forwarded description is HTML-escaped, `mcpUrl` is the canonical `publicUrl`, no secrets/PII logged or newly returned, no auth/authorization logic touched.
- **Performance:** N/A — pure per-request string composition, no loops, I/O, or queries. No concern.

### Verdict
**approved** (informational; a human still gives the GitHub green-check via the approve flow). Zero Critical findings — the two notes are non-blocking and one is already acknowledged in the PR's trade-off section.

jwfing (commented):
## Review — #103 "The repair page's last branch still recommended the no-op"

**Summary:** A focused, well-tested copy fix that stops the `default` branch of the OAuth error page from recommending the remove-and-re-add no-op, extracts the repair sentence into one shared `CLEARING_INSTRUCTION`, adds an `access_denied` branch that (correctly) withholds the clearing remedy, and hardens the forwarded-description path against blank/run-on rendering.

**Requirements context:** No matching spec/plan found — `insforge-mcp` has no `/docs/superpowers/` (or `docs/specs/`) directory. Assessed against the PR description, the referenced #101/#74, and the surrounding code (`src/http/oauth-error-response.ts`, `templates/oauth-error.ts`).

I reviewed all four dimensions.

### Critical
(none)

### Suggestion
(none blocking)

### Information

**Functionality — clearing advice now attaches to protocol-shaped codes.** `src/http/oauth-error-response.ts:301-321` — the `default` branch now appends `CLEARING_INSTRUCTION` (+ the reconnect `action`) to *every* unclassified code. The property test at `src/http/oauth-error-response.test.ts:128-149` pins this even for `unsupported_response_type`, whose real remedy ("send `response_type=code`") is not "clear your saved authorisation." The blast radius is small — these codes are platform-forwarded, browser-only, and the forwarded `error_description` still leads the message — and the PR body explicitly reasons about and accepts this trade-off ("the branch exists precisely for codes we cannot classify… per-code triage is a separate change"). Noting it only so the decision is on the record; not a blocker.

**Software engineering — `endsSentence` treats `:`/`;` as terminal.** `src/http/oauth-error-response.ts:233-236` — `/[.!?:;]$/` means a forwarded description ending in a colon (e.g. `"Details:"`) is left unpunctuated and renders as `"Details: In Claude Code: run /mcp…"`. Cosmetic only; the anti-run-on intent is preserved. The regex choice is reasonable — flagging as an edge worth an eye if platform descriptions ever end in a colon.

**Copy — `invalid_request` wording normalised.** `src/http/oauth-error-response.ts:258-271` vs. the old inline text: the extraction changed `"reconnect with the last."` → `"…the last one."`. Byte-identical for `invalid_client`; this is a benign, arguably-better normalisation for `invalid_request`. Just noting the user-visible copy delta since it isn't called out.

### What's solid
- **Refactor safety:** `CLEARING_INSTRUCTION` concatenates byte-for-byte to the previous inline `invalid_client` message — behaviour-preserving for the two branches that already carried it.
- **Security:** the forwarded `error_description` is reflected but goes through `escapeHtml` for both `message` and each `action` line in `templates/oauth-error.ts` — no reflected XSS introduced. No secrets/PII logged or added to responses; no auth/authz change; no new dependencies.
- **Performance:** pure string work, no I/O, no loops in a hot path — nothing to flag.
- **Test coverage — the strongest part.** The wording property is now asserted over *every* code in both directions (must-clear vs. must-not-clear), the `if (!action) return` guard-that-couldn't-fail is replaced with `expect(action).toBeDefined()`, and there are targeted regressions for blank-description (`%20`), the punctuation seam, and `access_denied` both ways. `access_denied` correctly drops its `action`, and no consumer outside this file depends on that (`humanFormOf` is only reached via `sendOAuthError`).

I could not execute the suite (shallow clone has no `node_modules`), but I traced each assertion against the implementation and they are consistent with the PR's reported "280 passed / property fails 7 ways on revert."

### Verdict
**approved** (informational — zero Critical findings; the three items above are Information only. A human still gives the explicit GitHub approval.)

jwfing (commented):
## Review — #103 The repair page's last branch still recommended the no-op

**Summary:** A tightly-scoped, well-tested fix that removes the remove-and-re-add no-op from the OAuth error page's `default`/`access_denied` paths, factors the real repair step into a shared `CLEARING_INSTRUCTION`, and adds correctly-scoped branches for `access_denied`, `unsupported_response_type`, and missing-PKCE — with property tests that cover every code rather than the two branches someone remembered.

### Requirements context
No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo, so there is no matching spec/plan — assessed against the PR description, linked context (#101, #74, #91), and the surrounding code. Intent is clearly documented in the PR body and in-code comments.

### Verification performed
Checked out head `c0a0e11` and ran the gates locally:
- `vitest run` (excluding the real-project integration test): **288 passed / 20 files** ✅
- Changed-file suite `oauth-error-response.test.ts`: **37 passed** ✅
- `tsc --noEmit`: clean on the changed files ✅
- `eslint` on all three changed files: 0 errors ✅
- `tsup` build: clean ✅

Traced the two load-bearing claims through the code and confirmed both:
- `sendOAuthError(req, res, status, body, human?)` spreads `{ ...humanFormOf(body, mcpUrl), ...human }` (`oauth-error-response.ts:389-404`), and `PKCE_REQUIRED_PAGE` sets `action: undefined` explicitly — so the override correctly *strips* the reconnect commands rather than leaving them attached. The page template renders no `<code>` block when `action` is falsy (`templates/oauth-error.ts:100-102`), so an actionless page is clean.
- `/oauth/callback` forwards the platform's verbatim `error` code into `sendOAuthError` when there is no registered redirect to bounce it to (`server.ts:668-671`), which is exactly how arbitrary/unknown codes reach the `default` branch. The PR's decision to keep that branch command-free and lead with the description is the right call.

---

### Findings

#### 🔴 Critical
(none)

#### 🟡 Suggestion
(none blocking; the two items below are informational)

#### ⚪ Information

**Functionality — `endsSentence` treats `:` and `;` as terminal punctuation** (`src/http/oauth-error-response.ts:233-236`)
The regex `/[.!?:;]$/` means a forwarded platform `error_description` that ends in a colon or semicolon (e.g. an OAuth `invalid_scope` message like `"Client not authorized for scopes:"`) is considered already-terminated, rendering `"…scopes: Start it again from your editor or client."` — a mild run-on rather than a clean sentence. Blast radius is cosmetic and only on platform-authored fragments, so no change is required; flagging only because `:`/`;` as sentence terminators is an unusual choice worth a one-line comment if it was deliberate.

**Software engineering — adjacent, pre-existing (out of scope): one browser-reachable path still returns raw JSON** (`src/http/server.ts:498-506`)
The over-long-`state` rejection returns `res.status(400).json(...)` directly rather than going through `sendOAuthError`, so a browser navigation that trips it would see raw OAuth JSON — the exact failure mode this file's helper exists to prevent (per its own header comment at `oauth-error-response.ts:5-23`). **This PR does not touch that block and it is not a regression** — noting it only because it is thematically adjacent to this change and would make a natural follow-up.

**Security — no security-relevant regressions.** The forwarded `error_description` is HTML-escaped at render (`templates/oauth-error.ts:108`), so leading with it introduces no reflected-XSS surface; `mcpUrl` is still built from the canonical `SERVER_CONFIG.publicUrl` (not the attacker-controlled `Host` header); no secrets/tokens are newly logged or returned; no auth/authorization checks are weakened; no new dependencies.

**Performance — no concerns.** All changes are pure, allocation-light string construction on a browser-only error path; no queries, loops, or blocking I/O introduced.

**Test quality — strong.** The new property tests assert on content (presence/absence of "Clear authentication", forbidding any "add(ing) it back" phrasing) and heading identity, not just length, and they cover both halves of the property (codes whose remedy *is* clearing vs. codes whose remedy is *not*), the PKCE override, blank/whitespace descriptions, and the punctuation seam. The two tables plus the "says what happened in its own words" test guard against the exact drift (`access_denied` sliding back into the catch-all) that caused the original bug.

---

No Critical findings. The change is correct, in-scope, well-tested, and safe. The two Information notes are optional follow-ups. Posted as a COMMENT — the explicit GitHub approval remains a separate human action.

jwfing (commented):
## Review — #103 (repair page must not recommend a no-op)

**Summary:** A tightly-scoped, well-tested correction of the OAuth error page so the catch-all `default` branch (and the miscategorised `access_denied`) stops recommending the remove-and-re-add no-op, plus a real crash fix for array-valued `error_description`; I verified it locally and found nothing blocking.

### Requirements context
No `/docs/superpowers/` (or any spec/plan) directory exists in this repo — assessing against the PR description, the `#101`/`#74` lineage it cites, and the surrounding code. Gates are `vitest` + `tsc` + `eslint` + `tsup`. I ran them against the head commit: `oauth-error-response.test.ts` → **42 passed**, `tsc --noEmit` → **clean**.

### Findings

#### Critical
(none)

#### Suggestion
- **Software engineering — the PKCE override's load-bearing seam is only tested in isolation** (`src/http/oauth-error-response.ts:404-419`, `src/http/server.ts:527-539`). The test pins `PKCE_REQUIRED_PAGE.action` as `undefined` as a constant, but the behaviour that actually reaches the browser depends on `sendOAuthError` merging `{ ...humanFormOf(body, mcpUrl), ...human }` — i.e. the override's *explicit* `action: undefined` key overwriting the `reconnectCommand(...)` that `humanFormOf('invalid_request')` returns. That "explicit undefined own-property wins the spread" is the subtle, regressable part (drop the `action: undefined` line and the constant test still passes while the page silently regains the commands). A one-line route/helper test asserting `sendOAuthError(..., PKCE_REQUIRED_PAGE)` renders no `<code>` block would close the gap the file is otherwise so careful about. Non-blocking.

#### Information
- **Functionality — sibling cast on the redirect-back path is unfixed but benign** (`src/http/server.ts:656-657`). The crash fix correctly guards the render path (`humanFormOf` now coerces non-string `error_description` to `undefined`, `oauth-error-response.ts:281-282`), and the fixed cast at `:670` was the reachable one. The twin `error_description as string` at `:657` feeds `URLSearchParams.set`, which stringifies an array to `"a,b"` rather than throwing — so no crash, just noted for completeness that the same latent cast pattern still lives one branch over.
- **Functionality — `endsSentence` treats `:` and `;` as terminal** (`oauth-error-response.ts:233-236`). A forwarded description ending in a colon renders `"Error: Start it again…"`. This looks intentional (avoids double punctuation) and only affects an attacker/platform-supplied string on the generic page; flagging only so it's a decision on record.
- **Software engineering — comment-to-code ratio is very high.** Consistent with the existing `reconnectCommand` style in this file, so not a request to change — the traces (Quinn/Max/Blair) are genuinely useful as the "do not simplify back" guardrail this PR exists to install.

### Dimension coverage
- **Security:** no concerns. The user/platform-controllable `error_description` is rendered only through `renderOAuthErrorPage`, which `escapeHtml`s heading, message, and every `action` line (`templates/oauth-error.ts:29,98-102`) — no XSS. Commands are built from `SERVER_CONFIG.publicUrl`, not the attacker-controlled `Host` header, so no host-header injection into pasted commands. No secrets/PII newly logged or returned; no auth/authorization path touched.
- **Performance:** no concerns — constant-time string work on an error path only.

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings; the array-`error_description` crash fix, the blank-description leading-dot fix, the punctuation-seam fix, and the branch re-categorisation are all real, correct, and backed by regression tests that the PR demonstrably fails against the old text. The Suggestion above is worth a follow-up line but does not block merge.

jwfing (commented):
## Review — #103 "The repair page's last branch still recommended the no-op"

**Summary:** Correctly finishes the #101 work by routing the OAuth error page's `default` branch away from the remove-and-re-add no-op, adds dedicated `access_denied` / `unsupported_response_type` / PKCE branches, and hardens the renderer against non-string `error_description` values — a clean, unusually well-tested bug fix.

### Requirements context
No matching spec/plan found — `insforge-mcp` has no `docs/superpowers/` (or `docs/`) directory, so this was assessed against the PR title/body, the linked prior work (#101, #74, #91), and the surrounding source. The PR body's claims were cross-checked against `server.ts` (the `/oauth/authorize` and `/oauth/callback` call sites) and the existing `reconnectCommand` documentation rather than taken on faith.

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering — the load-bearing override merge is untested at the `sendOAuthError` seam.** `src/http/oauth-error-response.ts:404-419` renders `{ ...humanFormOf(body, mcpUrl), ...human }`. The reason `PKCE_REQUIRED_PAGE` (`:252-259`) suppresses the reconnect commands is that it carries an **explicit `action: undefined` key** — the base `invalid_request` form returns `action: reconnectCommand(mcpUrl)` (`:308`), and only a present-with-`undefined` key overrides it in a spread. The tests assert `PKCE_REQUIRED_PAGE.action` in isolation (`oauth-error-response.test.ts:198`), but nothing drives `sendOAuthError`/`renderOAuthErrorPage` for the PKCE (or callback-override) path to prove the merged result actually has **no `<code>` block**. If a future edit ever omits that `action: undefined` key (relying on `Partial<HumanForm>` to "just leave it out"), the reconnect commands would silently leak back onto a page meant to have none, and no test would fail — the exact "it-cannot-fail" shape this PR is otherwise so careful to close. Worth one render-level assertion that the PKCE page emits no command block.

#### Information

- **Functionality — dedicated branches discard the platform's `error_description`.** The `default` branch now leads with the forwarded description (`:382`), but `access_denied` and `unsupported_response_type` (`:339-356`) return fixed copy and drop any platform-supplied detail. This looks deliberate (those codes are self-explanatory and the detail is still captured in analytics at `server.ts:645-648`), and I'm not asking for a change — just flagging that the two branches diverge from the "lead with the only account of what actually failed" principle stated for `default`.
- **Software engineering — `endsSentence` treats `:` and `;` as terminal.** `src/http/oauth-error-response.ts:233-236` — a platform description ending in a colon (e.g. `"Reason:"`) would render `"Reason: Start it again…"`. Harmless and vanishingly rare; noting only for completeness.
- **Style — comment volume.** The added JSDoc/inline comments are very long, but they match this file's established convention (`reconnectCommand` above them is equally dense) and each block carries a real "why". Not a finding — consistent with the repo's voice.

### Dimension coverage

- **Software engineering:** Strong. New property tests replace the "named two branches someone remembered" test with a property over *every* code (with and without a description), plus regression tests for the array/number/object/null `error_description` crash, blank-description handling, and the run-on-sentence seam. TDD is evident (author reports the property fails 7 ways against the old text). One coverage gap noted above.
- **Functionality:** Solid. Verified against `server.ts` that the callback forwards the platform code verbatim only when there's no registered redirect to bounce to (`:653-671`), that `access_denied` is the one code that actually reaches the catch-all as claimed, and that the PKCE override is wired at the sole `!code_challenge` site (`:527-539`). The `humanFormOf` normalization (`:281-282`) treating non-string as absent is correct and matches the tests. Claims in the PR body hold up.
- **Security:** Net-positive. This PR *fixes* an unauthenticated stack-trace disclosure — a repeated `?error_description=a&error_description=b` arrived as an array, hit `.trim()`/`.replace`, and rendered express's default error page (absolute paths, dependency tree) to any browser. Both layers are addressed: normalize in `humanFormOf` and coerce via `String(text)` in `escapeHtml` (`templates/oauth-error.ts:127-134`). All user-influenced fields (heading/message/action, including the forwarded description) are HTML-escaped at render, so no XSS is introduced. No secrets/PII newly logged; no auth/authorization checks weakened; no new dependencies.
- **Performance:** No concerns. Only constant-time string operations on the error path; no loops, queries, allocations, or blocking I/O added.

### Verdict
**approved** (informational — human approval is still a separate action). Zero Critical findings; the one Suggestion and the Information notes are non-blocking. This is a careful, well-reasoned fix with regression coverage that would have caught the original drift.

jwfing (approved):
LGTM - approved.

jwfing (commented):
## Review: The repair page's last branch still recommended the no-op

**Summary:** Corrects the `default` OAuth-error branch so it no longer recommends the measured remove‑and‑re‑add no‑op, factors clearing copy into one shared constant, adds `access_denied` / `unsupported_response_type` / PKCE branches, and hardens two layers against a non‑string `error_description` crash — all with the property tests that would have caught the regression. Verified locally: **52/52 tests pass** in the two changed test files, `tsc --noEmit` clean, `eslint` clean on all three source files.

**Requirements context:** No `/docs/superpowers/` (or `docs/specs/`) directory exists in this repo — assessed against the PR description, the linked #101/#74/#91 history quoted in the code comments, and the surrounding OAuth flow in `src/http/server.ts`.

---

### Findings

#### Critical
(none)

#### Suggestion
(none blocking)

#### Information

- **Functionality — `body.error` is cast, not normalized (`src/http/server.ts:668-671`).** The PR guards the non‑string `error_description` inside `humanFormOf` (`oauth-error-response.ts:281-282`), but `error` itself is still forwarded as `error as string` from `req.query`. A repeated `?error=a&error=b` arrives as an array. I confirmed this is **safe today**: `switch(<array>)` falls through to `default`, `escapeHtml` now coerces via `String()`, and nothing throws — the generic page renders. Noting only because it is the exact class of bug this PR fixes one field over; if you ever branch on `error` in a way that assumes string methods, the guard won't be there. No action required.

- **Security — reflected `error_description` on a branded page (`src/http/oauth-error-response.ts:380-383`).** `/oauth/callback` is an unauthenticated GET, so an attacker can craft `…/oauth/callback?error=x&error_description=<text>` and have arbitrary text render, led‑with, inside the InsForge‑branded error card. This is **HTML‑escaped** (verified — no XSS), and it is **pre‑existing** (the old `default` branch also rendered `body.error_description` directly), so it is not a regression from this PR. The only delta is that the text now sits inside trusted framing ("…Start it again from your editor or client."), which lends slightly more credibility to injected phishing copy. If you want to close it, a length cap or a "the server reported:" prefix on the forwarded description would help. Out of scope for this change.

---

### What I checked and found correct

- **PKCE override is wired right.** `PKCE_REQUIRED_PAGE` sets `action: undefined` explicitly, so the spread `{ ...humanFormOf(body, mcpUrl), ...human }` in `sendOAuthError` (`oauth-error-response.ts:416`) actually *clears* the `invalid_request` reconnect commands rather than leaking them — and `oauth-error-response.test.ts` pins that property. Had `action` been omitted instead of set to `undefined`, the commands would have survived; this is the correct choice.
- **Crash fix is genuinely two‑layered.** `humanFormOf` normalizes the known caller and `escapeHtml`'s `String(text)` (`oauth-error.ts:128`) backstops every browser‑visible field, so a future field can't reintroduce the same unauthenticated stack‑trace disclosure. Both layers are tested.
- **`endsSentence` / blank‑description handling** correctly turns `' '` → generic fallback and prevents the leading‑full‑stop and run‑on‑sentence seams; each has a dedicated test.
- **Test quality:** the rewrite from "the two branches someone remembered" to a property over every code, plus the `toBeDefined()` guard replacing the `if (!action) return` escape hatch, directly addresses the it‑cannot‑fail shape that let the original drift through. Software‑engineering dimension is a clear strength here.
- **Performance:** pure string operations, no new queries, loops, or blocking I/O — nothing on a hot path. No concerns.

---

### Verdict

`approved` (informational — human approval is still a separate action). Zero Critical findings; the two Information notes are pre‑existing/latent and non‑blocking. This is a tidy, well‑tested fix that both corrects the documented no‑op and closes a real unauthenticated info‑leak.

_Note: PR is already in `closed` state; posting the review as requested._

PR #107, opened 2026-08-01
refresh_token: stop every client dying after an hour

### LINKED ISSUE #104: DCR echoes unsupported grant_types back as registered
`POST /oauth/register` returns whatever `grant_types` the client asked for, without checking them against what the server supports. A client that trusts the registration response is told yes, then told no at the token endpoint.

## Measured on the deployed slug (`28b9b26`, 1.2.12)

```
DCR body                                    -> grant_types in the 201 response
{}                                          -> ["authorization_code"]                    honest default
{"grant_types":["authorization_code",
                "refresh_token"]}           -> ["authorization_code","refresh_token"]    ECHOED
{"grant_types":["password",
                "urn:...:device_code"]}     -> ["password","urn:...:device_code"]        ECHOED

/.well-known/oauth-authorization-server     -> grant_types_supported: ["authorization_code"]
/oauth/token with grant_type=refresh_token  -> 400 unsupported_grant_type
```

The default (no `grant_types` requested) is correct. Only the echo path is wrong.

## Why it matters

ChatGPT connectors and most MCP clients register asking for `refresh_token`. They receive a 201 confirming it and then fail at the token endpoint — the shape behind the "refresh fails during initialize" reports seen against other MCP servers. Today we also confirm the `password` grant, which we have never implemented.

## Expected

RFC 7591 §3.2.1: the response must be the *registered* metadata, not the request. Either
- filter to the intersection with `OAUTH_CONFIG.grantTypes` and return the corrected set, or
- reject with `invalid_client_metadata` when an unsupported grant is requested.

Filtering is the friendlier option and matches what the no-`grant_types` path already does.

## Notes

Independent of the refresh-token work on `feat/refresh-token-grant` — that removes the symptom for `refresh_token` only. The echo should be filtered regardless, and the same handler should be checked for `response_types` and `token_endpoint_auth_method`.

Found while auditing the ChatGPT-connector path; the rest of that path (DCR with a `https://chatgpt.com/...` callback, `/oauth/authorize`, RFC 8707 `resource`) works.

tonychang04:
## TL;DR

Every connected client dies about an hour after signing in and can only recover through a browser. The platform hands us a 30-day refresh token at every code exchange and we dropped it on the floor. This picks it up.

Not a flip blocker. Auth, so per Tony's 07-29 rule it needs QA sign-off before merge — @qa-bot's login-then-refresh is the outstanding item and waits on a credential that doesn't need a human.

## Why an hour

```
access-token.ts:151          min(ACCESS_TOKEN_TTL_SECONDS, platformRemaining)
platform ACCESS_TOKEN_EXPIRY 3600, a signed JWT — so the exp is readable
refresh_token in this repo   a response field we ignored + a grant we rejected
```

Annoying for a CLI. For a hosted connector — signed in once, in someone else's browser, then unattended — it is a broken integration.

## What the platform decides

Read from its oauth service rather than assumed, and both facts make this smaller than it looks:

- `REFRESH_TOKEN_EXPIRY = 30 * 24 * 3600` → ours accepts for 30 days. Accepting longer would advertise a life the credential inside does not have, which is the exact bug that had our access token claiming 24h while dying in one.
- Its refresh path **keeps the same refresh token** → no rotation, so one token is good for its whole window and there is no rotation race across concurrent refreshes.

## Shape

- `refresh-token.ts` — encrypted envelope, own label `mcp-refresh-token-v1`, fifth derived key. Encrypted rather than signed for the same reason as the access token, and it needs that reason more: this is the longer-lived credential by thirty days to one hour.
- Captured at the callback and carried through all three sealed envelopes. Our refresh token is minted in `createAuthorizationCode` beside the access token, because the user and project it needs are in scope exactly there — so the auth code carries two tokens of **ours** rather than one of ours next to a raw platform credential.
- The grant handler, and `grant_types_supported` now advertises it. Advertising is part of the feature: a client that cannot see `refresh_token` opens a browser every hour regardless of what the endpoint would answer.
- Closes #104 — registration echoed back whatever `grant_types` a client asked for, so a client requesting `password` was told yes and refused at the token endpoint.

## The three outcomes a client must tell apart

This is the part worth review attention.

```
our token will not open   400 invalid_grant     expired / forged / sealed under a
                                                rotated secret — ONE answer on
                                                purpose; distinguishing them is an
                                                oracle
platform unreachable      503 + Retry-After     not the client's fault, may work
                                                next time. Must NOT say "sign in
                                                again"
platform rejects US       503 + Retry-After     a wrong INSFORGE_CLIENT_SECRET, i.e.
                                                a botched rotation. Reported as
                                                invalid_grant it tells every user to
                                                re-auth, they all do, and every new
                                                sign-in fails identically
platform declines grant   400 invalid_grant     the grant is revoked or past 30 days.
                                                That one IS sign in again
```

The same refresh token goes back, not a new one: the platform keeps its own on refresh, so ours still names a live grant and re-sealing would hand the client a different string for no change in what it authorises.

## Verification

Unit + mutation on the sealed layers; the route branches black-boxed against a locally booted build, because this repo has no route harness and building one is a bigger job than the feature:

```
DCR password + device_code     400 invalid_client_metadata, naming both
DCR auth_code + refresh_token  201 — echoed because they are now TRUE
token, no refresh_token        400 invalid_request
token, garbage                 400 invalid_grant
token sealed under a ROTATED secret   400 invalid_grant
token OURS + valid, platform declines 400 invalid_grant
platform UNROUTABLE            503 + Retry-After: 5
platform rejects our creds     503 + Retry-After: 30 + operator log
code minted WITH a refresh token    -> response carries refresh_token
code minted WITHOUT one             -> response omits the key
```

@qa-bot independently confirmed the 503 branch discriminates, and found that the token response was dropping `refresh_token` entirely — minted, sealed, and discarded one line later, so no client could ever obtain one. Fixed and re-proved.

**Known gap, stated rather than implied:** a full login-then-refresh against the real platform has not run. Offline proves the wiring; only a real login proves the product, and a test of mine has twice now passed while the thing did not work.

## Rotation note for the runbook

`INSFORGE_CLIENT_SECRET` now invalidates **five** artifacts, not four. The refresh token is the one that makes a rotation user-visible longest — an access token dies within the hour anyway; a 30-day refresh token dies with the rotation and sends that user back through a browser.

---
## Summary by cubic
Adds end-to-end `refresh_token` support so clients can renew sessions without reopening a browser, fixing hourly disconnects. Also hardens OAuth flows and fixes a start-up regression in the published `insforge-mcp-server` binary.

- **New Features**
  - Capture the platform refresh token and seal it across state, code, and token using key label `mcp-refresh-token-v1` derived from `INSFORGE_CLIENT_SECRET`.
  - Add `/oauth/token` `refresh_token` grant with clear outcomes:
    - Invalid/expired/not-ours token → 400 `invalid_grant`.
    - Platform unreachable → 503 `temporarily_unavailable` + `Retry-After`.
    - Platform `invalid_client` (our creds wrong) → 503 + `Retry-After`.
  - Advertise `refresh_token` in `grant_types_supported` and include `refresh_token` in the `authorization_code` response. Return the same refresh token on refresh; issue a new access token.
  - Rotation note: a fifth derived key means rotating `INSFORGE_CLIENT_SECRET` now invalidates five artifacts; refresh tokens will require re-auth after rotation.

- **Bug Fixes**
  - DCR now rejects unsupported `grant_types` and `response_types` instead of echoing them (closes #104).
  - Token response includes `refresh_token` when minted; previously it was dropped.
  - Callback maps platform `invalid_client`/`unauthorized_client` to 503 `temporarily_unavailable` + `Retry-After` to avoid user re-auth loops when our `INSFORGE_CLIENT_SECRET` is wrong.
  - Export `app` and add a robust entry-point guard so routes are testable and the published `insforge-mcp-server` npm bin starts correctly (resolve `process.argv[1]` via `realpathSync`/`pathToFileURL`); importing no longer binds a port.

<sup>Written for commit 14bdc2e9474fff5cc5bd89db87ee6b0f4c3c29f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/107?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -95,8 +95,17 @@ export const OAUTH_CONFIG = {
   /** Scopes supported by this MCP server */
   supportedScopes: ['mcp:read', 'mcp:write', 'project:select'],
 
-  /** Grant types supported */
-  grantTypes: ['authorization_code'],
+  /**
+   * Grant types supported.
+   *
+   * `refresh_token` is advertised here, which is what puts it in
+   * `grant_types_supported` on the AS metadata and in the default registration
+   * response — the two places a client looks before it will ever try to renew.
+   * Advertising it is therefore part of the feature, not documentation of it: a
+   * client that does not see it here keeps signing in through a browser every
+   * hour even though the endpoint would answer.
+   */
+  grantTypes: ['authorization_code', 'refresh_token'],
 
   /** Response types supported */
   responseTypes: ['code'],
--- a/src/http/config.ts
+++ b/src/http/config.ts
@@ -221,6 +221,32 @@ export function accessTokenKey(): Buffer {
   return deriveAccessTokenKey(INSFORGE_CONFIG.clientSecret);
 }
 
+/**
+ * And a fifth, for the refresh token.
+ *
+ * Fifth purpose, fifth label. It seals a credential that outlives the access
+ * token by thirty times, so it is the one whose key most needs to be unrelated
+ * to the others: a refresh token is what an attacker would rather have.
+ *
+ * Note for whoever counts these next: rotating INSFORGE_CLIENT_SECRET now
+ * invalidates FIVE things rather than four, and the refresh token is the one
+ * that makes that rotation user-visible for longer — an access token dies in an
+ * hour anyway, but a refresh token someone was relying on for a month dies with
+ * the rotation and sends them back through a browser.
+ */
+const REFRESH_TOKEN_KEY_LABEL = 'mcp-refresh-token-v1';
+
+export function deriveRefreshTokenKey(clientSecret: string): Buffer {
+  if (!clientSecret) {
+    throw new Error('INSFORGE_CLIENT_SECRET is required: the refresh token key is derived from it');
+  }
+  return createHmac('sha256', clientSecret).update(REFRESH_TOKEN_KEY_LABEL).digest();
+}
+
+export function refreshTokenKey(): Buffer {
+  return deriveRefreshTokenKey(INSFORGE_CONFIG.clientSecret);
+}
+
 // ============================================================================
 // Redis Configuration
 // ============================================================================
--- a/src/http/credential-guards.test.ts
+++ b/src/http/credential-guards.test.ts
@@ -1,0 +1,175 @@
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import http from 'node:http';
+import type { AddressInfo } from 'node:net';
+
+/**
+ * The two guards that say "the platform rejected OUR credentials" rather than
+ * "your sign-in expired", driven through the real routes.
+ *
+ * These had NO coverage. QA proved it the only way that counts: she disabled
+ * each guard in turn and the suite stayed green at 314 both times. Both fixes
+ * were correct and both would have reverted silently — a refactor of either
+ * error-handling block puts a user back on "sign in again" during a rotation,
+ * with nothing going red. This exact class had already shipped once in this
+ * file, which is what makes an untested version of the fix unacceptable rather
+ * than merely untidy.
+ *
+ * Nothing here is unit-testable: the distinction lives in a route handler, and a
+ * test of an extracted helper would keep passing after someone deleted the call
+ * to it — the same shape as asserting a function's return value while the
+ * response drops the field. So this drives the routes, with the PLATFORM stubbed
+ * and only its answer changing between cases.
+ */
+
+/** The two shapes the platform's token endpoint actually returns, verbatim. */
+const PLATFORM_RESPONSES = {
+  // Our client_id/secret are wrong for this deployment — a botched rotation.
+  invalid_client: { status: 401, body: { error: 'invalid_client', message: 'Invalid client credentials' } },
+  // Our credentials are fine; the code or refresh token is not.
+  invalid_grant: { status: 400, body: { error: 'invalid_grant', message: 'Invalid authorization code' } },
+} as const;
+
+let platform: http.Server;
+let platformShape: keyof typeof PLATFORM_RESPONSES = 'invalid_client';
+let server: http.Server;
+let baseUrl: string;
+let issueRefreshToken: typeof import('./refresh-token.js').issueRefreshToken;
+let refreshTokenKey: typeof import('./config.js').refreshTokenKey;
+
+beforeAll(async () => {
+  // The platform stub, on an ephemeral port, so the env can point at it before
+  // the server module is imported and reads its config.
+  platform = http.createServer((_req, res) => {
+    const { status, body } = PLATFORM_RESPONSES[platformShape];
+    res.writeHead(status, { 'Content-Type': 'application/json' });
+    res.end(JSON.stringify(body));
+  });
+  await new Promise<void>((resolve) => platform.listen(0, '127.0.0.1', resolve));
+  const platformPort = (platform.address() as AddressInfo).port;
+
+  process.env.INSFORGE_API_BASE = `http://127.0.0.1:${platformPort}`;
+  process.env.INSFORGE_CLIENT_ID = 'test-client-id';
+  process.env.INSFORGE_CLIENT_SECRET = 'test-client-secret';
+  process.env.MCP_SERVER_URL = 'http://127.0.0.1';
+
+  // Imported AFTER the env is set, and importing no longer starts a listener —
+  // see the entry-point guard at the bottom of server.ts, which exists so this
+  // file can exist.
+  const [{ app }, refreshToken, config] = await Promise.all([
+    import('./server.js'),
+    import('./refresh-token.js'),
+    import('./config.js'),
+  ]);
+  issueRefreshToken = refreshToken.issueRefreshToken;
+  refreshTokenKey = config.refreshTokenKey;
+
+  server = app.listen(0, '127.0.0.1');
+  await new Promise<void>((resolve) => server.once('listening', () => resolve()));
+  baseUrl = `http://127.0.0.1:${(server.address() as AddressInfo).port}`;
+});
+
+afterAll(async () => {
+  await new Promise<void>((resolve) => server.close(() => resolve()));
+  await new Promise<void>((resolve) => platform.close(() => resolve()));
+});
+
+/** A refresh token of ours, sealed with the key this deployment derives. */
+function ourRefreshToken(): string {
+  return issueRefreshToken(
+    { userId: 'u1', platformRefreshToken: 'platform-refresh', projectId: 'p1' },
+    refreshTokenKey()
+  );
+}
+
+async function postRefresh(token: string) {
+  return fetch(`${baseUrl}/oauth/token`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+    body: new URLSearchParams({ grant_type: 'refresh_token', refresh_token: token }),
+  });
+}
+
+/**
+ * The callback needs a state the server will open, so this walks the real flow:
+ * register a client, start an authorize (which sets the sealed cookie and hands
+ * the platform a handle), then call back with a junk code carrying both.
+ *
+ * That is also exactly the container-side pre-flight from the cutover checklist
+ * — the check that proves the CONTAINER holds the right secret rather than
+ * proving the secret in someone's clipboard does.
+ */
+async function callbackWithJunkCode() {
+  const registered = await fetch(`${baseUrl}/oauth/register`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ redirect_uris: ['http://127.0.0.1:9999/cb'] }),
+  }).then((r) => r.json() as Promise<{ client_id: string }>);
+
+  const authorize = await fetch(
+    `${baseUrl}/oauth/authorize?response_type=code&client_id=${encodeURIComponent(registered.client_id)}` +
+      `&redirect_uri=${encodeURIComponent('http://127.0.0.1:9999/cb')}` +
+      `&code_challenge=${'a'.repeat(43)}&code_challenge_method=S256`,
+    { redirect: 'manual' }
+  );
+
+  const cookie = (authorize.headers.get('set-cookie') ?? '').split(';')[0];
+  const handle = new URL(authorize.headers.get('location') ?? '', 'http://x').searchParams.get('state');
+
+  return fetch(`${baseUrl}/oauth/callback?code=junk&state=${encodeURIComponent(handle ?? '')}`, {
+    headers: { cookie, accept: 'text/html' },
+  });
+}
+
+describe('the platform rejecting OUR credentials', () => {
+  it('answers the refresh grant with 503 and a Retry-After, never "sign in again"', async () => {
+    platformShape = 'invalid_client';
+    const response = await postRefresh(ourRefreshToken());
+
+    expect(response.status).toBe(503);
+    expect(response.headers.get('retry-after')).toBe('30');
+
+    const body = (await response.json()) as { error: string; error_description: string };
+    expect(body.error).toBe('temporarily_unavailable');
+    // The whole point: a wrong INSFORGE_CLIENT_SECRET must not send every
+    // connected user to re-authenticate, because they all would, and every new
+    // sign-in would fail the same way until an operator fixed the config.
+    expect(body.error_description).not.toMatch(/sign in again/i);
+  });
+
+  it('answers the sign-in callback with 503 and a Retry-After', async () => {
+    // The path a ROTATION actually hits: nobody holds a refresh token yet, so
+    // everyone is doing a fresh sign-in and arrives here instead.
+    platformShape = 'invalid_client';
+    const response = await callbackWithJunkCode();
+
+    expect(response.status).toBe(503);
+    expect(response.headers.get('retry-after')).toBe('30');
+    const page = await response.text();
+    expect(page).toMatch(/not with your account/i);
+    expect(page).not.toMatch(/sign in again/i);
+  });
+});
+
+describe('the platform declining the GRANT is still the user', () => {
+  it('keeps the refresh grant on 400 invalid_grant', async () => {
+    platformShape = 'invalid_grant';
+    const response = await postRefresh(ourRefreshToken());
+
+    expect(response.status).toBe(400);
+    const body = (await response.json()) as { error: string; error_description: string };
+    expect(body.error).toBe('invalid_grant');
+    expect(body.error_description).toMatch(/sign in again/i);
+  });
+
+  it('keeps the callback on 400, distinguishable from the credential case', async () => {
+    // The pair is what makes the container-side pre-flight readable: same
+    // request, two platform answers, two statuses. If these ever collapse to
+    // one status there is no unauthenticated way to tell a fumbled rotation
+    // from a bad code on the box that matters.
+    platformShape = 'invalid_grant';
+    const response = await callbackWithJunkCode();
+
+    expect(response.status).toBe(400);
+    expect(response.status).not.toBe(503);
+  });
+});
--- a/src/http/insforge-api.ts
+++ b/src/http/insforge-api.ts
@@ -354,6 +354,47 @@ export async function exchangePlatformCode(params: {
 }
 
 /**
+ * Trade the platform refresh token for a fresh platform access token.
+ *
+ * Same endpoint, same client credentials, same two-outcome contract as the code
+ * exchange above: a thrown InsforgeApiError means we could not get an answer, a
+ * returned body carrying `error` means the platform answered and declined. The
+ * caller needs those apart, because one is retryable and the other means the
+ * user has to sign in again.
+ *
+ * The platform does NOT rotate on use — its refresh path generates a new access
+ * token and keeps the same refresh token — so there is nothing to write back
+ * and no rotation race between two concurrent refreshes of one session. If that
+ * ever changes, the response's `refresh_token` becomes load-bearing and this
+ * comment is where to start.
+ */
+export async function refreshPlatformToken(params: {
+  refreshToken: string;
+  clientId: string;
+  clientSecret: string;
+}): Promise<PlatformTokens> {
+  const response = await platformFetch(`${OAUTH_API_BASE}/token`, {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({
+      grant_type: 'refresh_token',
+      refresh_token: params.refreshToken,
+      client_id: params.clientId,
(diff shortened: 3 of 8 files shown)


tonychang04:
## Sequencing: land and bake BEFORE the cutover window, not inside it

Per @infra-bot in-thread, recorded here so it does not live only in Slack.

The cutover window already carries three coupled changes — `MCP_SERVER_URL`, `INSFORGE_CLIENT_SECRET` and the DNS flip — and rollback stops being free the moment the first user re-authenticates on the new host. Adding a fourth change to the token endpoint, whose login-then-refresh path has not run against a real login, means that if anything misbehaves afterwards we cannot attribute it, on the one occasion we can least afford to be debugging.

**So: QA sign-off → merge → let it serve on the slug for a day → flip onto a build that has been running, not one that changed that morning.**

Checked whether the ordering costs users anything either way, since refresh interacts with the secret rotation (once this ships, `INSFORGE_CLIENT_SECRET` invalidates five artifacts rather than four, and the fifth is the 30-day one):

| | |
|---|---|
| lands **before** | bake-day sign-ins get refresh tokens; the rotation kills them at the window — but the flip forces a re-auth regardless |
| lands **after** | the flip cohort re-authenticates once at the flip and gets refresh later |

Same single re-authentication either way. There is no user-cost axis; attribution is the only one. Nobody should later read "refresh before rotation" as dangerous — it is neutral.

**Not a connector blocker at cutover.** An hourly re-login is bad for a hosted connector and is why this exists, but it is not worse than a four-variable flip that cannot be debugged.

jwfing (commented):
## Review — `refresh_token`: stop every client dying after an hour

**Summary:** A well-constructed, carefully-reasoned feature that picks up the platform refresh token we were dropping and adds a real `refresh_token` grant; the sealed-envelope layers are thoroughly unit-tested and the error taxonomy is thoughtful — no blocking issues, a few non-blocking gaps worth a look before the QA sign-off the PR already gates on.

### Requirements context
This repo has **no `docs/superpowers/` (or `docs/specs/`) directory** — assessed against the PR description, the linked issue **#104**, and the surrounding code (`access-token.ts`, `auth-state.ts`, `insforge-api.ts` conventions). The change is in-scope for its stated goal (renew sessions without a browser round-trip) and #104 (stop DCR echoing unsupported grants).

---

### Critical
(none)

---

### Suggestions

**1. Functionality — the load-bearing "platform does not rotate" assumption is unverified end-to-end, and the code discards a rotated token if the assumption is wrong.** `refreshPlatformToken` returns `PlatformTokens` including `refresh_token` (`src/http/insforge-api.ts:371-395`), but the grant handler ignores it and re-returns the client's *original* sealed token (`src/http/server.ts:1167-1176`). If the platform ever rotates on refresh, the platform refresh token embedded in that sealed envelope is dead after the first renewal and the session breaks after ~1 hour again — the exact failure this PR removes. The PR's own "known gap" says a real login-then-refresh has not run. This is the single highest-risk item; it's the right thing to nail during the QA sign-off the PR requires. Cheap belt-and-braces: if `tokens.refresh_token` is present and differs, re-seal instead of echoing.

**2. Functionality — the error taxonomy has a hole for platform-side 5xx with a JSON error body.** `refreshPlatformToken` never inspects HTTP status — it only branches on the parsed body (`insforge-api.ts:387-394`). The handler routes `error === 'invalid_client'` → 503 and network/parse failures → 503, but **any other structured error body falls through to `tokens.error || !tokens.access_token` → 400 `invalid_grant` "Sign in again."** (`src/http/server.ts:1138-1150`). So a platform-side 500/503 that happens to return `{"error":"server_error"}` or `{"error":"temporarily_unavailable"}` gets reported to every connected client as *their* grant expiring — the precise "blame the user's session for our/the platform's problem" this PR is built to avoid. The HTML-5xx case is handled correctly (`json()` throws → caught → 503), so this only bites when the platform returns a JSON error on a server fault. Consider treating a 5xx `response.status` as retryable (503) regardless of body.

**3. Software engineering — the #104 fix is incomplete for non-array `grant_types`/`response_types`.** Validation uses `asStringArray`, which treats a non-array as absent (`src/http/server.ts:439-459`), but the response still echoes the **raw** request value: `grant_types: grant_types || OAUTH_CONFIG.grantTypes` (`src/http/server.ts:496-497`). A client that sends `grant_types: "password"` as a *string* (not an array) passes validation and gets `"password"` echoed straight back — reintroducing the exact "told yes at registration, refused at the token endpoint" defect #104 closes, for the mistyped-input case. Low blast radius (compliant clients send arrays), but it undercuts the PR's own invariant that "a registration response is the server asserting what it will honour." Echo the validated/normalized array (or the defaults), not the raw body value.

**4. Software engineering / tests — no automated coverage for the `refresh_token` grant route branch.** The sealed layers (`refresh-token.test.ts`) and the three capture hops (`refresh-token-capture.test.ts`) are well covered, but the route branch that carries the whole 503-vs-400 argument (invalid_client → 503, unreachable → 503, grant declined → 400, success shape) has no test. The PR notes this repo has no route harness and that building one exceeds the feature — fair, and this is non-blocking. Worth flagging that the response-drop bug (refresh token minted, sealed, then discarded one line later) shipped and was caught only by manual QA; a single lightweight supertest-style exercise of this branch would have caught it and would guard the taxonomy in Suggestion #2.

---

### Information

**1.** Stale fallthrough message: the final `else` still says `"Only authorization_code grant type is supported"` (`src/http/server.ts:1177-1181`), which is now false — `refresh_token` is supported. Cosmetic.

**2.** The dedicated `invalid_client` path (Retry-After 30 + operator log, `src/http/server.ts:1117-1136`) only fires if the platform returns a *JSON* `{"error":"invalid_client"}` on the 401. If it returns a non-JSON 401 body, `json()` throws → caught → the generic 503/Retry-After 5 path. Still a 503 (safe direction), just less specific messaging and no operator log — noting in case the real platform's 401 body shape matters for observability.

---

### Dimension coverage
- **Software engineering:** strong — new key is symmetric-issue/read like the access token, null-vs-throw discipline preserved, ESM `.js` imports and the house comment style are consistent; gaps are the route-branch test (Suggestion #4) and the echo inconsistency (#3).
- **Functionality:** solves the stated bug; main risks are the unverified non-rotation assumption (#1) and the 5xx-body taxonomy hole (#2).
- **Security:** no new secrets logged (error logs are generic, tokens never printed); the refresh envelope is AES-256-GCM encrypted under a fifth distinct HMAC-derived key with cross-label rejection tested; the `invalid_client` → 503 (not re-auth) choice correctly avoids a re-auth stampede on a botched secret rotation. No auth/authz weakened. Note (design, not a defect): non-rotating refresh tokens mean a stolen token is usable for its full 30 days — an accepted trade, stated in the PR.
- **Performance:** one bounded (10s timeout) platform call per refresh, no N+1, no DB/index concerns, no blocking sync work on the event loop.

### Verdict
**`approved`** (informational — a human still gives the GitHub approval). Zero Critical findings; all items above are non-blocking. Recommend resolving Suggestions #1 and #2 (or explicitly confirming them) during the QA login-then-refresh the PR already requires before merge.

jwfing (dismissed):
LGTM - approved.

tonychang04:
## The container-side pre-flight depends on this branch

@manager found that the existing rotation check proves *the secret in your hand* authenticates — not that the **container** received that same value. Paste it correctly into the probe and wrongly into the Manufact env and the check still passes.

His closer: hand our own callback a junk code with a valid state, so the container performs a real token exchange with **its** configured secret.

On master the two outcomes are byte-identical — the platform's error is never rendered, so a wrong secret and a bad code both produce `400` and the same page. On this branch they separate by status, which makes the check scriptable rather than page-scraping:

```
platform says          container state              probe result
invalid_client   ->    WRONG secret in the env  ->  503  "Sign-in is temporarily unavailable"
invalid_grant    ->    right secret, bad code   ->  400  "Sign-in could not be completed"
```

Measured against a locally booted build of `d8a3c2e` with a stubbed platform returning each shape, driving the real two-step flow: register → authorize (capturing the `__Host-` cookie and state handle) → callback with `code=junk`. The `invalid_client` path also emits the operator log line naming the real cause.

So the callback branch here is not cosmetic: without it there is no unauthenticated way to tell a fumbled rotation from a bad code **on the box that actually matters**. That reinforces the land-and-bake-before-the-window sequencing rather than competing with it — the pre-flight that validates the rotation has to be deployed before the rotation it validates.

jwfing (commented):
## Review: `refresh_token` grant — end-to-end session renewal

**Summary:** A well-scoped, unusually well-tested implementation of the platform `refresh_token` grant that stops clients dying after an hour; the sealed-envelope design mirrors the existing `access-token.ts`/`auth-state.ts` patterns faithfully, and I found no blocking issues.

### Requirements context
This repo has **no `/docs/superpowers/` (or `docs/specs/`) spec directory** — assessed against the PR description, the referenced platform OAuth constants, linked issue #104, and the surrounding code (`access-token.ts`, `auth-state.ts`, `insforge-api.ts`). No matching spec/plan found.

---

### Critical
(none) — nothing here should block merge.

---

### Suggestion

**1. `unauthorized_client` is handled at the callback but not on refresh (functionality) — `src/http/server.ts:1156` vs `src/http/server.ts:829`.**
The callback path maps **both** `invalid_client` *and* `unauthorized_client` to a 503 "our-credentials-are-wrong" outcome. The refresh grant only checks `tokens.error === 'invalid_client'`. If the platform ever answers the refresh endpoint with `unauthorized_client` for a wrong `INSFORGE_CLIENT_SECRET`, the refresh path falls through to the generic `if (tokens.error || !tokens.access_token)` branch (`server.ts:1177`) and returns `400 invalid_grant` → "Sign in again." — which is exactly the rotation stampede this PR sets out to prevent. Confidence it fires is low (the primary `invalid_client` case is handled and tested, and the platform's documented answer is `invalid_client`), so this is a Suggestion rather than a blocker, but the two paths asserting the same contract should recognise the same error set. Consider `tokens.error === 'invalid_client' || tokens.error === 'unauthorized_client'` to match the callback.

**2. The refresh branch has no `try/catch` around `issueAccessToken` (software engineering / robustness) — `src/http/server.ts:1191`.**
`issueAccessToken` throws when the sealed token exceeds `MAX_ACCESS_TOKEN_LENGTH` (a large platform access token). Every other call site is guarded: the `authorization_code` branch wraps `exchangeCode` in `try/catch` (`server.ts:1088`), and `createAuthorizationCode`'s issue call is wrapped by the `selectProject` handler. Here the throw is unguarded — inside an async Express 4 handler that becomes an unhandled rejection and the request **hangs with no response** instead of returning a clean 500. Edge case (platform token > ~2950 chars), low blast radius, but worth wrapping the issue-and-respond tail so it degrades to a 5xx.

**3. Minor redundant decrypt (performance nit) — `src/http/server.ts:1200`.**
`readAccessToken(accessToken, accessTokenKey())` decrypts the token that was just sealed one line above, only to feed `accessTokenLifetimeSeconds`. You already hold the plaintext payload passed to `issueAccessToken`; calling `accessTokenLifetimeSeconds` on that payload avoids the extra GCM open. Negligible cost, purely tidiness.

---

### Information

**4. Stale error message in the fallthrough `else` — `src/http/server.ts:1216-1224`.**
The unsupported-grant branch still says *"Only authorization_code grant type is supported"*, but `refresh_token` is now supported and advertised in `grant_types_supported`. A client hitting the `else` (e.g. `password`) gets a message that contradicts discovery metadata. Cosmetic.

**5. No client authentication / refresh-token rotation on the grant — `src/http/refresh-token.ts`, `src/http/server.ts:1093`.**
The refresh grant authenticates purely by possession of the encrypted token (no `client_id` binding, no rotation). This is a deliberate, documented choice that matches the platform (which "keeps the same refresh token") and the existing access-token design (also unbound), so it's consistent rather than a regression — flagging only so a future reviewer doesn't re-derive it. Encryption is AES-256-GCM via the shared `sealAuthState`, with a distinct `mcp-refresh-token-v1` key label; `refresh-token.test.ts` proves cross-label and cross-secret tokens don't open, the credential never appears in the sealed blob, and the 30-day TTL boundary holds. Error logs name `INSFORGE_CLIENT_ID/SECRET` as misconfigured without printing their values — no secret leakage.

**6. Verification gap is disclosed, not hidden.** The author states a full real-platform login-then-refresh has not run, and per the 07-29 auth rule this is gated on QA sign-off before merge. The offline coverage here (sealed layers, the three-hop capture test, the credential-guard route tests that were proved by mutation) is strong; the one thing it cannot prove is the live wiring. Noting for the record — the process gate, not this diff, closes it.

---

### Dimension coverage
- **Software engineering:** Excellent. New behavior is covered per-layer (`refresh-token.test.ts`), per-hop (`refresh-token-capture.test.ts`), and at the route level with mutation-proven guards (`credential-guards.test.ts`). The entry-point guard (`server.ts:2020+`) using `realpathSync` + `pathToFileURL` correctly fixes the symlinked-bin startup regression and is the enabler for the route tests. Conventions (import style, `.js` specifiers, `null`-vs-throw error handling, comment density) match the file.
- **Functionality:** Solves the stated problem; the three-outcome contract (400 invalid_grant / 503+Retry-After / 400 grant-declined) is implemented and tested. Backward compatibility is handled — pre-change auth codes without a refresh token still redeem (`exchangeCode` spread, tested). Gaps limited to Suggestion #1.
- **Security:** No new user input reaches SQL/shell; the refresh token is encrypted (not signed), never logged, never placed in a URL/state. DCR now rejects unsupported `grant_types`/`response_types` instead of echoing them (#104), and the success echo is now a truthful subset. No new dependencies.
- **Performance:** One necessary extra platform round-trip on refresh; no N+1, no unbounded loops, no event-loop blocking. Only the trivial redundant decrypt in #3.

Zero Critical findings. Suggestions #1 and #2 are worth addressing before this carries production refresh traffic, but neither blocks merge. The GitHub green-check approval remains a separate human action; per repo policy this also still needs QA's login-then-refresh sign-off.

jwfing (commented):
## Review: `refresh_token` grant support (#107)

**Summary:** Picks up the platform refresh token that was previously read-and-dropped, seals it through all three envelopes, and adds a `refresh_token` grant with well-discriminated error outcomes — a correct, carefully-reasoned fix for the hourly client death, with one core-feature test gap worth closing before merge.

### Requirements context
This repo has no `docs/superpowers/` (or any spec dir) — assessed against the PR description, the linked issue #104, and the surrounding code (`access-token.ts`, `auth-state.ts`, `insforge-api.ts`) whose conventions this change mirrors. Per the PR itself, this is auth and needs QA sign-off (Tony's 07-29 rule) before merge; the outstanding item is a real login-then-refresh run.

### Findings

#### Critical
(none) — the implementation is correct and consistent with the established sealed-envelope pattern. The platform refresh token is sealed with AES-256-GCM under its own derived key label (`config.ts:243`), never logged or returned in the clear (`refresh-token.test.ts` proves this), and the three-outcome error contract (400 `invalid_grant` / 503 + `Retry-After` for both platform-unreachable and platform-rejects-our-creds) is implemented and tested on both the refresh grant and the callback (`credential-guards.test.ts`).

#### Suggestion
- **Software engineering / functionality — no happy-path test for the refresh grant.** `credential-guards.test.ts` boots the real app against a platform stub but only exercises `invalid_client` and `invalid_grant` shapes (`credential-guards.test.ts:29-33`). Nothing asserts the *success* path: platform returns a valid new `access_token` → `/oauth/token` issues a fresh access token, echoes the **same** `refresh_token`, and sets `expires_in` from the new platform JWT (`server.ts:1145-1215`). This is the feature's core behavior. The harness already exists — adding a `success` `platformShape` (returning a small JWT + refresh token) would black-box it cheaply and independently of a real account, closing part of the acknowledged "login-then-refresh not yet run" gap.
- **Functionality — DCR echo can omit `refresh_token` for a client that asks for it explicitly.** The registration response echoes the caller's `grant_types` when present: `grant_types: grant_types || OAUTH_CONFIG.grantTypes` (`server.ts:498`). A client that registers with an explicit `grant_types: ['authorization_code']` (valid — it passes the new guard) receives a response that does **not** list `refresh_token`, which by this PR's own "advertising is part of the feature" rationale (`config.ts:98-108`) would keep it opening a browser hourly. Blast radius is limited because the AS metadata endpoint still advertises it globally (`server.ts:367`, `grant_types_supported`), which is the primary discovery path — hence Suggestion, not Critical. Consider unioning the requested grants with the supported set, or noting that AS metadata is the intended discovery mechanism.

#### Information
- **Error-handling asymmetry in the refresh branch.** The `authorization_code` branch wraps its work in try/catch (`server.ts:1039-1092`), but the `refresh_token` branch calls `issueAccessToken` / `readAccessToken` outside any try/catch (`server.ts:1195-1205`). `issueAccessToken` throws on an over-length token. Because this is Express 5 (`package.json:56`), a rejected async handler is auto-forwarded to the default error handler (generic 500), so there is no hang — but the carefully-classified OAuth error contract this PR builds everywhere else is bypassed for that pathological case. Minor.
- **Success analytics client id.** `trackOAuthSuccess({ clientId: req.body.client_id || 'unknown' })` on refresh (`server.ts:1207-1210`) — refresh requests from public clients often omit `client_id`, so this will frequently record `'unknown'`. Cosmetic.

### Dimension coverage
- **Software engineering:** strong unit + mutation coverage on the sealed layers and the per-hop capture (`refresh-token.test.ts`, `refresh-token-capture.test.ts`) and route-level coverage of the credential-rejection discrimination; only the refresh-grant success path is untested (Suggestion above). Follows repo conventions (encrypted-not-signed envelope, separate key label, size bound, null-not-throw reader).
- **Functionality:** solves the stated hourly-death bug; correctly reuses the same non-rotating platform refresh token; older-code/no-refresh-token cases preserved (`exchangeCode` returns `undefined`, spread-in on the response). Closes #104 by rejecting unsupported `grant_types`/`response_types` at DCR instead of echoing them.
- **Security:** sound. Platform credential sealed under a fifth independent HMAC-derived key, not logged or leaked in error bodies; uniform `null` from `readRefreshToken` avoids an expired-vs-forged oracle; DCR change is a hardening (rejects `password`/`device_code`); no new dependencies; no auth check weakened.
- **Performance:** one extra outbound platform call per renewal (expected); no N+1, unbounded loop, or blocking work on the event loop.

### Verdict
**approved** (informational — a human still gives the GitHub approval via the approve flow). No blocking issues. The two Suggestions — a stubbed-success refresh-grant route test and the DCR echo gap — are worth addressing, and the PR-acknowledged real login-then-refresh QA run remains the gate before merge.

jwfing (approved):
LGTM - approved.

Issue #111, opened 2026-08-03
24 orphaned deployments still serve /oauth/* on public hostnames — shipping the OAuth fixes retires none of them


tonychang04:
Every deployment ever made for the `insforge-mcp` server is still answering HTTP on a public hostname. **25 machines are live right now**, all under the `mcp-bf35bbc3-*` app prefix; only **one** is the production target behind `mcp.insforge.dev`. The other **24 are orphaned predecessors**.

> **The hostname inventory is deliberately not in this issue.** This repo is public and #106 carries a complete unauthenticated reproduction; a list of live pre-fix hosts next to it is a recipe. Full inventory, per-machine versions and exposure status are in Notion: [insforge-mcp orphaned deployments — live inventory and exposure (2026-08-03)](https://app.notion.com/p/insforge-mcp-orphaned-deployments-live-inventory-and-exposure-2026-08-03-3b1db2d6fb2a81c897a6e8e16d3ebc6d).

## The exposure is live, not theoretical

Shipping the fixes for #102, #106 and #110 **retires nothing** — the vulnerable copies keep answering on their own hostnames.

Running #106's actual payload (`/oauth/callback?error=x&error_description=a&error_description=b` with `Accept: text/html`) against all 25 machines:

- **13 of the 25 return the stack trace** — `TypeError: text.replace is not a function` with absolute `/app/dist/...` paths, to an unauthenticated request, right now.
- `mcp.insforge.dev` and the three newest machines are **clean**. Production is genuinely fixed.

So the fix shipped and the vulnerable surface did not shrink. That is the whole problem in one sentence.

## Two control-plane facts that break the obvious triage

**1. Deployment status does not mean liveness.** `GET /api/v1/deployments?serverId=…` reports **5 running, 43 stopped, 33 failed**. **20 of the machines the API calls `stopped` answer HTTP right now.** `stopped` means "we stopped tracking it". Anything that infers live surface from deployment status — including our own tooling — gets a wrong answer.

**2. The reported `version` is not a security-posture signal.** Two machines report `1.2.12` **and leak**; three report `1.2.12` and are clean. The package version was bumped independently of the fix landing, so **there is no safe subset selectable by version, and no version-ordered reap**. All 24 get retired together. (Mechanism: the `escapeHtml2` frame lands on a different `dist` line on nearly every machine — the bundle genuinely differs build to build while the version string stays put.)

[...]

Issue #106, opened 2026-08-01
Unauthenticated stack-trace disclosure on /oauth/callback (live on the slug)


tonychang04:
## Live on the deployed slug, unauthenticated

```
GET /oauth/callback?error=x&error_description=a&error_description=b
Accept: text/html

400, and the body is Express's error page:
  TypeError: text.replace is not a function
    at escapeHtml2 (file:///app/dist/http-server.js:1701:15)
    at renderOAuthErrorPage (file:///app/dist/http-server.js:1694:10)
    at sendOAuthError (file:///app/dist/http-server.js:1746:49)
```

Absolute container paths and the dependency tree, to anyone who crafts the URL.
No login required — `/oauth/callback` is unauthenticated by design, and
`error_description` comes straight off `req.query`, where a repeated parameter
parses as an array.

`master` @ `28b9b26` reproduces it locally too, so this is not a deployment
artefact.

## The JSON path is fine — it is the HTML path only

```
same URL, no Accept: text/html   ->  {"error":"x","error_description":["a","b"]}   clean
?error=a&error=b                 ->  clean page (error is switched on, never escaped raw)
?error=x                         ->  clean page (control)
```

So it is one field, on the browser navigation that the callback always is.

## This is a regression the rewrite introduced, and the reason is worth stating

Production (`mcp.insforge.dev`, 1.0.0) does **not** leak — but not because it
handles the array. It has no HTML error page at all, and answers JSON
regardless of `Accept`:

```
mcp.insforge.dev  same request  ->  {"error":"weird","error_description":["a","b"]}
```

**The repair page — added to help stuck users — is what introduced the
disclosure.** Cutting `mcp.insforge.dev` over to this build ships the leak to
production.

## Previously reported and shipped anyway

I raised this exact `TypeError: text.replace is not a function` against PR #84
on 2026-07-31, as a review finding. #84 merged without it being fixed, and it
has been live since. A finding recorded only as a comment on an unmerged PR has
no owner once that PR merges — hence this issue, so it is tracked independently
of whichever PR happens to carry the fix.

## Fix

Normalising in `humanFormOf` (PR #103, `b03a467`) covers it, because
`sendOAuthError` spreads `humanFormOf(body)` straight into the renderer, so
nothing reaches `escapeHtml` unnormalised. Worth also making `escapeHtml`
defensive about non-strings, since it owns every browser-visible field and
currently trusts its input.

Filed separately from #103 so the leak does not depend on that PR landing.

[...]

PR #112, opened 2026-08-03
verify-deploy: fail the gate when a previous deployment is still serving

tonychang04:
**TL;DR** — the deploy gate now fails if a previous deployment is still answering. Run against production today it fails with **24 live predecessors**, which is #111 in one line of output.

## Why

Shipping a fix does not take the bug off the internet. Every deployment of this server is its own Fly machine on its own public hostname. `POST /deployments/{id}/activate` is documented as tearing down the previous production deployment — it does not, and nothing noticed for a month.

As of 2026-08-03: 24 predecessors still answering, 20 on a build predating the OAuth fixes, and **13 still serving the unauthenticated stack trace of #106** that production had already stopped serving. The fixes landed and the vulnerable surface did not shrink. Nothing in the repo could have told us that.

## What it does

Derives every predecessor hostname from the deployment number already present in `VERIFY_HEALTH_URL` (`mcp-<serverId>-<n>.fly.dev`, so `n-1 … 1`) and probes them, failing the gate on any that answers as us.

**It probes rather than queries, because both cheaper signals are actively misleading** — this is the part worth reviewing:

- the control plane reports `status: stopped` for machines that are serving traffic (20 of the 24), and
- `/health` reports a version bumped independently of the fixes landing, so **two machines advertised the fixed version while still leaking**.

There is no safe subset selectable by either field. "Does the hostname answer, as us" is the only true statement available.

## Two judgement calls

**Scored as `target`, not `external`.** These are other hosts, which normally means `external` under this file's doctrine. But a surviving predecessor is a property of *this* deploy — the thing it failed to do — not the health of a third party we depend on. It should gate.

**A bare 200 is not enough, and I had this wrong first.** The file already argues 40 lines above that "a 200 on /health proves something answered, not that WE answered". My first version counted any 200 as alive, which would block a clean deploy forever if a retired hostname were ever fronted by a proxy answering 200. It now requires `server === "insforge-mcp"`, the same standard the target `/health` is held to.

**When `VERIFY_HEALTH_URL` is absent it announces that it did not run.** Silence there reads as "no orphans found", which is the false green this exists to remove.

## Verified against production

```
FAIL  every previous deployment has been retired
      — 24 of 80 still answering /health: mcp-bf35bbc3-{41,43,…,80}.fly.dev
exit 1
```
and the passing direction, pointed at a deployment whose predecessors are genuinely retired:
```
PASS  every previous deployment has been retired
```
Both run against real infrastructure, not mocked. eslint clean.

## Cost, stated rather than hidden

Adds ~20s to the gate for 80 predecessors (32s total run). It grows with deployment count — at ~500 deployments this would be a couple of minutes. Bounded and concurrent (16 at a time), and retired hostnames fail at DNS in milliseconds, so only survivors cost a round trip. If that ever bites, the fix is to reap, which is the point.

## Does not fix

The 24 machines currently live. Reaping them is destructive and prod-adjacent and needs a human go-ahead — tracked in #111. This makes sure it cannot happen again silently.

Refs #111, #106

---
## Summary by cubic
Fail the deploy gate when any previous deployment still answers on its Fly hostname, so fixes aren’t left exposed on old hosts. Probes predecessor hosts derived from `VERIFY_HEALTH_URL` and only flags alive if `/health` returns `server: insforge-mcp`.

- **New Features**
  - Derives `mcp-<id>-<n>.fly.dev` predecessors from `VERIFY_HEALTH_URL` and probes them concurrently; fails if any answer as us.
  - Uses probing instead of control-plane status or version (both were misleading).
  - Requires `/health` to report `server: insforge-mcp`; a 200 alone doesn’t count.
  - Scored as `target` gate; if `VERIFY_HEALTH_URL` is missing/not Fly, it clearly reports that predecessors weren’t checked. Refs #111.

<sup>Written for commit eef74dba980f2b6db933be632d7beb6b867407cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/insforge-mcp/pull/112?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Strengthened deployment verification to detect and block releases when previous service instances remain reachable.
  * Added concurrent health checks to improve verification speed across prior deployments.
  * Deployment checks now clearly report when health-check configuration is missing or invalid instead of silently passing.

### DIFF
--- a/scripts/verify-deploy.mjs
+++ b/scripts/verify-deploy.mjs
@@ -211,6 +211,81 @@ if (memory && typeof memory.heapUsedPct === 'number') {
   console.log('  ~ heap not checked: /health reported no memory block');
 }
 
+// --- every predecessor has to be retired ----------------------------------
+//
+// Shipping a fix does not take the bug off the internet. Each deployment of
+// this server is its own Fly machine on its own public hostname, and on
+// 2026-08-03 twenty-four predecessors were still answering — most on a build
+// predating the OAuth fixes, thirteen still serving an unauthenticated stack
+// trace that production had already stopped serving. The fixes landed and the
+// vulnerable surface did not shrink (insforge-mcp#111).
+//
+// The platform is supposed to reap them: POST /deployments/{id}/activate is
+// documented as tearing down the previous production deployment. It does not,
+// and nothing noticed for a month — so the deploy gate asserts it from outside.
+//
+// This has to PROBE, and the two cheaper signals are both actively misleading:
+// the control plane reports `stopped` for machines that are serving traffic,
+// and /health reports a version that was bumped independently of the fixes
+// landing, so two machines advertised the fixed version while still leaking.
+// Neither field is load-bearing. "Does the hostname answer" is the only true
+// statement available, and it is the one this makes.
+//
+// Scored as 'target', not 'external', deliberately. These are other hosts, but
+// a surviving predecessor is a property of THIS deploy — the thing it failed
+// to do — not the health of some third party we happen to depend on.
+const flyHost = healthBase.match(/^https?:\/\/(mcp-[a-z0-9]+)-(\d+)\.fly\.dev$/i);
+if (!flyHost) {
+  // Never a silent skip. Saying nothing here reads as "no orphans found",
+  // which is exactly the false green this check exists to remove.
+  console.log(
+    '  ~ predecessors NOT checked: VERIFY_HEALTH_URL is not an mcp-<id>-<n>.fly.dev host.\n' +
+      '    Orphaned deployments keep serving the bugs this deploy just fixed.\n' +
+      '    On Manufact, set VERIFY_HEALTH_URL=https://mcp-<serverId>-<deployNo>.fly.dev'
+  );
+} else {
+  const [, appPrefix, currentDeploy] = flyHost;
+  const predecessors = [];
+  for (let n = Number(currentDeploy) - 1; n >= 1; n--) predecessors.push(n);
+
+  // Bounded concurrency: this is the widest thing the gate does and it runs on
+  // every deploy. Retired hostnames fail at DNS in milliseconds, so only the
+  // survivors — the ones worth waiting for — cost a round trip.
+  const BATCH = 16;
+  const alive = [];
+  for (let i = 0; i < predecessors.length; i += BATCH) {
+    const settled = await Promise.all(
+      predecessors.slice(i, i + BATCH).map(async (n) => {
+        // Same standard the target /health is held to above: a 200 proves
+        // something answered, not that OUR process did. A retired machine
+        // whose hostname is still fronted by a proxy answering 200 would
+        // otherwise block a clean deploy forever, and an orphan that has to
+        // be named to count for is the only one worth failing on — it is the
+        // one still serving /oauth/*.
+        const res = await fetchOrNull(`https://${appPrefix}-${n}.fly.dev/health`, {
+          headers: { accept: 'application/json' },
+        });
+        if (res?.status !== 200) return null;
+        const body = await res
+          .json()
+          .catch(() => undefined);
+        return body?.server === 'insforge-mcp' ? n : null;
+      })
+    );
+    alive.push(...settled.filter((n) => n !== null));
+  }
+  alive.sort((a, b) => a - b);
+
+  check(
+    'every previous deployment has been retired',
+    alive.length === 0,
+    alive.length === 0
+      ? `${predecessors.length} predecessor hostname(s) checked, none answering`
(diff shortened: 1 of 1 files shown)


jwfing (commented):
## Review — verify-deploy: fail the gate when a previous deployment is still serving

**Summary:** A tight, well-justified +75/-0 addition to `scripts/verify-deploy.mjs` that probes every predecessor `mcp-<id>-<n>.fly.dev` hostname and fails the gate on any that still answers *as us* — the logic is correct and no blocking issues were found.

### Requirements context
No spec/plan directory exists in this repo — there is no `docs/superpowers/`, no `docs/specs/`, and no `.claude/skills/` (verified in the workspace). This matches the known shape of `insforge-mcp` (script-and-server repo, gates are vitest/tsc/eslint/tsup). Intent was therefore assessed against the PR description and its linked issues **#111** (predecessors not reaped by `activate`) and **#106** (unauthenticated stack-trace leak). The implementation matches that stated intent: it asserts from the outside the property the platform fails to enforce.

### Findings

#### Critical
(none)

#### Suggestion

- **Functionality / integration — the new gate does not run in the automated workflow.** `scripts/verify-deploy.mjs:237` keys the whole check on `healthBase` (`VERIFY_HEALTH_URL`) matching a fly host, but `.github/workflows/verify-deploy.yml:59-62` only sets `DEPLOY_URL`/`EXPECTED_VERSION` and never sets `VERIFY_HEALTH_URL` (confirmed: the var is referenced nowhere outside this script). So on the scheduled/dispatch CI runs — the actual gate — this always takes the advisory "predecessors NOT checked" branch (`verify-deploy.mjs:238-245`) and never fails. The PR's headline ("the deploy gate now fails…") holds only for a manual Manufact invocation with `VERIFY_HEALTH_URL` pointed at the fly host. This is consistent with the *existing* strongest check (the version assertion is already gated behind the same var), so it's not a regression — but if the goal is to catch orphans in the pipeline, the workflow should derive and set `VERIFY_HEALTH_URL` (out of scope here; worth a follow-up).

- **Software engineering / tests — no automated coverage for the new logic.** The change adds nontrivial pure logic — the host regex `verify-deploy.mjs:237`, predecessor enumeration `:248-249`, and the `status===200 && body.server==='insforge-mcp'` alive-filter `:265-272` — but there is no unit test. This is consistent with the file's existing convention (the whole script is verified against real infra, and there is no test for any `scripts/*.mjs`), so it's non-blocking. That said, the parsing/filtering is testable without live infra; extracting a small `parseFlyHost()` / alive-classifier helper would let a vitest case pin the "proxy answers 200 with a foreign body ⇒ not counted" behavior the author (rightly) reasoned through by hand.

#### Information

- **Security — no concerns.** The probed hostnames are built from an operator-set env var constrained by a strict `^https?://(mcp-[a-z0-9]+)-(\d+)\.fly\.dev$` regex plus an integer counter, so there is no URL/shell injection surface; headers are static; the check emits only hostnames (no tokens/PII), and it consumes but never logs response bodies. The `server === 'insforge-mcp'` guard (`verify-deploy.mjs:272`) is the right call — a bare 200 from a proxy-fronted retired host will not falsely block a clean deploy.

- **Performance — bounded, but two small notes.** Concurrency is correctly capped at `BATCH=16` (`verify-deploy.mjs:254-276`), and the cost is transparently documented in the PR. Batches are processed *sequentially* (`await Promise.all` per slice), so one slow survivor causes head-of-line blocking that delays the next batch up to `TIMEOUT_MS` (10s); a promise-pool would smooth this, though it's immaterial at current deployment counts. Separately, the "retired hostnames fail at DNS in milliseconds" reasoning assumes `*.fly.dev` does not wildcard-resolve; if it does, each retired host still costs an edge round-trip — but the measured ~20s/80-predecessors figure is empirical and authoritative regardless.

- **Style — good reuse and consistency.** The block reuses the existing `check()` and `fetchOrNull()` helpers, matches the file's heavy-comment idiom, and the `'target'` scoping decision is defensible and clearly documented (`verify-deploy.mjs:234-236`). `Promise.all` cannot reject here because the mapped callback never throws (all fetch/JSON paths are null-guarded), so the gate won't crash into an unhandled rejection. The "never a silent skip" advisory branch is a nice touch that avoids a false green.

### Verdict
**approved** (informational) — zero Critical findings. The change is correct, in-scope, and well-argued; the Suggestions above are non-blocking follow-ups. Actual GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

